PR #10, opened 2022-01-28
Prepare locations dataset

AdamProbert:
This PR aims to provide a new spark job intended to be run on a glue cluster that is responsible for the following:
- Joining relevant fields from ASCWDS and CQC for use in later modelling
- Create a new feature `jobcount` to replace `wkrrecs` and `totalstaff`. `jobcount` will be used for all models moving forward as the sole value indicating number of jobs at a given location.
-  Clean up null values and format integer fields

### DIFF
--- a/jobs/prepare_locations.py
+++ b/jobs/prepare_locations.py
@@ -49,7 +49,7 @@ def main(workplace_source, cqc_location_source, cqc_provider_source, destination
 
     print(f"Reading CQC locations parquet from {cqc_location_source}")
     cqc_df = spark.read.parquet(
-        cqc_source).select(required_cqc_fields)
+        cqc_location_source).select(required_cqc_fields)
 
     print(f"Reading CQC providers parquet from {cqc_provider_source}")
     cqc_provider_df = spark.read.parquet(
--- a/jobs/prepare_locations.py
+++ b/jobs/prepare_locations.py
@@ -263,13 +263,25 @@ def collect_arguments():
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
-        "--workplace_source", help="Source s3 directory for ASCWDS workplace dataset", required=True)
+        "--workplace_source",
+        help="Source s3 directory for ASCWDS workplace dataset",
+        required=True
+    )
     parser.add_argument(
-        "--cqc_location_source", help="Source s3 directory for CQC locations api dataset", required=True)
+        "--cqc_location_source",
+        help="Source s3 directory for CQC locations api dataset",
+        required=True
+    )
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


jobs/prepare_locations.py
```diff
@@ -0,0 +1,256 @@
...
+            )
+        ), greatest(col("totalstaff"), col("wkrrecs"))
+    ).otherwise(col("jobcount")))
+
+    # Estimate job count from beds
+    input_df = input_df.withColumn("bed_estimate_jobcount", when(
+        (
+            col("jobcount").isNull() &
+            (col("numberofbeds") > 0)
+        ), (8.40975704621392 + (col("numberofbeds") * 1.0010753137758377001))
```
jimstamp, line 162: Magic numbers?

jobs/prepare_locations.py
```diff
@@ -0,0 +1,256 @@
...
+    # Remove rows with null for wkrrecs and totalstaff
+    input_df = input_df.filter("wkrrecs is not null or totalstaff is not null")
+
+    # Remove rows with null locationId
+    input_df = input_df.na.drop(subset=["locationid"])
+
+    return input_df
+
+
+def calculate_jobcount(input_df):
```
AdamProbert, line 99: Split this out into separate composable functions

AnthonyWheatley (commented):
I've had a look and whilst I don't fully understand all of the functions and methods used, it's very easy to read through and everything make sense in context with the comments you've added. 

I think Gary might be the best person to review the steps/methodology of creating the jobcount figure however (rows 90 to 231 I believe).

jobs/prepare_locations.py
```diff
@@ -0,0 +1,256 @@
...
+                ) |
+                (
+                    col("totalstaff").isNotNull() &
+                    col("wkrrecs").isNull()
+                )
+            )
+        ), coalesce(input_df.totalstaff, input_df.wkrrecs)
+    ).otherwise(coalesce(col("jobcount"))))
+
+    # Abs difference between totalstaff & wkrrecs < 5 or < 10% take average:
```
garypolzin, line 131: We spoke previously about changing all of the <5 and <10% to be a function, or something whereby if we changed it to <7 or <15% that we only need to change it once and it will apply to all

garypolzin (approved):
Feel like I can understand a lot of it and looks good - the change I've suggested would make blanket changing of rules easier/quicker/less prone to missing one

jobs/prepare_locations.py
AdamProbert: Aye good thinking - let me capture these values better.
Tah for the review

PR #11, opened 2022-02-02
Updated documentation for jupyter notebooks

AdamProbert:
Our notebooks docs are out of date - this updates them to account for our new EMR based approach


AnthonyWheatley (commented):
Looks good to me, only thing I would say is to add something which states that there is a financial implication of not terminating the cluster once you are finished with it (especially if this is left running overnight)

PR #14, opened 2022-02-07
Fix bulk download cqc location job

AdamProbert:
Our cqc location bulk download job failed this month due to a bad import:
image

This pr fixes this


AnthonyWheatley:
Looks good to me, seems like the fix has worked given the execution time was initially 45 seconds before it failed and now it has been running for 4 minutes

AnthonyWheatley (approved):
Looks good to me, seems like the fix has worked given the execution time was initially 45 seconds before it failed and now it has been running for 4 minutes

PR #21, opened 2022-02-16
Fixed test_filter_nulls on windows

AdamProbert:
This PR fixes a number of tests within `test_prepare_locations` that fail on windows machines.
The tests used to check the arrays had the same value and ordering, the ordering is unimportant. The new tests assert the same values irrespective of ordering utilising (the poorly named) `assertCountEqual`.

image


garypolzin (approved):
looks good to me!

PR #30, opened 2022-04-04
providerid now imports from CQC instead of ASCWDS

garypolzin:
there are 2 trello tickets related to this - both relate to dodgy looking providerids appearing in the dataset_locations_prepared which has knock on effects - turns out we were using providerid from ASCWDS which is manually entered by workplaces and not the providerid contained in the CQC dataset

https://trello.com/c/qKN1Xrak/216-general-locationids-appearing-as-providerids-in-datasetlocationsprepared-in-error
https://trello.com/c/i8KuvS7B/225-general-issue-incorrect-providerids-appearing-in-datasetlocationsprepared

### DIFF
--- a/jobs/prepare_locations.py
+++ b/jobs/prepare_locations.py
@@ -43,7 +43,6 @@ def get_ascwds_workplace_df(workplace_source, base_path=constants.ASCWDS_WORKPLA
         .select(
             col("locationid"),
             col("establishmentid"),
-            col("providerid"),
             col("totalstaff").alias("total_staff"),
             col("wkrrecs").alias("worker_record_count"),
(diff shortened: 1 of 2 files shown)


jobs/prepare_locations.py
```diff
@@ -74,6 +73,7 @@ def get_cqc_location_df(cqc_location_source, base_path=constants.CQC_LOCATIONS_B
         .parquet(cqc_location_source)
         .select(
             col("locationid"),
+            col("providerid"),
```
AdamProbert, line 76: Should add a test for this new column

AdamProbert (changes requested):

PR #49, opened 2022-05-30
Worker tracking file

garypolzin:
[Trello](https://trello.com/c/c1wxvsxg/293-epic1-track-movement-of-workers-year-on-year)
# Description

Replication of current Skills for Care process - to improve in the future

Select two time periods 12 months apart
Filter to workplaces who have updated recently (defined as within 6 months)
Filter to workplaces who appear in both places
Add worker level data from start period
Identify which workers stayed and which workers left their place of work

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


tests/unit/test_worker_tracking.py
```diff
@@ -0,0 +1,157 @@
...
+
+        filtered_df = worker_tracking.updated_within_time_period(
+            self.START_PERIOD_WORKPLACE_FILE
+        )
+
+        self.assertEqual(filtered_df.count(), 4)
+        self.assertEqual(filtered_df.columns, ["establishmentid"])
+
+        # asserts equivalent items are present in both sequences
+        self.assertCountEqual(
```
elcorbs, line 55: Should you be asserting that the lists are equivalent here? You check the count on line 51

garypolzin: Yea I'll drop line 51 and just keep this test - if this test passes then the count will also be correct

elcorbs (approved):
Looks good. The output data for this job will need to be partitioned, but we can do that together as a separate PR if you like.

PR #52, opened 2022-06-01
updated s3 filepaths from sfc-data-engineering to sfc-main-datasets

garypolzin:
s3 filepaths were changed as part of terraform changes but there were some old ones lingering in the code

### DIFF
--- a/terraform/pipeline/variables.tf
+++ b/terraform/pipeline/variables.tf
@@ -15,18 +15,18 @@ variable "region" {
 
 variable "ascwds_root_data_location" {
   type        = string
-  default     = "s3://sfc-data-engineering/domain=ASCWDS/"
+  default     = "s3://sfc-main-datasets/domain=ASCWDS/"
   description = "Root directory for all data sourced from the ASCWDS domain"
 }
 
 variable "data_engineering_root_data_location" {
   type        = string
-  default     = "s3://sfc-data-engineering/domain=data_engineering/"
+  default     = "s3://sfc-main-datasets/domain=data_engineering/"
   description = "Root directory for all data generated in the Data Engineering domain"
 }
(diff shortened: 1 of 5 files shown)


terraform/pipeline/variables.tf
```diff
@@ -15,18 +15,18 @@ variable "region" {
...
   type        = string
-  default     = "s3://sfc-data-engineering/domain=data_engineering/"
+  default     = "s3://sfc-main-datasets/domain=data_engineering/"
   description = "Root directory for all data generated in the Data Engineering domain"
 }
 variable "cqc_root_data_location" {
   type        = string
-  default     = "s3://sfc-data-engineering/domain=CQC/"
+  default     = "s3://sfc-main-datasets/domain=CQC/"
   description = "Root directory for all data sourced from the CQC domain"
```
elcorbs, line 30: Unless I'm mistaken, these variables aren't being used anywhere - can they be deleted?

utils/utils.py
```diff
@@ -57,7 +57,7 @@ def identify_csv_delimiter(sample_csv):
 
 
 def generate_s3_dir_date_path(domain, dataset, date):
-    dir_prepend = "s3://sfc-data-engineering"
+    dir_prepend = "s3://sfc-main-datasets"
```
elcorbs, line 60: It would be better to pass this in as a variable as the bucket name will be different for different workspaces. However, if you need to get this in quickly, we can work on that together afterwards.

terraform/pipeline/variables.tf
garypolzin: I couldn't find them anywhere else so I've deleted

elcorbs (approved):

garypolzin:
re-opening as new PR

PR #63, opened 2022-06-10
Build aggregation function for training columns

LauraZM9:
Ticket: [[Epic1] PREP - build column aggregation functions](https://trello.com/c/0YYXeZA6)

I've built the aggregation function for the training columns along with tests for it.


tests/unit/test_prepare_workers.py
```diff
@@ -32,6 +32,31 @@ def test_get_dataset_worker_has_correct_rows_number(self):
...
+        )
+
+    def test_replace_training_columns(self):
+        spark = utils.get_spark()
+        df = spark.read.parquet(self.TEST_ASCWDS_WORKER_FILE)
+        training_cols = utils.extract_col_with_pattern("^tr\d\d[a-z]", WORKER_SCHEMA)
+
+        df = prepare_workers.replace_training_columns(df)
+
+        self.assertTrue("training", df.columns[-1])
```
elcorbs, line 57: should this be assertEqual?

tests/unit/test_prepare_workers.py
```diff
@@ -32,6 +32,31 @@ def test_get_dataset_worker_has_correct_rows_number(self):
...
+        aggregated_col = prepare_workers.get_aggregated_training_column(
+            df, training_columns
+        )
+
+        self.assertEqual(aggregated_col.columns[-1], "training")
+        self.assertEqual(
+            aggregated_col.first()["training"],
+            '{"tr01": {"latestdate": 0, "count": 1, "ac": 0, "nac": 0, "dn": 0}}',
+        )
+
```
elcorbs, line 49: It might be good to also have a test that checks if the tr##flag is 0 then it isn't present in the training json.

tests/unit/test_prepare_workers.py
LauraZM9: Yes, that is correct! Changing it now.

tests/unit/test_prepare_workers.py
LauraZM9: Will this do it? If I check exactly what the first value in the "training" column looks like and also check that all the other training flags are zero.

PR #64, opened 2022-06-14
Add instructions to readme for setting up aws creds for terraform to use

elcorbs:


LauraZM9 (approved):
Good idea with the example file!

PR #68, opened 2022-06-16
Parameterize job names in step function & add logging

elcorbs:
# Description
- Parameterized job names in the ethnicity step function definition so that they will work across workspaces.
- Added a cloudwatch log group and changed the step function to write logs here. This is so that logs will be persisted even if a state machine is recreated. - [Trello card for this](https://trello.com/c/0WGQxpsC/318-add-cloud-watch-logging-to-state-machine-executions)
- Gave all workspaces read access to the CQC data in the main datasets bucket. So that we can test pipelines in a workspace without having to first copy over the CQC data.


terraform/modules/glue-crawler/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "crawler_name" {
+    value = aws_glue_crawler.crawler.name
+}
```
LauraZM9, line 3: You can add an empty line at the end.

LauraZM9 (approved):
This looks ok to me.

PR #69, opened 2022-06-16
Build aggregation function for job role

LauraZM9:
# Description
Ticket: [[Epic1] PREP - build column aggregation functions - job role](https://trello.com/c/QnTUCy60)

Aggregation function for `job_role`.

# How to test
Same way as described in the README.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/prepare_workers.py
```diff
@@ -14,8 +13,18 @@ def main(source, destination):
...
+    columns_to_be_aggregated_patterns = {
+        "training": {"pattern": "^tr\d\d[a-z]", "udf_function": get_training_into_json},
+        "job_role": {"pattern": "^jr\d\d[a-z]", "udf_function": get_job_role_into_json},
+        "qualifications": {},
+    }
+
     # TODO - replace training/jb/ql columns with aggregated columns
-    main_df = replace_training_columns(main_df)
+    for col_name, info in columns_to_be_aggregated_patterns.items():
+        for pattern, udf_function in info.items():
```
elcorbs, line 24: Do you need a loop here? Could you directly use `info["pattern"]` and `info["udf_function"]` instead?

jobs/prepare_workers.py
```diff
@@ -33,23 +42,24 @@ def get_dataset_worker(source):
     return worker_df
 
 
-def replace_training_columns(df):
-    training_cols = utils.extract_col_with_pattern(
-        "^tr\d\d[a-z]", worker_schema.WORKER_SCHEMA
-    )
-    df = add_aggregated_training_column(df, training_cols)
+def replace_columns_after_aggregation(df, col_name, pattern, udf_function):
```
elcorbs, line 45: This is named "after_aggregation" but looks like it is also performing the aggregation.

jobs/prepare_workers.py
```diff
@@ -33,23 +42,24 @@ def get_dataset_worker(source):
...
+    cols = utils.extract_col_with_pattern(pattern, WORKER_SCHEMA)
```
elcorbs, line 46: Can we rename `cols` to something like `cols_to_aggregate` to make it clearer?

elcorbs (commented):
I think the use of the dictionary is fine but maybe makes it a bit more confusing to read. 
The only other way I would do it would be to just call the function twice with the different values. So just repeating the following each time
```
match_training_columns = "^tr\d\d[a-z]"
main_df = replace_columns_after_aggregation(
        main_df, "training", match_training_columns , get_training_into_json
)
```
but if there are more to aggregate then this might be too much repetition.

jobs/prepare_workers.py
LauraZM9: That is correct, I didn't think of that. Thanks

jobs/prepare_workers.py
LauraZM9: Is this better?

PR #75, opened 2022-06-20
Github actions to trigger a circle ci workflow to destroy terraform

elcorbs:
Circle CI doesn't have an option to hook into branch deletion events on its own, so using github actions as an intermediary.

This PR creates a GitHub action that's triggered on a branch being deleted. 
The Github action will trigger a circle ci workflow pipeline, we pass the branch name to the pipeline via the `GHA_Meta` parameter. ([docs for the orb that does this](https://github.com/marketplace/actions/trigger-circleci-pipeline)). 
In the circle CI pipeline conditionals have been added to only run a new delete dev environments workflow if it's triggered in this way.
This workflow will run terraform destroy and delete the workspace.

I've commented out the step to destroy the terraform and have printed some info in Github actions and circleci to check it's hooked up properly first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,19 @@
 version: 2.1
 
+parameters:
+  GHA_Actor:
+    type: string
+    default: ""
+  GHA_Action:
+    type: string
+    default: ""
+  GHA_Event:
+    type: string
+    default: ""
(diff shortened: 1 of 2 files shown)


LauraZM9 (approved):
🤞

PR #77, opened 2022-06-21
Prepare locations - CQC location data cleaning

elcorbs:
[Trello](https://trello.com/c/yA3U9fTn/322-epic2-a-prepare-data-to-apply-model-feature-engineering-and-vectorise-columns)

# Description
A couple of transformations to the CQC location data in the prepare locations job.
- convert "dormancy" field from "Y"/"N" values to a boolean
- replace any occurrences of "Yorkshire & Humberside" with "Yorkshire and The Humber" in the region field. Both values were being used for the same region.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/prepare_locations.py
```diff
@@ -186,6 +186,15 @@ def get_cqc_location_df(cqc_location_source, import_date=None):
...
     cqc_df = format_import_date(cqc_df)
+    cqc_df = cqc_df.withColumn(
+        "region",
+        (
+            when(
+                cqc_df.region == "Yorkshire & Humberside", "Yorkshire and The Humber"
+            ).otherwise(cqc_df.region)
+        ),
+    )
+    cqc_df = cqc_df.withColumn("dormancy", cqc_df.dormancy == "Y")
```
LauraZM9, line 197: I find this line a bit confusing. Is it going to be `True` if `cqc_df.dormancy == "Y"` and `False` if not?

elcorbs: Yeah that's right. I think alternatively I  could use when/otherwise, so it would look like 
```
cqc_df.withColumn("dormancy", (
 when(cqc_df.dormancy == "Y", True),
 .otherwise(False)
)
```
which I thought was a bit long compared to just the conditional, but do you think it's clearer?

LauraZM9: Both are ok, I wasn't sure if I understood the first one correctly. 
I thought James needed `0` or `1` for the model. Will this work too or he'll have to change it again in his code?

elcorbs: It's a boolean in the notebook, so I guess it's fine 🤷‍♀️

LauraZM9: Oh, ok 😅

PR #78, opened 2022-06-21
Explode services offered column into a column per service

elcorbs:
[Trello](https://trello.com/c/yA3U9fTn/322-epic2-a-prepare-data-to-apply-model-feature-engineering-and-vectorise-columns)
# Description

Add a column for each service with a binary value for whether or not a location offers that service (found in the array of services offered).

Also added `pytest-watch` as a dev python package. It watches the python files for changes and then reruns the tests if it detects any. 
So you can run `ptw` to watch all the files, or `ptw path/to/test` to watch and run one test file or directory.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/locations_feature_engineering.py
```diff
@@ -15,9 +15,58 @@ def main(prepared_locations_source):
...
 
+    locations_df = explode_services(locations_df)
+
     return locations_df
 
 
+def explode_services(locations_df):
+    service_columns = services_lookup().items()
+    for column_name, service_description in service_columns:
+
```
LauraZM9, line 26: Can this be one line?
```suggestion
for column_name, service_description in services_lookup().items():
```
The name `service_columns` to me feels like it's only the columns.

elcorbs: because I've moved the lookup to another file it's quite long on one line now so I've just moved the `.items()` method call to the line below I'm not introducing a different variable name here anymore.

PR #88, opened 2022-06-23
Handle import dates for prepare workers

LauraZM9:
# Description
[[Epic1] PREP - Build function to handle all import dates](https://trello.com/c/SYOE2yyM)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -419,10 +419,8 @@ class UtilsTests(unittest.TestCase):
     def test_format_import_date_returns_date_format(self):
         df = utils.format_import_date(self.test_workplace_df)
 
-        self.assertEqual(
-            self.test_workplace_df.schema["import_date"].dataType, StringType()
-        )
         self.assertEqual(df.schema["import_date"].dataType, DateType())
+        self.assertEqual(str(df.select("import_date").first()[0]), "2022-01-01")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


garypolzin (approved):
thought format_import_date was already in utils but looks good

tests/unit/test_utils.py
```diff
@@ -410,6 +416,14 @@ def test_extract_col_with_pattern(self):
...
         )
 
+    def test_format_import_date_returns_date_format(self):
+        df = utils.format_import_date(self.test_workplace_df)
+
+        self.assertEqual(
+            self.test_workplace_df.schema["import_date"].dataType, StringType()
+        )
+        self.assertEqual(df.schema["import_date"].dataType, DateType())
+
```
garypolzin, line 426: please can you add a test which shows it more visually with an example too? 

Not 100% sure this works but something like (not ideal that the test file has 01 for day and month, could maybe change one in test file?)

```
self.assertEqual(self.test_workplace_df.select("import_date").first()[0], "01/01/2022")
formatted_df = utils.format_import_date(self.test_workplace_df)
self.assertEqual(str(df.select("import_date").first()[0]), "2022-01-01")
```

garypolzin (approved):
looks good - just requested a more visual checks, find them quite useful - especially when things go wrong!

PR #90, opened 2022-06-27
Add salary calculation function

LauraZM9:
# Description
[[Epic1] PREP/TRANSFORM - Build salary calc functions](https://trello.com/c/3vxyr8WY)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


elcorbs (commented):
Few suggestions for refactoring the if statements and writing more tests. I realise that you might be limited in the tests based on the combinations of data you can preset using the data generation package. Maybe there's a way to set multiple scenarios out or you could call the function to generate the data in each test and pass in the values you want to give for those columns for the different scenarios.

tests/test_file_generator.py
```diff
@@ -561,6 +561,13 @@ def generate_ascwds_worker_file(output_destination):
         .withColumnSpec("ql37year", 2021)
         .withColumnSpec("ql313app", 1)
         .withColumnSpec("ql313year", 2013)
+        .withColumnSpec("emplstat", 190)
+        .withColumnSpec("conthrs", 8.5)
+        .withColumnSpec("averagehours", 26.5)
+        .withColumnSpec("zerohours", "Yes")
+        .withColumnSpec("salaryint", 250)
+        .withColumnSpec("salary", 5200)
+        .withColumnSpec("hrlyrate", 100.5)
```
garypolzin, line 570: not sure what the best way of doing this is but it would be good to test the hours and pay rates for each eventuality, including some 'dodgy' ones where both `contracted_hrs` and `average_hrs` have been completed.

I've got quite a lot of mini test files in `test_ingest_cqc_care_directory` which do this sort of thing, instead of trying to create one file at the start which does everything

jobs/prepare_workers.py
```diff
@@ -136,6 +174,57 @@ def extract_qualification_info(row, qualification):
...
+
+    if row["emplstat"] not in ["Permanent", "Temporary"]:
+        if average_hrs:
+            return average_hrs
+
+    if average_hrs > 0:
+        return average_hrs
+
+    if contracted_hrs > 0:
+        return contracted_hrs
```
garypolzin, line 199: can you flip the order of these so we prioritise adding `contracted_hrs` over `average_hrs` please

jobs/prepare_workers.py
```diff
@@ -136,6 +174,57 @@ def extract_qualification_info(row, qualification):
...
+    return contracted_hrs, average_hrs
+
+
+def calculate_hourly_pay(row):
+    if row["salaryint"] == 250:
+        if row["salary"]:
+            try:
+                return round(row["salary"] / 52 / row["hrs_worked"], 2)
+            except:
+                return None
```
garypolzin, line 220: the exception might capture this but the formula will only work correctly if `hrs_worked` is >0

garypolzin (changes requested):

jobs/prepare_workers.py
LauraZM9: Good point, Gary! The exception was only capturing if `hrs_worked` was 0.

jobs/prepare_workers.py
```diff
@@ -136,6 +174,50 @@ def extract_qualification_info(row, qualification):
...
+
+    if row["emplstat"] in [190, 191]:
+        if row["zerohours"] == 1:
+            if average_hrs:
+                return average_hrs
+        else:
+            if contracted_hrs:
+                return contracted_hrs
+
+    if row["emplstat"] in [192, 193, 194, 196]:
```
elcorbs, line 189: Can we have a comment here about what these numbers mean please.

jobs/prepare_workers.py
```diff
@@ -136,6 +174,50 @@ def extract_qualification_info(row, qualification):
...
+        print("here", average_hrs)
```
elcorbs, line 190: Do you still need this print?

elcorbs (approved):
Looks good!

garypolzin (approved):

PR #94, opened 2022-07-04
Apply estimate residential jobs model

elcorbs:
[Trello](https://trello.com/c/iZbSL3nw/323-epic2-b-include-residential-role-calculation-model-in-estimate-jobs-replacing-magic-numbers)
# Description

Replaces two methods to model care home jobs by using a trained ML model stored in S3. 

It first gets the latest version of the model and the locations ml features dataset from S3.
It filters the features dataset to just keep the rows we are trying to model here (care homes with historical data) and applies the model to this dataset to make a prediction. 
It will then join this filtered down data with predications to the whole dataset based on locationid and snapshot date.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


JimmySunDance (approved):
Reviewed with Emma and Laura - Looks good to me :) +1

LauraZM9 (approved):

PR #97, opened 2022-07-06
Fix prepare workers job

LauraZM9:
# Description
[[Epic1] Fix prepare_workers job](https://trello.com/c/L9kbTLZi)

- Create a cleaning function that will cast the string columns to floats or integers
- Use smaller datasets to test each aggregation function in isolation
- Create a bigger dataset to test main is performing all the aggregations with the expected results
- Modify the schema so it will match the one in the Glue Catalog

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


tests/unit/test_prepare_workers.py
```diff
@@ -56,95 +89,250 @@ def test_main_adds_aggregated_columns(self):
...
+                "2.0",
+                "0",
+                "1",
+                "0",
+                "1",
+                "10.0",
+                "15.3",
+                "13",
+            ),
+        ]
```
elcorbs, line 203: Please can you turn the formatting off for these lines and have them as single rows

elcorbs (approved):

PR #99, opened 2022-07-07
Increase resources for prepare locations

elcorbs:
# Description
- Update glue job terraform module to allow the number of workers and worker type to be configurable.
- Increases the number of workers for the prepare locations job and gives it workers with more memory
- Upgraded glue version of prepare locations job

With the current resource allocation, the prepare locations job can't handle running against all the data with all versions.

### DIFF
--- a/terraform/modules/glue-job/optional-inputs.tf
+++ b/terraform/modules/glue-job/optional-inputs.tf
@@ -25,3 +25,15 @@ variable "glue_version" {
     error_message = "Must be one of 1.0, 2.0 of 3.0"
   }
 }
+
+variable "worker_type" {
+  description = "Glue worker type"
+  default     = "Standard"
+  type        = string
+}
+
+variable "number_of_workers" {
+  description = "Number of glue workers"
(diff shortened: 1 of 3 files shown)


terraform/modules/glue-job/optional-inputs.tf
```diff
@@ -25,3 +25,25 @@ variable "glue_version" {
...
+variable "number_of_workers" {
+  description = "Number of glue workers"
+  default     = 2
+  type        = number
+
+  validation {
+    condition     = var.number_of_workers >= 2
+    error_message = "Must be at least 2"
+  }
+}
```
LauraZM9, line 49: This means that by default they will run as before but you can also change the worker type and the number of workers, right?

LauraZM9 (approved):

terraform/modules/glue-job/optional-inputs.tf
elcorbs: Yep

PR #104, opened 2022-07-12
Partition worker dataset by snapshot date

elcorbs:
[Trello](https://trello.com/c/BgxSkZOG/340-epic4-add-date-partitions-to-data-outputs-from-transformations)
# Description
- Partitions workers prepared data by snapshot year, month & day. Getting this from the import_date but when we join with location we can take it from there.
- Change the job to only process dates after the last processed one.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


LauraZM9 (approved):
👍

PR #108, opened 2022-07-14
Ingest ONS data with lookup tables

elcorbs:
[Trello](https://trello.com/c/kWyv4ybS/339-add-ons-region-into-the-prepared-locations-data)
# Description
A glue job to save all ONS CSVs at a source location as parquet files in the datasets bucket and then do the same with any lookup CSVs. The job will check which import_dates have already been saved in the destination and ignore those.

The ONS csv must be saved in the subdirectory `/dataset=postcode-directory/year=2022/...` and the field lookup CSVs must be save in a sub directory for each field, for example `/dataset=postcode-directory-field-lookups/field=rgm/year=2022/...`.

Also added crawlers to crawl both of these datasets.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/ingest_ons_data.py
+++ b/jobs/ingest_ons_data.py
@@ -1,0 +1,74 @@
+import sys
+
+import pyspark.sql.functions as F
+from pyspark.sql.utils import AnalysisException
+
+
+from utils import utils
+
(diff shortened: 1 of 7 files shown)


garypolzin (approved):
Looks good - I like the method of saving the csv file lookups in s3

PR #117, opened 2022-07-21
Add missing columns to the earlier version of the worker datset

elcorbs:
[Trello](https://trello.com/c/21NY9kDy/368-read-data-separately-in-prepareworkers-according-to-the-version)
# Description
Reading in the two versions of data separately, adding in any missing columns to the earlier worker datasets and filling them with null values then unioning the two datsets together.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/prepare_workers.py
```diff
@@ -104,9 +104,25 @@ def get_dataset_worker(source, schema, since_date=None):
...
+    )
+
+    worker_df_v0 = worker_df_v0.withColumn("previous_pay", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("savedate", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("previous_mainjrid", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("dayssick_savedate", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("pay_savedate", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("flujab2020", F.lit(None))
+    worker_df_v0 = worker_df_v0.withColumn("derivedfrom_hasbulkuploaded", F.lit(None))
+
```
LauraZM9, line 121: Would it be better to do a loop here?

tests/unit/test_prepare_workers.py
```diff
@@ -88,19 +89,38 @@ def test_main_aggregates_right_columns(self):
...
+        df = prepare_workers.main(self.TEST_ASCWDS_WORKER_FILE, self.TEST_SCHEMA)
+
+        v0_worker = df.where(df.workerid == "855821").first()
+
+        # Set missing column values to none
+        self.assertIn("savedate", df.columns)
+        self.assertEqual(v0_worker.savedate, None)
+        self.assertIn("pay_savedate", df.columns)
+        self.assertEqual(v0_worker.pay_savedate, None)
+
```
LauraZM9, line 118: There is a method called `assertIsNone` that we could use. I need to modify some of my tests too 😆

LauraZM9 (approved):
Looks good!

tests/unit/test_prepare_workers.py
elcorbs: Nice, I will replace them

PR #119, opened 2022-07-21
State machine to run crawler & some tidy up

elcorbs:
# Description
- A state machine to run a crawler. If the crawlers already running it will wait for the crawler to finish before running it again.
- Renamed the TransformASCWDS state machine to Ingest ASCWDS.
- Using the run crawler state machine in the ingest ASCWDS state machine so that it doesn't error when ingesting both csv files.
- Deleted the state machine that was created for the demo as it's been superseded by the Data Engineering pipeline state machine.


LauraZM9 (approved):
I can't say much about this 😅

PR #120, opened 2022-07-22
Use dataframe with only the date columns to work out start & end dates

elcorbs:
# Description

Dropping all the columns apart from import date to calculate the start and end dates for the worker tracking job massively improves the performance of the job.


garypolzin (approved):
yea looks great thanks

PR #121, opened 2022-07-22
Set the default glue version to 3 when creating a glue job

elcorbs:
# Description

This sets the default version of glue to 3 when we create a glue job using the terraform glue job module.

Generally, It's best practice to be using the newest version of the software. In addition, using glue version 3 will give us version 3 of spark. They have also reduced glue job start-up times in v. 3 which will save money on running costs. 

I'm keeping the jobs that are currently version 2 as version 2 for now but if we work on them can test them in version 3 and switch them if it doesn't break anything.

### DIFF
--- a/terraform/modules/glue-job/optional-inputs.tf
+++ b/terraform/modules/glue-job/optional-inputs.tf
@@ -18,7 +18,7 @@ variable "trigger_schedule" {
 variable "glue_version" {
   description = "Version of glue to use for the job. Default of 2.0"
   type        = string
-  default     = "2.0"
+  default     = "3.0"
 
   validation {
     condition     = contains(["1.0", "2.0", "3.0"], var.glue_version)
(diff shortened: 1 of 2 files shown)


terraform/modules/glue-job/optional-inputs.tf
```diff
@@ -18,7 +18,7 @@ variable "trigger_schedule" {
 variable "glue_version" {
   description = "Version of glue to use for the job. Default of 2.0"
```
garypolzin, line 19: change Default of 3.0

terraform/pipeline/glue.tf
```diff
@@ -24,6 +25,7 @@ module "ingest_ascwds_dataset_job" {
   glue_role       = aws_iam_role.sfc_glue_service_iam_role
   resource_bucket = module.pipeline_resources
   datasets_bucket = module.datasets_bucket
+  glue_version    = "2.0"
```
garypolzin, line 28: have you left some with 2.0 in as they haven't been tested on 3.0 whereas some others have so you've removed the glue_version?

garypolzin (approved):

terraform/pipeline/glue.tf
garypolzin: yea makes sense

PR #123, opened 2022-07-26
Stop the ONS lookup crawler from crawling the denormalised ONS date

elcorbs:
# Description
The lookups crawler should only be crawling the ONS lookup tables, there is another crawler that crawls ONS data & denormalised ONS data


garypolzin (approved):
hard to find fault with this one!

PR #131, opened 2022-07-29
PyDeequ package install & example glue job

elcorbs:
- Add steps to the deployment pipeline to install DeeQu jar & PyDeequ python package.
- Add the Jar and python zip file to glue jobs by default.
- Change the crawlers to exclude .json files, as the metrics will be stored as json files.
- An example glue job that runs some simple analysis & checks on the workplace dataset.

### DIFF
--- a/terraform/pipeline/glue.tf
+++ b/terraform/pipeline/glue.tf
@@ -194,7 +194,7 @@ module "collect_dq_metrics_on_workplaces_job" {
   glue_version    = "2.0"
 
   job_parameters = {
-    "--source" = "s3://sfc-main-datasets/domain=ASCWDS/dataset=workplace/"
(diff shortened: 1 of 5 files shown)


terraform/pipeline/glue.tf
```diff
@@ -185,6 +185,19 @@ module "bulk_cqc_locations_download_job" {
   }
 }
 
+module "collect_dq_metrics_on_workplaces_job" {
+  source          = "../modules/glue-job"
+  script_name     = "collect_dq_metrics_on_workplaces_job.py"
+  glue_role       = aws_iam_role.sfc_glue_service_iam_role
+  resource_bucket = module.pipeline_resources
+  datasets_bucket = module.datasets_bucket
+  glue_version    = "2.0"
```
garypolzin, line 194: can this be glue version 3.0?

garypolzin (approved):
can't see anything obvious but thanks for creating something as a starting point!

PR #138, opened 2022-08-08
Fix duplicate cols in worker tracking

garypolzin:
# Description
Had issue with duplicate columns when trying to run analysis.
Decided it was best to only import the relevant columns, then we can join the outputed data frame on to which ever worker file we deem necessary.

# Developer checklist
- [X] Unit test


JimmySunDance (approved):
LGTM

PR #145, opened 2023-02-01
Lock in version of black and remove outdated dependency to resolve broken build pipeline 

Catherine-Q-MT:
# Description
https://trello.com/c/w1jNNTF4

The circle CI build was breaking and in trying to resolve it was discovered an outdated dependency broke the pipenv install (which would block users from updating on their machines)

Circle CI issue
- python dependency 'black' release a breaking change today(1st Feb). We do not specify which version to install in either our requirements file or within the steps in terraform. 
- this meant developers who had not recently updated had no linting errors however when they pushed builds they would fail the linting step,

resolution:
-  lock in the versions in both places so they are consistent. 

Databricks issue:
- in previous work, databricks was being used but later that usage was removed. 
- the version specified was taken out of use and so when users ran either 'pipenv install' or 'pipenv update' they would see an error and they would not be able to successfully setup the environment on thier machine, blocking development

# How to test
Pipeline should build successfully
run the data pipeline upon merge to show no regression (note a partial run was done in the dev deploy with no issue but full end to end test after merging reccomended)

Post check
- Users should rebase their code on their machines and run pipenv update to ensure changes are reflected 

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


MarionJHolloway (commented):
Black version specified and databricks removed

MarionJHolloway (approved):

PR #156, opened 2023-02-09
adding in job source logic to coalesce job calculation and enhancing …

Catherine-Q-MT:
…tests

# Description
once we have added the new column to our dataset we need to start adding the logic for populating the source. 

# How to test
Unit tests should pass and contain assertion that test the contents of the job source column, these should then be visible in Athena if we run the prepare locations job

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_calculate_job_coalesce_total_staff_worker_records.py
+++ b/tests/unit/test_calculate_job_coalesce_total_staff_worker_records.py
@@ -40,16 +40,18 @@ class TestJobCountCoalesceWorkerRecords(unittest.TestCase):
             ("1-000000002", 30, None, 25, None, None),
             ("1-000000002", 35, 40, 25, None, None),
         ]
+
+        rule_name = "coalesce totalstaff worker records"
         df = self.spark.createDataFrame(data=rows, schema=self.calculate_jobs_schema)
 
-        df = calculate_jobcount_coalesce_totalstaff_wkrrecs(df)
+        df = calculate_jobcount_coalesce_totalstaff_wkrrecs(input_df=df, rule_name=rule_name)
         self.assertEqual(df.count(), 3)
         df = df.collect()
         self.assertEqual(df[0]["job_count"], 20)
-        self.assertEqual(df[0]["job_count_source"], "coalesce_total_staff_wkrrecs")
+        self.assertEqual(df[0]["job_count_source"], rule_name)
 
         self.assertEqual(df[1]["job_count"], 30)
-        self.assertEqual(df[1]["job_count_source"], "coalesce_total_staff_wkrrecs")
+        self.assertEqual(df[1]["job_count_source"], rule_name)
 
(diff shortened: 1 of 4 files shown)


garypolzin (approved):
One potential idea but feel free to disregard

tests/unit/test_calculate_job_coalesce_total_staff_worker_records.py
```diff
@@ -35,16 +36,15 @@ def setUp(self):
...
         df = calculate_jobcount_coalesce_totalstaff_wkrrecs(df)
         self.assertEqual(df.count(), 3)
-
         df = df.collect()
-        self.assertEqual(df[0]["job_count"], 20)
-        self.assertEqual(df[1]["job_count"], 30)
-        self.assertEqual(df[2]["job_count"], None)
+        self.assertEqual(df[0]["job_count"], 20, "coalesce_total_staff_wkrrecs")
+        self.assertEqual(df[1]["job_count"], 30, "coalesce_total_staff_wkrrecs")
+        self.assertEqual(df[2]["job_count"], None, None)
```
garypolzin, line 50: these tests aren't actually testing the source text, this is testing that the first two objects are equal, the third object isn't part of the tests

you'll need two tests for each, one for the job_count column and one for the job_count_source
self.assertEqual(df[0]["job_count"], 20)
self.assertEqual(df[0]["job_count_source"], "coalesce_total_staff_wkrrecs")

garypolzin (changes requested):

utils/prepare_locations_utils/job_calculator/calculate_jobcount_coalesce_totalstaff_wkrrecs.py
```diff
@@ -26,8 +25,24 @@ def calculate_jobcount_coalesce_totalstaff_wkrrecs(input_df):
...
+            F.when(
+                F.col("job_count_temp").isNotNull(), F.col("job_count_temp")
+            ).otherwise(F.coalesce(F.col("job_count")))
+        ),
     )
+    output_df = df_with_job_count_populated.drop("job_count_temp")
+    return output_df
 
 
 def select_the_non_null_value_of_total_staff_and_worker_record_count(input_df):
```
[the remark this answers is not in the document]
Catherine-Q-MT: great idea. will change

tests/unit/test_calculate_job_coalesce_total_staff_worker_records.py
Catherine-Q-MT: As discussed, how github is displaying this is confusing. in split mode we can see its written as you specified:
image

Catherine-Q-MT: thanks! will correct

PR #159, opened 2023-02-10
solution to show job source cols that requires less code

Catherine-Q-MT:
# Description
Our dependency on Spark's When and Otherwise Logic causes issues if we try and add the job count source logic within our individual rule sets and this is especially true with the bed count estimate logic which is already complex. 

If we elevate this logic to the job calculator we will not have these issues and we write less code and leave our rule sets as they are. We can see it matches the comments in the test cases:
image

# How to test
run prepare locations (you may need to delete a dataset created today or do this on your deploy) and confirm you can see values populated in the job count source
# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


garypolzin (approved):
Nice, happy with that!

PR #161, opened 2023-02-10
Additional tests for total_staff equalling worker_record_count

garypolzin:
Wrote these as part of the refactoring ticket, which we no longer need but may as well push the tests


Catherine-Q-MT (approved):
This is excellent and much clearer

PR #166, opened 2023-02-17
Add typing to csv to parquet

MarionJHolloway:
# Description
Created a new script for importing csv to parquet which enforces data typing.
Created a schema for importing job estimates from SPSS 

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


tests/unit/test_spss_job_estimates_schema.py
```diff
@@ -0,0 +1,68 @@
+import unittest
+
+from schemas.spss_job_estimates_schema import SPSS_JOBS_ESTIMATES
+
+
+class TestSpssJobEstimatesSchema(unittest.TestCase):
```
Catherine-Q-MT, line 6: as you have created a data class you can use the asdict method and make this test way less verbose which I would reccomend (dataclasses.asdict(<your class instance>)))

tests/unit/test_utils.py
```diff
@@ -336,6 +354,128 @@ def test_read_csv(self):
...
+        )
+
+        df = utils.read_csv_with_defined_schema(
+            self.example_csv_for_schema_tests_with_datetype, schema
+        )
+        df.show()
+
+        row_one = df.collect()[0]
+
+    # assert isinstance(row_one.string_field, type(None))
```
Catherine-Q-MT, line 473: are these supposed to be commented out?

MarionJHolloway: I have removed this test as it was no longer required. We no longer use date types in the schema.

Catherine-Q-MT (approved):

PR #168, opened 2023-02-21
Job_count process update to include min_permitted_value

garypolzin:
- Currently we have a min permitted value (set to 3) but it isn't used. Any job_count value should only be chosen if the total_staff or worker_record_count figure is greater than the min permitted

- Merged 'coalesce' (select the none null value) and 'tiny values' (select the none tiny value) into one script - one value above permitted and one value below permitted.

- Split out some logic so we can see whether the chosen value came from worker_record_count or total_staff

- Added additional tests and amended any tests which were previously giving the undesired output

[Trello](https://trello.com/c/KpUSA8iq/63-values-below-the-minimum-permitted-shouldnt-be-accepted-as-a-value-in-jobcount-process)


utils/prepare_locations_utils/job_calculator/job_calculator.py
```diff
@@ -13,53 +13,62 @@
...
+def calculate_jobcount(
+    input_df, total_staff_column, worker_records_column, output_column_name
+):
+    source_output_column_name = output_column_name + "_source"
 
     print("Calculating job_count...")
 
-    input_df = input_df.withColumn(column_name, F.lit(None).cast(IntegerType()))
-    input_df = input_df.withColumn(source_column_name, F.lit(None).cast(StringType()))
+    input_df = input_df.withColumn(output_column_name, F.lit(None).cast(IntegerType()))
```
joemc4, line 28: based on conversations today and the s3 bucket data type issue, I am just flagging this to be apprehensive

utils/prepare_locations_utils/job_calculator/job_calculator.py
```diff
@@ -19,7 +19,10 @@
 
 
 def calculate_jobcount(
-    input_df, total_staff_column, worker_records_column, output_column_name
+    input_df,
+    total_staff_column: str,
+    worker_records_column: str,
+    output_column_name: str,
 ):
```
joemc4, line 23: again, I am just flagging anything where we have set data type just out of caution incase something goes wrong we can flag

tests/unit/test_calculate_jobs.py
```diff
@@ -2,46 +2,58 @@
...
+    calculate_job_count_schema = StructType(
+        [
+            StructField("locationid", StringType(), False),
+            StructField("total_staff", IntegerType(), True),
+            StructField("worker_record_count", IntegerType(), True),
+            StructField("number_of_beds", IntegerType(), True),
+        ]
+    )
+
     def setUp(self):
```
joemc4, line 27: again, I am just flagging anything where we have set data type just out of caution incase something goes wrong we can flag

tests/unit/test_calculate_jobs_common_checks.py
```diff
@@ -78,6 +80,95 @@ def test_selected_column_is_not_null(self):
         self.assertEqual(result_df[0]["result"], False)
         self.assertEqual(result_df[1]["result"], True)
 
+    def test_selected_ascwds_job_count_is_at_least_the_min_permitted_when_true(self):
```
Catherine-Q-MT, line 83: nice refactoring here!

Catherine-Q-MT (approved):

PR #175, opened 2023-03-07
Filter cqc sector in estimate job count

MarionJHolloway:
# Description
Added a function to estimate job counts which filters to only the rows where cqc sector is independent. Empty strings are excluded.
Added a test to document behaviour of new function 
Added assertion to cqc sector creation test to check behaviour of empty strings

# How to test
Tests pass

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


tests/unit/test_filter_locations_prepared_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+            ("1-000000002", "Local authority"),
+            ("1-000000003", ""),
+        ]
+        df = self.spark.createDataFrame(
+            rows, schema=self.filter_locations_prepared_schema
+        )
+
+        filtered_df = filter_to_only_cqc_independent_sector_data(df)
+
+        self.assertEqual(filtered_df.count(), 1)
```
garypolzin, line 45: add a second assert to double check that the one left is the "independent" one

garypolzin (changes requested):
I'd rename the whole job & test file to either be more specific (like the job name = filter_to_only_cqc_independent_sector_data) or if you intend to build in multiple filters then change from 'locations_prepared_dataset' to 'estimate_job_count_dataset', as that's the datset which is being filtered

Catherine-Q-MT (approved):

garypolzin (approved):

PR #183, opened 2023-04-03
Rolling average for all locations, split by primary service

garypolzin:
Creates a rolling average column in estimated jobs for every location, split by primary service type

Need to double check it's looking ok in tableau prior to merging

[Trello ticket](https://trello.com/c/RBejz9ta/109-make-rolling-average-code-more-efficient)


jobs/estimate_job_counts.py
```diff
@@ -41,6 +41,9 @@
     filter_to_only_cqc_independent_sector_data,
 )
 
+# Note: using 88 as a proxy for 3 months
+NUMBER_OF_DAYS_IN_ROLLING_AVERAGE = 88
```
MarionJHolloway, line 45: I like that you moved this into estimate job counts - much easier to edit if needed

utils/estimate_job_count/models/primary_service_rolling_average.py
```diff
@@ -0,0 +1,87 @@
...
+    df = join_rolling_average_into_df(df, rolling_average_df)
+
+    return df
+
+
+def filter_to_locations_with_known_job_count(df: DataFrame) -> DataFrame:
+    return df.where((F.col(JOB_COUNT).isNotNull()) & (F.col(JOB_COUNT) > 0))
+
+
+def calculate_job_count_sum_and_count_per_service_and_time_period(
```
MarionJHolloway, line 38: I would recommend refactoring this. Functions should do one thing. That makes them easier to modify later down the line. I would either separate this into a function for each variable, or create sub-functions within it for each variable and then use this as a "put it all together" kind of function.

utils/estimate_job_count/models/primary_service_rolling_average.py
```diff
@@ -0,0 +1,87 @@
+from pyspark.sql import DataFrame
+import pyspark.sql.functions as F
+from pyspark.sql import Window
+
+from utils.utils import convert_days_to_unix_time
+from utils.estimate_job_count.column_names import (
```
MarionJHolloway, line 6: I think we could do with extracting the columns used in creating the rolling average as well - there's a few magic strings in here that are reused within this script

utils/estimate_job_count/models/primary_service_rolling_average.py
```diff
@@ -0,0 +1,87 @@
...
+def calculate_rolling_sum(
+    df: DataFrame, col_to_sum: str, number_of_days: int, new_col_name: str
+) -> DataFrame:
+    return df.withColumn(
+        new_col_name,
+        F.sum(col_to_sum).over(rolling_average_time_period(UNIX_TIME, number_of_days)),
+    )
+
+
+def rolling_average_time_period(unix_date_col: str, number_of_days: int):
```
MarionJHolloway, line 72: 1) need to specify return type here
2) rename as something like "calculate_rolling_average_time_period"

utils/utils.py
```diff
@@ -180,6 +181,18 @@ def format_import_date(df, fieldname="import_date"):
...
+def convert_date_to_unix_timestamp(
+    df: pyspark.sql.DataFrame, date_col: str, date_format: str, new_col_name: str
+) -> pyspark.sql.DataFrame:
+    return df.withColumn(
+        new_col_name, F.unix_timestamp(F.col(date_col), format=date_format)
+    )
+
+
+def convert_days_to_unix_time(days: int):
+    return days * 86400
```
MarionJHolloway, line 193: Refactor to remove magic number

MarionJHolloway (changes requested):
Looks good! - just a few points I spotted that need tidying up

tests/unit/test_utils.py
```diff
@@ -590,6 +590,28 @@ def test_format_import_date_returns_date_format(self):
...
+        ]
+        df = spark.createDataFrame(row, schema=column_schema)
+        df = utils.convert_date_to_unix_timestamp(
+            df, "snapshot_date", "yyyy-MM-dd", "snapshot_date_unix_conv"
+        )
+
+        df = df.orderBy("locationid").collect()
+        self.assertEqual(df[0]["snapshot_date_unix_conv"], 1672531200)
+
+    def test_convert_days_to_unix_time(self):
```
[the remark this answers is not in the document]
garypolzin: this one is a bit different - we type the number of days to be used in the rolling average calculation but the calculation itself needs it in unix time, so it's just the numerical input

rangeBetween(-convert_days_to_unix_time(number_of_days), 0)

utils/estimate_job_count/models/primary_service_rolling_average.py
garypolzin: hmm, it's only doing one aggregation so in that sense it's only doing one thing? Seems overkill and more at risk of them becoming out of sync to do the same aggregation function separately twice, once for sum and once for count, then joining them together?

I tend to worry less about functions which we haven't created ourselves, groupBy/agg/sum/count are all standard and will be fully tested as they are, so we can trust that they work as expected a lot more than a self written function

utils/estimate_job_count/models/primary_service_rolling_average.py
garypolzin: this returns the window to be used in the calculation, not sure what type that is!

I've renamed to "define_window_specifications"

tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -0,0 +1,91 @@
...
+            self.rolling_sum_df, "col_to_sum", 3, "rolling_total"
+        )
+        self.assertEqual(df.count(), 7)
+        df = df.collect()
+        self.assertEqual(df[0]["rolling_total"], 10.0)
+        self.assertEqual(df[3]["rolling_total"], 54.0)
+        self.assertEqual(df[4]["rolling_total"], 64.0)
+        self.assertEqual(df[6]["rolling_total"], 21.0)
+
+    def test_join_rolling_average_into_df(self):
```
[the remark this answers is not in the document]
garypolzin: Need a hand with this, can't figure out how to do it!

utils/estimate_job_count/models/primary_service_rolling_average.py
garypolzin: Created a dataclass for these

utils/estimate_job_count/models/primary_service_rolling_average.py
garypolzin: rename to calculate_job_count_aggregates_per_service_and_time_period

tests/unit/test_model_primary_service_rolling_average.py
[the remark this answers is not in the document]
garypolzin: check 88 is 88

MarionJHolloway (approved):
Looks good, thanks

PR #187, opened 2023-05-05
Pir import with schema

garypolzin:
Currently we use csv_to_parquet with doesn't have a schema, but this does....

[Trello](https://trello.com/c/qdrqaQsJ/3-create-pir-data-schema-for-monthly-upload)


MarionJHolloway (approved):
Looks good to me!

PR #193, opened 2023-06-08
Dpr determine areas including carers on adass

MarionJHolloway:
# Description
This is the first step in the new pipeline which estimates whether adass survey data includes carers and calculates a percentage of service users (only) who employ staff. 

This process involved creating a lot of new variables which were similar but subtly different. When reviewing PR can you please check these are clear and easy to understand.

# How to test
Outputs will be slightly different to previous figures, but are currently very similar.
All tests are passing.
Data has been downloaded, inspected and looks accurate.

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


tests/unit/test_determine_areas_including_carers_on_adass.py
```diff
@@ -0,0 +1,329 @@
...
+from pyspark.sql import SparkSession
+from pyspark.sql.types import (
+    StructType,
+    StructField,
+    StringType,
+    IntegerType,
+    FloatType,
+)
+
+from utils.prepare_direct_payments_utils.determine_areas_including_carers_on_adass import (
```
garypolzin, line 13: for consistency with other jobs can you change to

`import utils.prepare_direct_payments_utils.determine_areas_including_carers_on_adass as job`

then later on you'd call the function `job.determine_areas_including_carers_on_adass`
saves the long list of functions too!

tests/unit/test_determine_areas_including_carers_on_adass.py
```diff
@@ -0,0 +1,274 @@
...
+                StructField(DP.IMD_SCORE, FloatType(), True),
+                StructField(DP.DPRS_ADASS, FloatType(), True),
+                StructField(DP.DPRS_EMPLOYING_STAFF_ADASS, FloatType(), True),
+            ]
+        )
+        df = self.spark.createDataFrame(rows, schema=test_schema)
+        output_df = job.determine_areas_including_carers_on_adass(df)
+        output_df_list = output_df.sort(DP.LA_AREA).collect()
+
+        self.assertEqual(
```
garypolzin, line 290: I've had issues with loads of decimal places previously and came across this approach

`self.assertAlmostEquals(output_df_list[0][DP.PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF], 0.5001596, places=5)`

garypolzin: places then cuts both numbers at the same number of dp and compares them

utils/prepare_direct_payments_utils/determine_areas_including_carers_on_adass.py
```diff
@@ -0,0 +1,146 @@
+import pyspark.sql.functions as F
+from pyspark.sql import DataFrame
+
+from utils.prepare_direct_payments_utils.direct_payments_column_names import (
+    DirectPaymentColumnNames as DP,
+)
+
+MOST_RECENT_YEAR = 2021
+CARERS_EMPLOYING_PERCENTAGE = 0.0063872289536592
```
garypolzin, line 9: might be worth adding a comment for CARERS_EMPLOYING_PERCENTAGE
where does it come from and do we intend to use this one number forever?

tests/unit/test_determine_areas_including_carers_on_adass.py
```diff
@@ -0,0 +1,274 @@
...
+    DirectPaymentColumnNames as DP,
+)
+
+
+class TestDetermineAreasIncludingCarers(unittest.TestCase):
+    def setUp(self):
+        self.spark = SparkSession.builder.appName("test_areas_including_carers").getOrCreate()
+
+        warnings.simplefilter("ignore", ResourceWarning)
+
```
garypolzin, line 26: I've started adding a test_main function to check row length (it's also quite handy for running that one test (with/without a show function) to check the whole job runs ok without running all the individual tests each time

MarionJHolloway: Sounds good, this isn't the full job, so I've created a similar test for this step and then when the job is finished, I can add something like this to the whole job.

PR #212, opened 2023-11-21
Upload capacity tracker data

MarionJHolloway:
# Description
A script to import care home and non-residential capacity tracker csv files to two parquet datasets.

# How to test
Unit tests passing
Run in branch - behaving as expected

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/jobs/ingest_capacity_tracker_data.py
+++ b/jobs/ingest_capacity_tracker_data.py
@@ -9,7 +9,7 @@ from schemas.capacity_tracker_schema import (
 from utils import utils
 
 
-def main(source, destination):
+def main(care_home_source, care_home_destination, non_res_source, non_res_destination):
     df = utils.read_csv_with_defined_schema(source, SPSS_JOBS_ESTIMATES)
     df_with_formatted_date = df.withColumn(
         "snapshot_date_formatted", F.col("Snapshot_date")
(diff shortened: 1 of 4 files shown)


jobs/ingest_capacity_tracker_data.py
```diff
@@ -0,0 +1,98 @@
...
+    df = utils.read_csv_with_defined_schema(source, schema)
+    df = add_column_with_formatted_dates(
+        df,
+        date_column_name,
+        date_format,
+    )
+    utils.write_to_parquet(df, destination, False)
+
+
+def add_column_with_formatted_dates(df, old_column, date_format):
```
garypolzin, line 41: change old_column to be more descriptive, the function above has date_column_name which would work

jobs/ingest_capacity_tracker_data.py
```diff
@@ -0,0 +1,98 @@
...
+import pyspark.sql.functions as F
+
+from schemas.capacity_tracker_schema import (
+    CAPACITY_TRACKER_CARE_HOMES,
+    CAPACITY_TRACKER_NON_RESIDENTIAL,
+)
+from utils import utils
+
+
+def main(care_home_source, care_home_destination, non_res_source, non_res_destination):
```
garypolzin, line 12: are you able to replicate the destination path from another job, such as ingest_cqc_pir_data
that job automatically constructs the destination file path

garypolzin (changes requested):
nice, neat and easy to follow but I think the auto generation of the destination will be advantageous if we plan to do this frequently. If it's going to be fairly infrequent then you could argue we don't need it

MarionJHolloway:
I've made the requested changes. Everything is running happily in branch.

garypolzin (approved):

PR #215, opened 2023-12-08
Update README.md

RoyPriceCS50:
I've copied the text from card in my KANBAN pile to Readme.md file.

Card was about linking Power BI and AWS.


MarionJHolloway (approved):
Looks good.

PR #220, opened 2024-01-08
Estimate iepa ratios

MarionJHolloway:
# Description
Adds a new step to calculate the PA ratio from survey data. Years which did not have a survey have a PA ratio hard coded, as this is unlikely to change value in the future.

# How to test
Tests passing
Runs in branch
Compared to previous estimates and preferred


utils/direct_payments_utils/estimate_direct_payments/calculate_pa_ratio.py
```diff
@@ -0,0 +1,70 @@
...
+    w = (
+        Window.partitionBy(F.lit(0))
+        .orderBy(F.col(DP.YEAR_AS_INTEGER).cast("long"))
+        .rangeBetween(-(range), 0)
+    )
+    df = df.withColumn(DP.COUNT, F.lit(1))
+    df = df.withColumn(DP.COUNT_OF_YEARS, F.sum(df[DP.COUNT]).over(w))
+    df = df.withColumn(DP.SUM_OF_RATIOS, F.sum(df[DP.AVERAGE_STAFF]).over(w))
+    df = df.withColumn(
+        DP.RATIO_ROLLING_AVERAGE, df[DP.SUM_OF_RATIOS] / df[DP.COUNT_OF_YEARS]
```
garypolzin, line 68: probably low priority but we are using a mix of F.col("col_name") and df["col_name"] in our code, should we try to stick to one for consistency?
Both look fairly common in our code, so we can deal with this at another point if we decide to and leave as it is now

utils/direct_payments_utils/estimate_direct_payments/calculate_pa_ratio.py
```diff
@@ -0,0 +1,70 @@
...
+    pa_ratio_df = add_in_missing_historic_ratios(average_survey_df, spark)
+    pa_ratio_df = apply_rolling_average(pa_ratio_df)
+    pa_ratio_df = pa_ratio_df.select(DP.YEAR_AS_INTEGER, DP.RATIO_ROLLING_AVERAGE)
+    return pa_ratio_df
+
+
+def exclude_outliers(survey_df: DataFrame) -> DataFrame:
+    survey_df = survey_df.where(
+        (survey_df[DP.TOTAL_STAFF_RECODED] < 10.0)
+        & (survey_df[DP.TOTAL_STAFF_RECODED] > 0.0)
```
garypolzin, line 27: assign magic numbers to a named variable to explain what they are

tests/unit/test_calculate_pa_ratio.py
```diff
@@ -0,0 +1,205 @@
...
+            (2015, None),
+            (2016, None),
+            (2017, 1.0),
+            (2018, None),
+            (2019, 1.0),
+            (2020, 1.0),
+            (2021, 1.0),
+            (2022, 1.6),
+            (2023, 2.2),
+        ]
```
garypolzin, line 120: would be good to decide when test files should be in the test job and when they should be in the test file generator. Larger ones make test files quite long (but also it's easier to read answers when the test file is in the same script)

garypolzin (changes requested):
sorry had to look through it quite quick but see what you think to the feedback below and we'll go from there

utils/direct_payments_utils/estimate_direct_payments/calculate_pa_ratio.py
MarionJHolloway: I think this is a bigger issue - we might need a style guide or similar - question for MadeTech perhaps?

utils/direct_payments_utils/estimate_direct_payments/calculate_pa_ratio.py
MarionJHolloway: Fixed!

tests/unit/test_calculate_pa_ratio.py
MarionJHolloway: I've moved these out and into separate files

garypolzin (approved):
happy with emended code and decisions to defer :)

PR #222, opened 2024-01-09
Create non pipeline utils

MarionJHolloway:
# Description
Added a folder with non-pipeline scripts to be run locally


x_non_pipeline_utils/cqc_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+        "Service type - Remote clinical advice service",
+        "Service type - Acute services without overnight beds / listed acute services with or without overnight beds",
+        "Service type - Ambulance service",
+        "Service type - Blood and Transplant service",
+        "Service type - Diagnostic and/or screening service - single handed sessional providers",
+        "Service type - Hospice services at home",
+        "Service type - Hyperbaric Chamber",
+        "Service type - Prison Healthcare Services",
+        "Service type - Specialist college service",
+    ]
```
garypolzin, line 41: We should probably get rid of service types and get them from SERVICE_DICT instead

x_non_pipeline_utils/add_sector_and_service_to_CQC_register.py
```diff
@@ -0,0 +1,81 @@
...
+    return data
+
+
+def get_main_service(row):
+    service = Values.service_not_found
+
+    for service_type in CqcCategories.SERVICE_TYPES:
+        if row[service_type] == Values.yes:
+            service = service_type
+            return service
```
garypolzin, line 58: We should probably get rid of the SERVICE_TYPES list and get them from SERVICE_DICT keys instead - they are repeated at the moment

garypolzin (changes requested):

garypolzin (approved):

PR #232, opened 2024-01-19
Define column names in one location

MarionJHolloway:
# Description
Added column names for datasets as dataclasses in utils scripts in IND CQC pipeline
Updated column names in scripts that won't be worked on over the next ten weeks

# How to test
All unit tests passing.
As part of the commit history, I used a test script to compare changes to schemas which weren't covered by unit tests.

# Developer checklist
- [x ] Unit test
- [x ] Linked to Trello ticket
- [x ] Documentation up to date


utils/ind_cqc_column_names/estimate_jobs_columns.py
```diff
@@ -0,0 +1,28 @@
+from dataclasses import dataclass
+
```
garypolzin, line 2: I'd be tempted to merging all the column headings used in the 'CQC IND filled post estimating' process into one list of columns

The jobs are currently split up but it's one process, so a lot of the columns will be repeated. We'll need to add ICB at some point, and maybe rename job_count to filled_post_count (quite easy to do this now given we're writing prepare_locations from scratch and not editing the current one)

Looking at the miro board it looks like we'll have 5+ data files for the process of estimating CQC IND jobs, so that's a lot of repeated column names?

utils/ind_cqc_column_names/ascwds_worker_columns.py
```diff
@@ -0,0 +1,702 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class AscwdsWorkerColumns:
```
garypolzin, line 5: for all the other datasets we've made the labels more readable, such as
care_home: str = "carehome"

but we've kept all the ascwds ones the same. We could just rename the ones we use for use for now then built up the list as we come to use other variables. Ideally we'd carry on the same naming conventions so 
employment_status: str = "emplstat"
main_job_role: str = "mainjrid"
etc

utils/ind_cqc_column_names/ascwds_worker_columns.py
```diff
@@ -0,0 +1,702 @@
+from dataclasses import dataclass
+
+
```
garypolzin, line 3: This is just a random one but given all these files have somewhat changeable columns do you think we sort (and maintain them) in alphabetical order?
particularly for ascwds with 600+ columns in each file, it's easy to find 'amhp' alphabetically but less so just looking for it

garypolzin (changes requested):

schemas/cqc_care_directory_schema.py
```diff
@@ -1,179 +1,171 @@
...
-        StructField("Service_type_Dental_service", StringType(), True),
-        StructField(
-            "Service_type_Diagnostic_andor_screening_service", StringType(), True
-        ),
+        StructField(ColNames.ServiceType.community_healthcare, StringType(), True),
+        StructField(ColNames.ServiceType.dental, StringType(), True),
+        StructField(ColNames.ServiceType.diagnostics, StringType(), True),
         StructField(
-            "Service_type_Diagnostic_andor_screening_service_single_handed_sessional_providers",
+            ColNames.ServiceType.diagnositc_sessional,
```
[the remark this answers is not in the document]
MarionJHolloway: Good spot, thanks

utils/ind_cqc_column_names/ascwds_worker_columns.py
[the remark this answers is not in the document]
MarionJHolloway: Direct payments column headings are quite different and can eventually be refactored into a similar folder structure. I kept this specific to this pipeline for now to avoid interdependencies later down the line

utils/ind_cqc_column_names/estimate_jobs_columns.py
MarionJHolloway: I was quite torn over which way to go with this. Maybe a compromise would be to move all the classes into one script and separate out the columns which recur in each script (e.g. location id, run dates), but keep the script specific column names in separate classes, in that same script?

utils/ind_cqc_column_names/ascwds_worker_columns.py
MarionJHolloway: Yes, good idea, I will sort them

utils/ind_cqc_column_names/ascwds_worker_columns.py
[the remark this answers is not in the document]
garypolzin: merging my responses for both this comment and the 'one list per pipeline job'

I guess what I'm saying is that at this point, most of them aren't specific to this pipeline (this ASCWDS file is the version we'll use as a starting point for all ASCWDS tasks, same for CQC.

and generally conscious that we'll soon have lots of very similar datasets which will almost be identical to each other such as:
ascwds and ascwds_anon (prov and workplace)
cqc registered (for this process) and cqc_deregistered (for reconciliation)
locations_prepared, locations_prepared_cleaned, with estimates added, with job roles added

so just being wary of loads of loads of 'column_name' lists which will need maintaining, and jobs failing because we update some of them but forget one further downstream - feels better to have one list per pipeline which contains everything for the pipeline?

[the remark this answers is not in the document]
MarionJHolloway: If we only have one file for the whole pipeline it will be very long. Several thousand rows. With data classes you can just import the classes that are relevant to the script.

Additionally, the raw data files are a separate case. They feed the schemas and don't necessarily match the formatting of the strings in later stages of the pipeline. 

One solution could be to have the data classes feed each other (so later scripts pull in values from scripts earlier in the pipeline). For example, in prepare_locations_cleaned.py we could put something like:

import prepare_locations_column_names

location_id = prepare_locations_column_names.location_id

utils/ind_cqc_column_names/ascwds_worker_columns.py
MarionJHolloway: Yeah, I agree - it's about 1500 variables to rename, so we could put in the time up front or just clean them up as we use them. I'm happy either way :)

utils/ind_cqc_column_names/ascwds_worker_columns.py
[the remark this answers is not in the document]
garypolzin: It wouldn't be that long, the full ASCWDS files aren't part of the pipeline. The pipeline for estimating jobs starts with the merging process, so at the moment it would be a combination of the
44 col names in prepare_locations
22 col names in prepare_locations_cleaned (however I think all but one col_name is in prepare_locations)
23 col names in estimate_jobs (I think only 7 are 'new' columns

so these 3 column lists with 89 column names could be reduced to one list of 52 names - the initial list in prepare_locations plus the additional 1 in prepare_locations_cleaned and 7 in estimate_jobs

garypolzin (approved):

PR #233, opened 2024-01-23
use legacy parquet date write mode for spark utils session

james-francis-MT:
# Description
Adding in some spark config and using the general spark session in utils

# How to test
Ran the glue job manually to check against a data file that was previously failing to ingest
Will also run this job against all past worker files to check there are no other errors

# Developer checklist
[Trello ticket](https://trello.com/c/lW1CVrNh/117-add-datetimerebasemodeinwritelegacy-config-to-the-ascwds-worker-glue-job)


garypolzin (approved):
I'll create another ticket off the back of this one to actually deal with this dodgy imports but thanks for get the job running again!

PR #243, opened 2024-01-29
Add column with main service

james-francis-MT:
# Description
Add the main service column to the cqc locations dataset when cleaning

# How to test
Unit tests have been added but the job should be ran after merging to main

[Trello ticket](https://trello.com/c/IgRoV5Im/25-epic-4-add-column-with-main-service-must)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/clean_cqc_location_data.py
```diff
@@ -4,13 +4,24 @@
...
+    CqcLocationApiColumns,
+)
+from pyspark.sql.dataframe import DataFrame
+import pyspark.sql.functions as F
 
 cqcPartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
 
+NURSING_HOME_IDENTIFIER = "Care home with nursing"
+NONE_NURSING_HOME_IDENTIFIER = "Care home without nursing"
+NONE_RESIDENTIAL_IDENTIFIER = "non-residential"
```
garypolzin, line 17: might be worth moving these 3 into a generic place (TBC), they are called in a lot of places in the pipeline

james-francis-MT: I can only see these used in prepare locations at the moment so maybe it is worth moving them when we have more use cases

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -29,12 +31,35 @@ def test_main_runs(self, read_from_parquet_patch, write_to_parquet_patch):
         read_from_parquet_patch.return_value = self.test_clean_cqc_location_df
         job.main(self.TEST_SOURCE, self.TEST_DESTINATION)
         write_to_parquet_patch.assert_called_once_with(
-            self.test_clean_cqc_location_df,
+            ANY,
```
MarionJHolloway, line 34: Was there a reason for this change?

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -29,12 +31,35 @@ def test_main_runs(self, read_from_parquet_patch, write_to_parquet_patch):
...
             self.TEST_DESTINATION,
             append=True,
             partitionKeys=self.partition_keys,
         )
 
+    def test_allocate_primary_service_type(self):
+        PRIMARY_SERVICE_TYPE_COLUMN_NAME = "primary_service_type"
+
+        test_primary_service_df = self.spark.createDataFrame(
```
MarionJHolloway, line 43: The test for main created the data frame in the setup step - we probably want to decide when we're going to create this for consistancy. I don't think either method really impacts our test speed much.

jobs/clean_cqc_location_data.py
```diff
@@ -19,6 +30,27 @@ def main(cqc_location_source: str, cleaned_cqc_location_destintion: str):
     )
 
 
+def allocate_primary_service_type(df: DataFrame):
+    return df.withColumn(
+        "primary_service_type",
```
MarionJHolloway, line 35: Can you abstract this please?

MarionJHolloway (commented):
Looks good, just a few small queries :)

tests/unit/test_clean_cqc_location_data.py
james-francis-MT: Now the dataframe is being altered it doesn't match the one that is read. Not sure how else to write this expect

tests/unit/test_clean_cqc_location_data.py
james-francis-MT: That dataframe has so much in - when adding new test rows it would have been pretty hard so I decided to make a new test one. I'm not sure which is best though

tests/unit/test_clean_cqc_location_data.py
MarionJHolloway: Ahh that makes sense, thanks :)

tests/unit/test_clean_cqc_location_data.py
MarionJHolloway: I guess this dataframe isn't likely to be used elsewhere in this test script. Maybe we can agree that anything that is used more than once can be created in the setup, else in the test itself is fine?

tests/test_file_schemas.py
```diff
@@ -187,7 +192,6 @@ class ASCWDSWorkerSchemas:
     )
 
 
-@dataclass
```
MarionJHolloway, line 190: Was this deletion intentional?

MarionJHolloway (approved):
Just one line to check, happy for you to merge if it was removed deliberately :)

tests/test_file_schemas.py
james-francis-MT: Ahh good spot - happened when merging main in

PR #244, opened 2024-01-30
Create universal function for labels

MarionJHolloway:
# Description
Creating a universal function for applying labels to data as part of cleaning data files. The function should work for categorical variables in any of our 6 raw data files.

https://trello.com/c/AQoO2Fej/104-epic-create-a-universal-function-for-adding-labels-for-categorical-variables-must 

# How to test
Unit tests are passing.
It is not possible to test in branch yet - we do not have the labels coded up yet.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


tests/unit/test_cleaning_utils.py
```diff
@@ -0,0 +1,383 @@
...
+            self.expected_categorical_labels[nationality_labels][3],
+        )
+        self.assertEqual(
+            returned_data[4][nationality_labels],
+            self.expected_categorical_labels[nationality_labels][4],
+        )
+        self.assertEqual(
+            returned_data[5][nationality_labels],
+            self.expected_categorical_labels[nationality_labels][5],
+        )
```
garypolzin, line 142: is another option for tests is to also make the finished/expected dataframe?
then we only need one assertion to say does the starting df run through the process match the expected df?

I know it's our standard approach but I find that testing every cell value quite difficult to read/change if required (works ok for smaller dfs with only a few values to test)

garypolzin (commented):
haven't been able to give this a proper review but just scanned through quickly, looks great :)

tests/unit/test_cleaning_utils.py
MarionJHolloway: Let me see if I can :)

tests/test_file_data.py
```diff
@@ -565,3 +565,35 @@ class CQCLocationsData:
             ],
         ),
     ]
+
```
RoyPriceCS50, line 568: This line is blank!

RoyPriceCS50 (commented):
This is good.

james-francis-MT (approved):
Nice

garypolzin (approved):
Yep like this approach, test scripts are much more pleasant and it's nice having the before & after dfs next to each other

PR #248, opened 2024-02-02
Create universal scale cleaning function

james-francis-MT:
# Description
Create a function for bounding scale values in dataframes

# How to test
Unit tests have been added - not used in prod code yet. Can test when we use it

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/TscsiZ1H/94-epic-create-a-universal-function-for-cleaning-scale-variables-must

### DIFF
--- a/tests/unit/test_cleaning_utils.py
+++ b/tests/unit/test_cleaning_utils.py
@@ -232,6 +232,28 @@ class TestCleaningUtilsScale(unittest.TestCase):
         original_float_values = self.test_scale_df.select("float").collect()
         self.assertEqual(returned_float_values, original_float_values)
 
+    def test_set_column_bounds_int_value_equals_bound(self):
+        returned_df = job.set_column_bounds(
+            self.test_scale_df, "int", "bound_int", 23, 23
+        )
+
+        returned_bounded_int = (
+            returned_df.where(F.col("int") == 23).select("bound_int").first()[0]
+        )
+        self.assertEqual(returned_bounded_int, 23)
+
+    def test_set_column_bounds_float_value_equals_bound(self):
+        returned_df = job.set_column_bounds(
+            self.test_scale_df, "float", "bound_float", 10.1, 10.2
+        )
+
+        returned_bounded_int = (
+            returned_df.where(F.round(F.col("float")) == 10)
+            .select("bound_float")
+            .first()[0]
+        )
+        self.assertAlmostEqual(returned_bounded_int, 10.1, 3)
+
     def test_set_column_bounds_int_below_lower_bound_are_set_to_null(self):
         returned_df = job.set_column_bounds(
             self.test_scale_df, "int", "bound_int", lower_limit=25, upper_limit=100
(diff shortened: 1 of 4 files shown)


utils/cleaning_utils.py
```diff
@@ -59,3 +60,51 @@ def convert_labels_dict_to_dataframe(labels: dict, spark: SparkSession) -> DataF
...
+):
+    if lower_limit is None and upper_limit is None:
+        return df
+
+    if lower_limit > upper_limit:
+        raise Exception(
+            f"Lower limit ({lower_limit}) must be lower than upper limit ({upper_limit})"
+        )
+
+    new_df = df.alias("df_with_column_bound")
```
james-francis-MT, line 76: After some googling I think pyspark dataframes are immutable and so this isn't needed

tests/unit/test_cleaning_utils.py
```diff
@@ -209,3 +205,152 @@ def test_replace_labels_replaces_values_in_new_column_when_new_column_name_is_su
...
+            self.test_scale_df, "int", "bound_int", lower_limit=25, upper_limit=100
+        )
+
+        returned_bounded_int = (
+            returned_df.where(F.col("int") == 23).select("bound_int").first()[0]
+        )
+
+        self.assertEqual(returned_bounded_int, None)
+
+    def test_set_column_bounds_float_below_lower_bound_are_set_to_null(self):
```
MarionJHolloway, line 246: Could you tweak the test names so it's clear which behaviour we want from the limit value? Or alternatively add a test for how that value behaves if you prefer.

MarionJHolloway (changes requested):

tests/unit/test_cleaning_utils.py
james-francis-MT: I can add a test that checks behaviour when the value is equal to the limit?

MarionJHolloway: Yup, that would be a simpler way to write it! Thanks.

james-francis-MT: Hmm, floats are a bit awkward as 10.1 is actually not 10.1 - e.g:
`10.100000381469727 != 10.1`

This means that bounds may not be fully accurate for floats. What do you think is the best way to deal with this?

MarionJHolloway: "set spark.sql.legacy.timeParserPolicy=LEGACY"

james-francis-MT: Yeah that works for testing but when applying an upper bound that is equal to a float it may be set to null

MarionJHolloway: Oh I see what you mean. Lets put it through for now, and we can check in with Will and Gary if that will be a problem on Monday. I doubt it will because our boundaries will probably be a lot higher than any valid values in our data at this point in the pipeline, but they will be able to confirm.

MarionJHolloway (approved):
Looks good, thanks!

PR #253, opened 2024-02-08
Do not clean cqc location rows already cleaned

james-francis-MT:
# Description
There is no need to re clean and append data that has already been cleaned - this change makes sure that we filter out data that has already been cleaned beforehand

# How to test
This has been tested on branch by running this multiple times with the same data source and checking the output

[Trello ticket](https://trello.com/c/sZzUXKgK/144-epic-4-set-up-cleaning-script-so-it-does-not-process-rows-already-cleaned-and-only-append-new-rows-must)

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


tests/unit/test_utils.py
```diff
@@ -686,6 +690,80 @@ def test_get_latest_partition_uses_correct_partition_keys(self):
...
+            "Input dataframe must have import_date column" in str(context.exception),
+        )
+
+    @patch("utils.utils.read_from_parquet")
+    def test_remove_already_cleaned_data_returns_the_same_df_if_there_is_no_cleaned_data(
+        self, read_from_parquet_mock: Mock
+    ):
+        read_from_parquet_mock.return_value = self.spark.createDataFrame(
+            [],
+            FilterCleanedValuesSchema.sample_schema,
```
MarionJHolloway, line 714: If the script has never been run, would a schema exist in s3? I'm not sure what s3 would return if the dataframe has never existed.

MarionJHolloway (approved):
Looks good, just one question in the comments - not sure if it will need addressing or not so have approved these changes

tests/unit/test_utils.py
james-francis-MT: Ahh yeah this test is quite confusing - I'll update it if I can to throw the same error that you get when this happens in s3

james-francis-MT: I think I can remove this test as it is tested more accurately below in the final test

PR #255, opened 2024-02-08
Split locations file into registered and deregistered

MarionJHolloway:
# Description
Split the locations file into registered and deregistered locations

https://trello.com/c/A6GwGtyQ/37-epic-4-split-file-based-on-registered-or-de-registered-locations-must

# How to test
Unit tests passing
Ran successfully in branch. No deregistered locations in athena with sample data.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -1,4 +1,5 @@
 import sys
+import warnings
 
 from utils import utils
 
@@ -98,19 +99,22 @@ def join_cqc_provider_data(locations_df: DataFrame, provider_df: DataFrame):
 def split_dataframe_into_registered_and_deregistered_rows(
     locations_df: DataFrame,
 ) -> DataFrame:
+    invalid_rows = locations_df.where(
+        (locations_df[CQCL.registration_status] != "Registered") & (locations_df[CQCL.registration_status] != "Deregistered")
+    ).count()
+    
+    if invalid_rows != 0:
+        warnings.warn(
+            f"{invalid_rows} row(s) has/have an invalid registration status and have been dropped."
+        )
+
     registered_df = locations_df.where(
         locations_df[CQCL.registration_status] == "Registered"
     )
     deregistered_df = locations_df.where(
         locations_df[CQCL.registration_status] == "Deregistered"
     )
-    rows_without_registration_status = locations_df.where(
-        locations_df[CQCL.registration_status].isNull()
-    ).count()
-    if rows_without_registration_status != 0:
-        print(
-            f"{rows_without_registration_status} rows are missing a registration status and have been dropped."
-        )
+    
     return registered_df, deregistered_df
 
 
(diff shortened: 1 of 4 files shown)


jobs/clean_cqc_location_data.py
```diff
@@ -90,6 +95,25 @@ def join_cqc_provider_data(locations_df: DataFrame, provider_df: DataFrame):
...
+    registered_df = locations_df.where(
+        locations_df[CQCL.registration_status] == "Registered"
+    )
+    deregistered_df = locations_df.where(
+        locations_df[CQCL.registration_status] == "Deregistered"
+    )
+    rows_without_registration_status = locations_df.where(
+        locations_df[CQCL.registration_status].isNull()
+    ).count()
+    if rows_without_registration_status != 0:
```
RoyPriceCS50, line 110: Is it better to catch rows where status is not "Registered" and not "Deregistered"?

If CQC add a new category it wont be captured here.

MarionJHolloway: Good question. I went with this method because currently there are empty rows in this column. If a new category appeared, we'd probably want to update the code to handle it anyway. I included the print statement in case we end up with missing data at some point in the future. I could also change it to raise an error, if we want to know as soon as that happens. What do you think is best?

RoyPriceCS50: The print statement won't pick up new cateogries at the moment, so we wont know when a new catgory has appeared.

I think having where (null or (not equal registered & not equal deregistered) captures all the eventualities.

Can the script complete with an error raised?  When registered/not registered first appeared in CQC, the weighting had to be re-done, so i think raising an error makes it more obvious a problem has occured that needs addressing.
Print statement could be missed.

MarionJHolloway: I can raise an error that aborts the script if there are rows that are not registered or deregistered and reports how many? Is that what you mean?

RoyPriceCS50: I meant let the script complete with an error raise.

I don't know how error's are logged within the pipeline. Whether raised errors are logged for review later, or print statements are logged somewhere.

I just think a terminal print statement could be overlooked.

james-francis-MT: What would we do with the new category? Is it worth thinking about this now or can we create a ticket to decide what behaviour we want when / if this happens? I don't think it should block this PR from being merged

james-francis-MT (approved):

jobs/clean_cqc_location_data.py
james-francis-MT: Can this be moved to a new ticket so we can get this one over the line - seems like something that we would want to decide for every case like this rather than specifically this one

MarionJHolloway: Update pushed and tested in branch.

PR #257, opened 2024-02-09
format cqc location date fields

james-francis-MT:
# Description
Formatting date fields in cqc location data

# How to test
Can be tested in main once merged

[Trello ticket](https://trello.com/c/sL59mLvc/138-epic-4-format-date-fields-must)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -30,6 +30,8 @@ NURSING_HOME_IDENTIFIER = "Care home with nursing"
 NONE_NURSING_HOME_IDENTIFIER = "Care home without nursing"
 NONE_RESIDENTIAL_IDENTIFIER = "non-residential"
 
+DATE_COLUMN_IDENTIFIER = "registration_date"
+
 
 def main(
     cqc_location_source: str,
@@ -42,6 +44,12 @@ def main(
     cqc_location_df = utils.remove_already_cleaned_data(
         cqc_location_df,
         cleaned_cqc_location_destintion,
+    )
+
+    cqc_location_df = utils.format_date_fields(
+        cqc_location_df,
+        date_column_identifier=DATE_COLUMN_IDENTIFIER,
+        raw_date_format="yyyy-MM-dd",
     )
 
     cqc_location_df = remove_invalid_postcodes(cqc_location_df)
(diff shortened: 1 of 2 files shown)


jobs/clean_cqc_location_data.py
```diff
@@ -30,6 +30,8 @@
 NONE_NURSING_HOME_IDENTIFIER = "Care home without nursing"
 NONE_RESIDENTIAL_IDENTIFIER = "non-residential"
 
+DATE_COLUMN_IDENTIFIER = "registration_date"
```
MarionJHolloway, line 33: There are multiple date columns in this dataset that will want formatting. If you look at the function definition, you can omit this parameter, as is will run on all columns with date in the column name.

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -55,6 +57,7 @@ def test_main_runs(
 
         self.assertEqual(remove_already_cleaned_data_patch.call_count, 1)
         self.assertEqual(read_from_parquet_patch.call_count, 2)
+        format_date_fields_mock.assert_called_once()
```
MarionJHolloway, line 60: Will probably need to update this too - maybe to check the call count?

MarionJHolloway (approved):
Just a couple of small changes to make before merging.

jobs/clean_cqc_location_data.py
james-francis-MT: I had a look at the schema and the only date columns either had this in the name or were 'import_date'. This behaviour is what currently happens in prepare_locations. Is it okay to also attempt to convert import date to a date type?

tests/unit/test_clean_cqc_location_data.py
james-francis-MT: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once This checks that the function was called exactly once

MarionJHolloway: Sorry, I was thinking it was called once for each date column - Friday brain!

jobs/clean_cqc_location_data.py
MarionJHolloway: It might be worth having a chat about this one because it will impact the date alignment too

PR #260, opened 2024-02-13
Format import date in cleaning scripts

james-francis-MT:
# Description
In each cleaning script, add a new column with the formatted import date

# How to test
Can be tested alongside https://trello.com/c/eTLzu2PQ/129-epic-create-a-generic-date-matrix-function-for-selecting-which-importdates-should-be-used-to-join-datasets-together-must

[Trello ticket](https://trello.com/c/Pj3tYAqr/175-create-new-column-with-a-formatted-import-date)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -285,6 +285,7 @@ class CQCProviderData:
             "Adult social care",
             "Southampton, Itchen",
             "Southampton",
+            "20230405",
         ),
         (
             "1-10000000002",
@@ -309,6 +310,7 @@ class CQCProviderData:
             "Adult social care",
             "Southampton, Itchen",
             "Southampton",
+            "20230405",
         ),
         (
             "1-10000000003",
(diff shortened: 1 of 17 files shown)


tests/test_file_data.py
```diff
@@ -347,6 +347,7 @@ class CQCProviderData:
             "Adult social care",
             "Southampton, Itchen",
             "Southampton",
+            "20230405",
```
MarionJHolloway, line 350: Do we need still need these changes if it's being tested in the cleaning utils?

james-francis-MT: I would have to mock out the col_to_date function if not as the main function attempts to use this column - I tried to avoid adding another @patch decorator as I think having to many is confusing

MarionJHolloway (approved):
Thanks for getting this sorted!

PR #265, opened 2024-02-15
Purge outdated workplace data

james-francis-MT:
# Description
Purge outdated ascwds workplace data

# How to test
Tested in branch and checked the processed data - new column was added and after checking a few of the rows with "purge_data=true" it seems to be doing what it should be 

Feel free to double check here: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/purge-outdated-workplace-data-clean_ascwds_workplace_data_job/runs

[Bucket with cleaned data](https://s3.console.aws.amazon.com/s3/buckets/sfc-purge-outdated-workplace-data-datasets?region=eu-west-2&bucketType=general&prefix=domain%3DASCWDS/dataset%3Dworkplace_cleaned/year%3D2024/month%3D02/day%3D15/import_date%3D20240215/&showversions=false)

[Ticket](https://trello.com/c/MyES0K1r/45-epic-5-purge-and-tidy-up-file-for-pipeline-using-update-date-must)
# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


jobs/clean_ascwds_workplace_data.py
```diff
@@ -18,13 +20,19 @@ def main(source: str, destination: str):
     ascwds_workplace_df = utils.read_from_parquet(source)
 
     ascwds_workplace_df = cUtils.column_to_date(
-        ascwds_workplace_df, PartitionKeys.import_date, AWPClean.import_date
```
james-francis-MT, line 21: The wrong string was being used here so corrected

jobs/clean_ascwds_workplace_data.py
```diff
@@ -45,6 +53,32 @@ def cast_to_int(df: DataFrame, column_names: list) -> DataFrame:
     return df
 
 
+def purge_outdated_workplaces(df: DataFrame, comparison_date_col: str) -> DataFrame:
```
james-francis-MT, line 56: Not sure how to make this any more readable - suggestions welcome

MarionJHolloway: You could abstract some of these steps into small functions so you can describe what they're doing?

MarionJHolloway (changes requested):

jobs/clean_ascwds_workplace_data.py
```diff
@@ -25,6 +27,10 @@ def main(source: str, destination: str):
         ascwds_workplace_df, [AWP.total_staff, AWP.worker_records]
     )
 
+    ascwds_workplace_df = purge_outdated_workplaces(
+        ascwds_workplace_df, "ascwds_workplace_import_date"
```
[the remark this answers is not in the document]
james-francis-MT: Ahh yes, I forgot to change after merging main - nice spot

jobs/clean_ascwds_workplace_data.py
james-francis-MT: Yeah I can try that but they will be very specific to this function - I guess that's okay though

MarionJHolloway: Yeah, I think that's fine, if it makes it clearer what it's doing

james-francis-MT:
@MarionJHolloway What do you think of the test structure? This allows you to run individual tests for a function but still keep the general setUp stuff

jobs/clean_ascwds_workplace_data.py
james-francis-MT: I've extracted out some of the functionality - hope that makes it a bit more readable

jobs/clean_ascwds_workplace_data.py
```diff
@@ -45,6 +53,52 @@ def cast_to_int(df: DataFrame, column_names: list) -> DataFrame:
...
+        ),
+    )
+
+    df_with_latest_update = calculate_latest_update_to_workplace_location(
+        df_with_purge_date, comparison_date_col
+    )
+
+    df_with_purge_data = df_with_latest_update.withColumn(
+        "purge_data", F.col("latest_update") < F.col("purge_date")
+    )
```
garypolzin, line 74: please can you change so that
if F.col("latest_update") < F.col("purge_date") is True then "Purge"
otherwise "keep"

easier than having to check/remember if True means that the data is good or if True means delete them!

garypolzin (changes requested):

jobs/clean_ascwds_workplace_data.py
james-francis-MT: Sure, in my head if( purge_data ){ purge_data() } made sense but happy either way

garypolzin: yea, there's another purge which we do (but isn't part of this project) so we'll have 2 purge columns

it definitely does make sense, but I just know I'll keep looking at it thinking is it right that I'm seeing the True's or should they be ones which we need to be removed

keep=keep and purge=delete is better for my tired brain!

MarionJHolloway (approved):
Looks good to me

garypolzin (approved):
looks great thanks!

PR #268, opened 2024-02-16
Creating value labels class ascwds

luke-a-williams:
# Description
Roy and I paired on this ticket in order to create the first standard for storing label data in a data class.
We also upgraded functions that would directly use this new structure.

Agreed upon the list of tuples structure, as discussed with the wider team

Ticket: https://trello.com/c/HsKEfAef/99-epic-6-create-data-structure-in-utils-to-hold-label-codes-categorical-variables-must

# How to test

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


utils/value_labels/ascwds_worker_mainjrid.py
```diff
@@ -0,0 +1,49 @@
...
+        (36, "Occupational therapist assistant"),
+        (37, "Registered Nursing Associate"),
+        (38, "Nursing Assistant"),
+        (39, "Assessment officer"),
+        (40, "Care co-ordinator"),
+        (41, "Care navigator"),
+        (42, "Any Childrens/young peoples job role"),
+        (43, "Deputy manager"),
+        (44, "Learning and development lead"),
+        (45, "Team leader"),
```
garypolzin, line 48: I think these numerical values should be strings (we don't have a schema for ASCWDS data)

garypolzin (changes requested):
looks good! Let me know if my comment isn't correct and I'll approve

utils/value_labels/ascwds_worker_mainjrid.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@dataclass
+class AscwdsWorkerValueLabelsMainjrid:
+    """The possible values of the mainjrid column in ascwds worker data"""
+
+    main_job_role_id: str = AWK.main_job_role_id
+    main_job_role_id_labeled: str = AWK.main_job_role_id + "_labeled"
+
+    labels_list = [
```
james-francis-MT, line 15: What is the advantage of this data type over something simple like a dictionary?

utils/value_labels/ascwds_worker_mainjrid.py
james-francis-MT: Personally feel it should match the type of the actual data - so in this case a string

garypolzin: yea that's fine, go ahead as it is for now

garypolzin (approved):

utils/cleaning_utils.py
```diff
@@ -22,7 +22,7 @@ def apply_categorical_labels(
     add_as_new_column: bool = True,
 ) -> DataFrame:
     for column_name in column_names:
-        labels_df = convert_labels_dict_to_dataframe(labels[column_name], spark)
+        labels_df = convert_labels_to_dataframe(labels[column_name], spark)
```
james-francis-MT, line 25: How does this line work with the current label data type?

You have to create a dict before hand? e.g - 
`{'column_name': AscwdsWorkerValueLabelsMainjrid.labels_list}`

luke-a-williams: So I didn't build this other function, which is also untested, this was just refactoring all instances of the function that will be directly using this new data class. I think this is worth raising, because this other function isn't being used anywhere last I checked.
But ultimately, the function you are talking about is left over from some refactoring Marion did, and probably needs changing so it accepts things like `labels_list` directly, rather than calling from a dictionary.

utils/value_labels/ascwds_worker_mainjrid.py
luke-a-williams: The advantage of defining it as a list of tuples I believe was agreed to be because of how easy it is to construct spark dataframes from it. It's easy to pass a list of tuples that will correspond to new rows into the createDataFrame spark function with a schema defined as done elsewhere.

This dataframe is then what is to be used when applying categorical labels

utils/cleaning_utils.py
luke-a-williams: I.e. old dataframe with target column + key/value dataframe of conversions + join logic = new dataframe with converted columns. That's roughly how I understand this piece. Does that help?

james-francis-MT: Well previously it was built to handle a dict of dicts but as we have changed the interface of the inner function (`convert_labels_to_dataframe`) it indirectly changes the interface of the outer function (`apply_categorical_labels`) - just wondering if this was thought about when deciding on the label data structure

MarionJHolloway (approved):
The updated version looks good, happy to approve

PR #270, opened 2024-02-19
refactor write to parquet method to handle other modes

james-francis-MT:
# Description
Refactor write to parquet helper function

# How to test
Hard to test as this is used in a bunch of places - unit tests pass but coverage is not the best

We may have issues if somewhere has been missed but we will know how to fix it so can act quickly

[Trelllo ticket](https://trello.com/c/jKFUpxxR/201-general-refactor-write-to-parquet-to-have-overwrite-option)

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/jobs/clean_ascwds_worker_data.py
+++ b/jobs/clean_ascwds_worker_data.py
@@ -31,7 +31,7 @@ def main(
     utils.write_to_parquet(
         ascwds_worker_df,
         cleaned_worker_destination,
-        True,
+        "append",
         [
             PartitionKeys.year,
             PartitionKeys.month,
--- a/jobs/bulk_download_cqc_locations.py
+++ b/jobs/bulk_download_cqc_locations.py
@@ -23,7 +23,7 @@ def main(destination):
             df = locations_df
 
     df = df.dropDuplicates([ColNames.location_id])
-    utils.write_to_parquet(df, destination, True)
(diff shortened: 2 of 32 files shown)


jobs/clean_ascwds_worker_data.py
```diff
@@ -31,7 +31,7 @@ def main(
     utils.write_to_parquet(
         ascwds_worker_df,
         cleaned_worker_destination,
-        True,
+        "append",
```
MarionJHolloway, line 34: I think there's a list of scripts in the tickets that need to be overwrite instead of append for now.

MarionJHolloway (changes requested):
Could you check the ticket for which scripts need to be overwrite instead of append?

jobs/clean_ascwds_worker_data.py
james-francis-MT: Ahh I thought that was a different ticket - I was just keeping the exact same behaviour as before in this refactor

james-francis-MT: I'll merge this and then change them in a different PR

MarionJHolloway (approved):
Approved - will need a separate ticket for changing some scripts to overrwrite

PR #279, opened 2024-02-21
Amend pir carehome rules

garypolzin:
# Description
This ticket does two things:
- Amends the code previously written so only 'Community' gets a 'N' value ('Shared Lives' or anything else returns None). This is so that the services in this match link up with CQC careHome data)
- I noticed that this function wasn't actually added to main previously so it wasn't doing anything. Added it in and updated tests to account for that.

[Trello](https://trello.com/c/TOLlbnIV/273-epic-2-amend-addcarehomecolumn-function-must)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_clean_cqc_pir_data.py
+++ b/tests/unit/test_clean_cqc_pir_data.py
@@ -33,6 +33,7 @@ class CleanCQCpirDatasetTests(unittest.TestCase):
             Schemas.expected_care_home_column_schema,
         )
 
+    @patch("add_care_home_column")
     @patch("utils.cleaning_utils.column_to_date")
     @patch("utils.utils.remove_already_cleaned_data")
     @patch("utils.utils.write_to_parquet")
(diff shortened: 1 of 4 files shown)


tests/unit/test_clean_cqc_pir_data.py
```diff
@@ -113,7 +85,10 @@ def test_add_care_home_column_categorises_care_homes_correctly(self):
         expected_df = self.test_expected_care_home_column_df
         expected_data = expected_df.sort(PIRClean.location_id).collect()
 
-        self.assertCountEqual(expected_data, returned_data)
+        self.assertCountEqual(expected_data[0], returned_data[0])
+        self.assertCountEqual(expected_data[1], returned_data[1])
+        self.assertCountEqual(expected_data[2], returned_data[2])
+        self.assertCountEqual(expected_data[3], returned_data[3])
```
james-francis-MT, line 91: What was wrong with the previous assertion?

garypolzin: it was fine, but with there now being two when clauses and an otherwise, if the test failed it was just saying the whole df didn't match, so it wasn't immediately clear which of the steps was an issue

doing it at row level should narrow it down basically. But the test is fundamentally the same

luke-a-williams (approved):
GTM as discussed

PR #283, opened 2024-02-22
format ons import date in clean cqc loc script

james-francis-MT:
# Description
Format ons import date to be used for the join

# How to test
Unit tests pass. Can be tested fully here: https://trello.com/c/xVrWiYbf/59-epic-4-add-columns-with-current-geographies-must

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -30,6 +30,8 @@ NONE_RESIDENTIAL_IDENTIFIER = "non-residential"
 
 DATE_COLUMN_IDENTIFIER = "registration_date"
 
+ONS_FORMATTED_IMPORT_DATE_COL = "ons_postcode_import_date"
+
 
 def main(
     cqc_location_source: str,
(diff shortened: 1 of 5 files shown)


jobs/clean_cqc_location_data.py
```diff
@@ -30,6 +30,8 @@
 
 DATE_COLUMN_IDENTIFIER = "registration_date"
 
+ONS_FORMATTED_IMPORT_DATE_COL = "ons_postcode_import_date"
```
garypolzin, line 33: Can you add this col name to utils>column_names>raw_data_files>ons_columns
so it matches the format of row 51 (should be fine in 'raw' given we don't have a cleaning script)

garypolzin (changes requested):
Might be worth adding a test in main to assert that this function is called twice like we do with the other utils function

jobs/clean_cqc_location_data.py
james-francis-MT: Thought it felt strange being there as it isn't actually a column in the raw data but I can add it

garypolzin: yea fair point, it's easy to change later so leave it here then

garypolzin (approved):

PR #287, opened 2024-02-22
Remove duplicate entries carehome pir

luke-a-williams:
# Description
Introduced the logic required for ticket https://trello.com/c/WgXg2dlh/191-epic-2-remove-duplicates-from-pir-data-must
Split into two parts:
- The new utils function `latest_datefield_for_grouping` which in the specific case of multiple submission dates existing for a pir import date still returns duplicates, but grouped to the required grouping
- an additional drop duplicates function call after to remove said duplicate same submission id, to only return one submission id per grouping.

I also refactored some other tests in the region to be more to purpose and requiring less return value patching.

# How to test
## Notebook
I created a notebook demonstrating this functionality on the larger dataset as I needed to be careful with exactly how the duplicates were being dropped after filtering. If you need to, you can spin up the notebook [here](https://e-df1p29ja8qe2vrn13q5ydsmzg.emrnotebooks-prod.eu-west-2.amazonaws.com/workspace/doc/tree/pir_clean_dedupe.ipynb) or otherwise find it within my notebooks under `pir_clean_dedupe`
## Integration test
This run shows my latest run was successful: 
[run history](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/remove-duplicate-entries-careh-clean_cqc_pir_data_job/runs)

I've designed the function and placed it in utils so that the grouping and filtering to return the latest of a target column can be re-used elsewhere, although I'm happy to isolate it to the clean_cqc_pir area if required

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


utils/utils.py
```diff
@@ -226,3 +226,36 @@ def collect_arguments(*args):
     parsed_args, _ = parser.parse_known_args()
 
     return (vars(parsed_args)[arg[0][2:]] for arg in args)
+
+
+def latest_datefield_for_grouping(
+    df: DataFrame, grouping_column_list: list[Column], date_field_column: Column
```
luke-a-williams, line 232: This type was not sub-scriptable in the test run, so changing

tests/test_file_data.py
```diff
@@ -619,6 +619,22 @@ class CQCpirData:
...
+        ("1-1199876096", "Y", "2022-07-01", "2022-05-20"),
+        ("1-1199876096", "Y", "2023-06-01", "2023-05-12"),
+        ("1-1199876096", "Y", "2023-06-01", "2023-05-24"),
+        ("1-1199876096", "Y", "2023-06-01", "2023-05-24"),
+    ]
+    subset_for_latest_submission_date_after_filter_deduplication = [
+        ("1-1199876096", "Y", "2022-02-01", "2021-05-07"),
+        ("1-1199876096", "Y", "2022-07-01", "2022-05-20"),
+        ("1-1199876096", "Y", "2023-06-01", "2023-05-24"),
+    ]
```
garypolzin, line 635: might be good to add a duplicate row but change the careHome value from "Y" to "N" to check that isn't removed

Also to replicate raw data the 3rd column should be in the format "20220201" and the final column is the same date but as an actual date --> date(2022, 2, 1)

tests/unit/test_utils.py
```diff
@@ -686,5 +706,71 @@ def test_get_latest_partition_uses_correct_partition_keys(self):
...
+        )
+        self.assertFalse(
+            after_df.selectExpr(
+                'ANY(cqc_pir_submission_date="2023-05-12") as date_present'
+            )
+            .collect()[0]
+            .date_present
+        )
+        # No other removes are removed
+        self.assertEqual(after_df.count(), 4)
```
garypolzin, line 772: Can I just double check the functionality - is this 4 rows at this point as the exact duplicate row keeps both records (as both are most recent)

then the other function removes one of these (rank=1) so that the final dataset correctly has 3 rows

garypolzin (changes requested):

tests/unit/test_utils.py
luke-a-williams: To answer your question with that knowledge, it is 4 rows because the duplicate submission date isn't eliminated at this point, and the test case was picked deliberately to show this. We could argue we need a test case to show that submission date appears twice, to be fair.

tests/test_file_data.py
luke-a-williams: We should probably discuss this if I'm changing data types as a lot of the logic was built around the original column being a string, and needing to be converted to a date.

tests/unit/test_utils.py
garypolzin: Yea that's what I thought thanks, just wanted to double check!

tests/test_file_data.py
garypolzin: yes that's correct, they are strings initially but the functions in main on row 21 and 24 convert the original strings to dates

Your function starts after these two functions on row 33 so at that point the two columns are dates, not strings

garypolzin (approved):

PR #294, opened 2024-02-27
Update read from parquet to only select some columns

garypolzin:
# Description
Some of our datasets are huge and we're now getting to the point where we're merging data.
This function will allow us to slim data frames down ready for merging without affecting any current functionality

Not entering an option means that the original functionality will occur, so this won't affect any existing data/jobs

[Trello](https://trello.com/c/boX0ryEA/284-general-change-readfromparquet-function-to-have-optional-column-list-must)

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -478,6 +478,11 @@ class UtilsTests(unittest.TestCase):
         self.assertEqual(df.columns, ["col_a", "col_b", "col_c"])
         self.assertEqual(df.count(), 3)
 
+    def test_read_from_parquet_imports_all_rows(self):
+        df = utils.read_from_parquet(self.example_parquet_path)
+
+        self.assertEqual(df.count(), 2270)
+
     def test_read_from_parquet_imports_all_columns_when_column_list_is_None(self):
         df = utils.read_from_parquet(self.example_parquet_path)
 
@@ -508,7 +513,6 @@ class UtilsTests(unittest.TestCase):
                 CQCColNames.uprn,
             ],
         )
-        self.assertEqual(df.count(), 2270)
 
     def test_read_from_parquet_only_imports_selected_columns(self):
         column_list = [
@@ -529,7 +533,6 @@ class UtilsTests(unittest.TestCase):
                 CQCColNames.registration_status,
             ],
         )
-        self.assertEqual(df.count(), 2270)
 
     def test_write(self):
         df = utils.read_csv(self.test_csv_path)
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -478,7 +478,7 @@ class UtilsTests(unittest.TestCase):
         self.assertEqual(df.columns, ["col_a", "col_b", "col_c"])
         self.assertEqual(df.count(), 3)
 
-    def test_read_from_parquet(self):
+    def test_read_from_parquet_imports_all_columns_when_column_list_is_None(self):
         df = utils.read_from_parquet(self.example_parquet_path)
 
         self.assertCountEqual(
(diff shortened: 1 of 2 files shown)


utils/utils.py
```diff
@@ -90,14 +90,25 @@ def generate_s3_datasets_dir_date_path(destination_prefix, domain, dataset, date
     return output_dir
 
 
-def read_from_parquet(data_source) -> pyspark.sql.DataFrame:
+def read_from_parquet(
+    data_source: str, selected_columns: list[str] = None
```
luke-a-williams, line 94: Does this work? Have you tried running it? I have a feeling it won't recognise `list[str]`... should this be `List[str]`? Please double check this and confirm

tests/unit/test_utils.py
```diff
@@ -510,6 +510,27 @@ def test_read_from_parquet(self):
...
+
+        self.assertCountEqual(
+            df.columns,
+            [
+                CQCColNames.provider_id,
+                CQCColNames.name,
+                CQCColNames.registration_status,
+            ],
+        )
+        self.assertEqual(df.count(), 2270)
```
luke-a-williams, line 532: I noticed this df.count() is the same as the one where all columns are considered, how is this possible?
count() normally counts the number of non-NaN entries in each cell of each row, and there will be considerably less data if less columns are present? Or am I missing something?

utils/utils.py
luke-a-williams: I had trouble with something similar in [this commit](https://github.com/NMDSdevopsServiceAdm/DataEngineering/pull/287/commits/8c392f67cd5ecb05a24a9a7b738808c96d9130e3) (which removes it)

luke-a-williams: Checked it myself, it's fine, disregard.

tests/unit/test_utils.py
garypolzin: df.count in pyspark code is a count of rows, so it shouldn't be affected by the number of columns

utils/utils.py
luke-a-williams: That's what I thought James, but I've just run that code locally and it hasn't complained with str inside it

tests/unit/test_utils.py
luke-a-williams: OK so there is still 2270 rows, so how is that any different to the previous test with more columns?

luke-a-williams: Like, I could put any number of columns in and I would always get this same count.

luke-a-williams: (so I don't think this line tests anything as it's independent of the number of columns). The previous assertion showing the column count is enough?)

utils/utils.py
james-francis-MT: I thought it gave no extra info but I also just tested it locally and it does help e.g
`a: list[str] = ['a', 'b', 'c']`
the type of `a` is `list[str]` as expected
but also if I do `b = a[0]`
the type of `b` is `str` which is pretty nice

james-francis-MT: So it is quite nice to put the type there if you can guarantee that all elements of the list are strings

james-francis-MT (approved):

utils/utils.py
garypolzin: just catching up on all these conversations after lunch!
is this ok as it is or does it need (or would you advise) changing it?

tests/unit/test_utils.py
garypolzin: yea I guess so, just thought I'd repeat the row count length in case my function changed it at all

should probably be 3 separate checks to be fair

one to check all columns are present
one to check some columns are present
separate check for row count

garypolzin: I've done that @luke-a-williams, there are 3 tests now

utils/utils.py
luke-a-williams: Its fine as is I think is the short summary. We didn't realise how much additional power this list[str] gives. It seems to work best with primitives though, noted going forward.

luke-a-williams (approved):

PR #295, opened 2024-02-27
Apply job role labels to workers

MarionJHolloway:
# Description
https://trello.com/c/AGBViNOT/48-epic-6-apply-labels-to-ascwds-worker-file-for-mainjrid-job-roles-must

To complete this ticket we had to 

*Change the function that applies the labels to the dataframe
*Limit the datasets to only the relevant columns
*Update tests and test data
*Change the structure of the labels in utils

# How to test
Unit tests passing
Runs successfully in branch on full dataset

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


terraform/pipeline/glue.tf
```diff
@@ -107,12 +107,14 @@ module "ingest_ascwds_dataset_job" {
...
-  script_name     = "clean_ascwds_worker_data.py"
-  glue_role       = aws_iam_role.sfc_glue_service_iam_role
-  resource_bucket = module.pipeline_resources
-  datasets_bucket = module.datasets_bucket
-  glue_version    = "3.0"
+  source            = "../modules/glue-job"
+  script_name       = "clean_ascwds_worker_data.py"
+  glue_role         = aws_iam_role.sfc_glue_service_iam_role
+  worker_type       = "G.2X"
+  number_of_workers = 6
```
luke-a-williams, line 114: Do we need this level of power now? Can it be reduced?

utils/cleaning_utils.py
```diff
@@ -18,64 +13,27 @@
...
-    spark: SparkSession,
     labels: dict,
     column_names: list,
     add_as_new_column: bool = True,
 ) -> DataFrame:
     for column_name in column_names:
-        labels_df = convert_labels_to_dataframe(labels[column_name], spark)
+        labels_dict = labels[column_name]
+        print(labels_dict)
+        print(column_name)
```
luke-a-williams, line 23: Can these be removed?

utils/cleaning_utils.py
```diff
@@ -18,64 +13,27 @@
...
         if add_as_new_column == True:
             new_column_name = column_name + "_labels"
-            df = replace_labels(df, labels_df, column_name, new_column_name)
+            print(new_column_name)
+            df.show()
+            df = df.withColumn(new_column_name, F.col(column_name))
+            df.show()
+            df = df.replace(labels_dict, subset=new_column_name)
+            df.show()
```
luke-a-williams, line 31: can shows and prints be removed?

utils/utils.py
```diff
@@ -91,7 +91,7 @@ def generate_s3_datasets_dir_date_path(destination_prefix, domain, dataset, date
 
 
 def read_from_parquet(
-    data_source: str, selected_columns: list[str] = None
+    data_source: str, selected_columns: list = None
```
luke-a-williams, line 94: Why are we changing this back? Has something changed which breaks this?

utils/value_labels/ascwds_worker/worker_label_dictionary.py
```diff
@@ -3,6 +3,6 @@
 )
 
 
-ascwds_worker_label_dict = {
-    AscwdsWorkerValueLabelsMainjrid.column_name: AscwdsWorkerValueLabelsMainjrid.labels_list
+ascwds_worker_labels_dict = {
+    AscwdsWorkerValueLabelsMainjrid.column_name: AscwdsWorkerValueLabelsMainjrid.labels_dict
```
luke-a-williams, line 7: OK I see what you've done here, making it a dictionary with a nested dictionary you can call later, nice!

luke-a-williams (changes requested):
Couple of terraform and print statement changes

utils/utils.py
luke-a-williams: James discovered that in python 3.7, the version glue uses, this doesn't work. Whilst in python 3.9, the version our project is written in, this does work. This is why it looks ok locally but falls over in Glue jobs

james-francis-MT (approved):

tests/unit/test_cleaning_utils.py
```diff
@@ -176,45 +167,13 @@ def test_apply_categorical_labels_adds_new_columns_with_replaced_values_when_new
     ):
         returned_df = job.apply_categorical_labels(
             self.test_worker_df,
-            self.spark,
             self.label_dict,
             [AWK.gender, AWK.nationality],
         )
         returned_data = returned_df.sort(AWK.worker_id).collect()
         expected_data = self.expected_df_with_new_columns.sort(AWK.worker_id).collect()
         self.assertEqual(returned_data, expected_data)
```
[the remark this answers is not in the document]
MarionJHolloway: Yes the ones removed were for functions no longer needed and didn't cover testing of functions that remain

tests/unit/test_cleaning_utils.py
```diff
@@ -176,45 +167,13 @@ def test_apply_categorical_labels_adds_new_columns_with_replaced_values_when_new
...
-            new_column_name=gender_labels,
-        )
-        returned_data = returned_df.sort(AWK.worker_id).collect()
-        expected_df = self.spark.createDataFrame(
-            Data.expected_rows_replace_labels_with_new_column,
-            Schemas.expected_schema_replace_labels_with_new_columns,
-        )
-        expected_data = expected_df.sort(AWK.worker_id).collect()
-        self.assertEqual(returned_data, expected_data)
-
```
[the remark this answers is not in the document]
MarionJHolloway: Yes, have removed these now - good spot!

tests/unit/test_cleaning_utils.py
```diff
@@ -387,30 +346,6 @@ def test_set_bounds_for_columns_set_bounds_for_all_columns(self):
...
-        self.assertEqual(output_df.schema, self.sample_df.schema)
-        self.assertEqual(output_df.count(), self.sample_df.count())
-        self.assertEqual(output_df.columns, self.sample_df.columns)
-
-    def test_convert_labels_to_dataframe_function_calls(self):
-        job.convert_labels_to_dataframe(self.sample_list, self.spark_mock)
-
-        self.spark_mock.createDataFrame.assert_called_once_with(self.sample_list, ANY)
-
-
```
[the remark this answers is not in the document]
MarionJHolloway: Labels schema has gone and data.gender is needed elsewhere

utils/utils.py
MarionJHolloway: As Luke said, it works in CircleCI but it breaks in Glue. When I ran this job in branch to test before the PR, this was the step that was causing it to crash, so I fixed it.

MarionJHolloway: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/job/apply-job-role-labels-to-worke-clean_ascwds_worker_data_job/run/jr_cfe75c387ab810b6f811c5b0498a5f888a74b07afe85cdf74ad7e7d1a97e53f4

This is the run that shows the error

garypolzin: yea it was broken on the 27th when you ran that job so we changed it yesterday - hence if you pull from main it will change to 'List[str]' (capital L works, lower case l doesn't)

but it doesn't need changing again to 'list'

MarionJHolloway: Ah I see, sorry, didn't spot the capitalisation! I will add back in

garypolzin: no worries, this caused a lot of confusion yesterday!

came across another function yesterday too which works fine in our python 3.9.6 but doesn't in Glues python 3.7, might be something to think about

jobs/clean_ascwds_worker_data.py
```diff
@@ -11,18 +11,41 @@
...
+
+WORKER_COLUMNS = [
+    AWKClean.location_id,
+    AWKClean.import_date,
+    AWKClean.establishment_id,
+    AWKClean.main_job_role_id,
+    AWKClean.year,
+    AWKClean.month,
+    AWKClean.day,
+]
```
garypolzin, line 26: sorry I was meaning to move out of this script and into the same place as the column names/values - it's a short list now but will end up being 30+ long when we get round to bringing in gender etc

utils > column_names > cleaned_data_files > ascwds_worker_cleaned_values

so there would be 3 classes in there
- column names
- values
- column list for import

MarionJHolloway: Isn't the list of columns quite specific to this script? We may want different columns from the same files in different scripts depending on what get joined where?

james-francis-MT: I said the same thing in my PR 😆

garypolzin: yea it is, but I also thought that the file above was also specific to this file?

if we filter want to import a subset of the worker file in our ind_cqc_pipeline then we'd put the column list to import into
utils > column_names > ind_cqc_pipeline_columns

garypolzin: at the moment every job has a column name file, so I just assumed that it was 1-1 and the import list for that job could go in the list

either way, this cleaned ASCWDS job will end up having at least 30+ columns imported so it would be good to have something in place now to account for that list getting long

MarionJHolloway: No, the column names are specific to the dataset, not the job. This job also import the column names for the workplace file, for example, because that file is loaded here too. I'll push through what the changes would look like.

Generally, good reasons to abstract code are to prevent repetition (i.e. it's used in many places, write it once and reference it many times, like the column names) for unit testing functions in logical blocks. 

We also don't have a plan for what we are doing with the worker data yet and keeping the short list of selected columns in the job is less commiting than abstracting it. We might end up having multiple cleaned worker files, each with a subset of data, so we don't have to rerun everything in the pipeline, who knows.

MarionJHolloway: I also feel like this is a bigger decision about how we structure our code than is within the scope of this ticket. It will impact a lot of code that isn't within scope here. This ticket was about getting the labels on the worker file, selecting the columns was only necessary to stop the glue job falling over. I've certainly not considered the implications for abstracting this in all the scripts.

MarionJHolloway: I've pushed through something like what you suggested

garypolzin: yea that's what I was thinking

The end goal should be one 'clean' ascwds file - so that file can contain all the columns we pull into it plus the names of them all and any values
We'll have the other files which will contain all the categorical label names etc

then all other pipelines use this clean version as the starting point. but they'll have their own column_name file (like ind_cqc_pipeline_columns does, and that can contain the column import list for that pipeline

garypolzin (approved):
Pretty sure all my comments have been addressed so I'll approve now, thanks!

terraform/pipeline/glue.tf
```diff
@@ -107,12 +107,14 @@ module "ingest_ascwds_dataset_job" {
...
-  script_name     = "clean_ascwds_worker_data.py"
-  glue_role       = aws_iam_role.sfc_glue_service_iam_role
-  resource_bucket = module.pipeline_resources
-  datasets_bucket = module.datasets_bucket
-  glue_version    = "3.0"
+  source            = "../modules/glue-job"
+  script_name       = "clean_ascwds_worker_data.py"
+  glue_role         = aws_iam_role.sfc_glue_service_iam_role
+  worker_type       = "G.1X"
+  number_of_workers = 3
```
luke-a-williams, line 114: Did you want to keep this as 3? Or change it back to the default of 2?

luke-a-williams (changes requested):
minor thing

jobs/clean_ascwds_worker_data.py
luke-a-williams: I think the column names directory should be kept to just that - column names defined by a dataclass for a dataset.
For me, there are 3 main ideologies which I think are relevant here:

- Having a central point of truth for schema definition, otherwise known as standardising column names, at a point during or before the bronze data layer.
- Aiming for minimal code entanglement.
- Keeping code clean and maintainable.

The original column names, i.e. ascwds_worker_column names, define the original schema in the bronze layer, and this class is extended for the cleaned column names, as it's an extension of the original dataset columns. As a general rule of thumb, it should always be inherited from there, and it ensures that location is the central point of truth for those column definitions. **They should be associated with that dataset, NOT the job.**
If you start defining columns per job in an abstract script, it risks over abstraction and reduced clarity and readability of the code, as well as risks eroding the data dictionary you have per dataset.

As for code cleanliness and maintainability, if you start defining workplace variables in worker column name files, and vice versa, you create a need to keep repeating this everytime, creating all of these new lists that are specific to each select statement. As you increase the number of cleaning jobs, and selecting subsets of data, that file starts to grow unmanageable, and it's likely you will miss exact definitions that suit your purpose that have already been created, increasing risk of duplicating code. I think in this case it is an over-abstraction, when actually it is perfectly viable and cleaner to define the selected columns **within the script they are being select**. This also increases readability in glue, where your view is isolated to the job itself.

Hopefully that makes some degree of sense.

garypolzin: Yea fair enough

was just thinking about those files with 800 columns, and we might want to import 50 of them. The script then starts with 50 rows of all the columns to import, and then we're frequently merging in one/two/three other datasets too, so could easily end up with 100+rows of all the columns to import into that job.

So let's not do this approach, but I think we need an approach

luke-a-williams: Lets have a chat Gary, because your points are completely valid and I want to make sure we are on the same page and find a good solution

luke-a-williams (approved):

PR #297, opened 2024-02-28
Apply universal scale function

RoyPriceCS50:
# Description
Applied function to create a new column in ascwds workplace file where total staff and worker records are recoded to remove values below lower limit. Lower limit was set to 1 for both.

# How to test
Ran the login separelty and also checked output in Athena for correct result. Saw totalstaff/worker records orginal values change from 0 to null.

# Developer checklist
- [x ] Unit test
- [x] Linked to Trello ticket
- [x ] Documentation up to date

### DIFF
--- a/tests/unit/test_clean_ascwds_workplace_data.py
+++ b/tests/unit/test_clean_ascwds_workplace_data.py
@@ -48,6 +48,7 @@ class MainTests(IngestASCWDSWorkerDatasetTests):
     @patch(
         "jobs.clean_ascwds_workplace_data.create_column_with_repeated_values_removed"
     )
+    @patch("utils.cleaning_utils.set_column_bounds")
     @patch("utils.utils.format_date_fields", wraps=format_date_fields)
     @patch("utils.utils.write_to_parquet")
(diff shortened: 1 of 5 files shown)


RoyPriceCS50:
Link to successful pipeline run

https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/apply-universal-scale-function-clean_ascwds_workplace_data_job/runs

tests/unit/test_clean_ascwds_workplace_data.py
```diff
@@ -48,6 +48,7 @@ def setUp(self) -> None:
     @patch(
         "jobs.clean_ascwds_workplace_data.create_column_with_repeated_values_removed"
     )
+    @patch("utils.cleaning_utils.set_column_bounds")
```
james-francis-MT, line 51: Can we keep the implementation in this mock to make sure that the output works with following functions?

james-francis-MT (approved):

garypolzin (approved):
Looks good to me

PR #299, opened 2024-02-29
Import current geographies into cqc location data

james-francis-MT:
# Description
Join most current geography data into cqc location data

# How to test
Has been tested in branch and is able to join in geography data to location data

https://trello.com/c/xVrWiYbf/59-epic-4-add-current-geographies-into-cqc-locations-file-must

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -77,7 +77,7 @@ def main(
     cqc_location_df = remove_invalid_postcodes(cqc_location_df)
 
     cqc_location_df = cqc_location_df.withColumn(
-        CQCL.postcode, F.trim(F.col(CQCL.postcode))
+        CQCL.postcode, F.upper(F.regexp_replace(F.col(CQCL.postcode), " ", ""))
     )
 
     cqc_location_df = join_current_ons_postcode_data(
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -19,6 +19,7 @@ from utils.column_names.cleaned_data_files.cqc_provider_cleaned_values import (
 )
 from utils.column_names.cleaned_data_files.cqc_location_cleaned_values import (
     CqcLocationCleanedColumns as CQCLClean,
+    ons_cols_to_import,
 )
 from utils.column_names.raw_data_files.ons_columns import (
     OnsPostcodeDirectoryColumns as ONS,
@@ -45,15 +46,7 @@ def main(
     cqc_location_df = utils.read_from_parquet(cqc_location_source)
     cqc_provider_df = utils.read_from_parquet(cleaned_cqc_provider_source)
     ons_postcode_directory_df = utils.read_from_parquet(
-        ons_postcode_directory_source,
-        selected_columns=[
-            ONS.import_date,
-            ONS.cssr,
-            ONS.region,
-            ONS.icb,
-            ONS.rural_urban_indicator_2011,
-            ONS.postcode,
-        ],
+        ons_postcode_directory_source, selected_columns=ons_cols_to_import
     )
 
     cqc_location_df = utils.format_date_fields(
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -76,9 +76,7 @@ def main(
 
     cqc_location_df = remove_invalid_postcodes(cqc_location_df)
 
-    cqc_location_df = cqc_location_df.withColumn(
-        CQCL.postcode, F.upper(F.regexp_replace(F.col(CQCL.postcode), " ", ""))
-    )
+    cqc_location_df = utils.normalise_column_values(cqc_location_df, CQCL.postcode)
 
     cqc_location_df = join_current_ons_postcode_data(
         cqc_location_df, current_ons_postcode_directory_df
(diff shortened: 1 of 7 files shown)


jobs/clean_cqc_location_data.py
```diff
@@ -78,6 +99,29 @@ def main(
...
+    COLS_TO_RENAME.remove(ONS.postcode)
+
+    new_ons_col_names = [STRING_TO_PREPEND + col for col in COLS_TO_RENAME]
+
+    current_ons_df = ons_df
+
+    for i in range(len(COLS_TO_RENAME)):
+        current_ons_df = current_ons_df.withColumnRenamed(
+            COLS_TO_RENAME[i], new_ons_col_names[i]
+        )
```
james-francis-MT, line 118: Is there an easier way to rename all columns bar import_date and postcode?

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -81,6 +86,11 @@ def test_main_runs(
             partitionKeys=self.partition_keys,
         )
 
+        final_df: DataFrame = write_to_parquet_patch.call_args[0][0]
+        expected_cols = list(asdict(CQCLCleaned()).values())
+        for col in final_df.columns:
+            self.assertIn(col, expected_cols)
```
james-francis-MT, line 92: This just asserts that the cleaned dfs columns are in the dataclass for cqc cleaned - ideally I think we would check this the other way around to make sure that we have all the columns that we want but currently the cqcl cleaned dataclass contains values that aren't in the cleaned df

jobs/clean_cqc_location_data.py
garypolzin: not sure if this is any better, or even if it works, but found a similar thing online:

```
current_ons_df = current_ons_df.select([
    F.col(c).alias("`current_" + c + "`") if c not in ("import_date", "postcode") else F.col(c)
    for c in current_ons_df.columns
])
```

jobs/clean_cqc_location_data.py
```diff
@@ -53,6 +66,8 @@ def main(
         ons_postcode_directory_df, Keys.import_date, ONS_FORMATTED_IMPORT_DATE_COL
```
garypolzin, line 66: we could drop `import_date` at this point, will save hassle with the two postcode joins later on in the script (both renaming it in the "_current" script and also dropping it as a duplicate in the joins)

garypolzin (changes requested):
Looks good, some potential suggestions but welcome to disagree with any!

jobs/clean_cqc_location_data.py
```diff
@@ -41,7 +44,17 @@ def main(
...
-    ons_postcode_directory_df = utils.read_from_parquet(ons_postcode_directory_source)
+    ons_postcode_directory_df = utils.read_from_parquet(
+        ons_postcode_directory_source,
+        selected_columns=[
+            ONS.import_date,
+            ONS.cssr,
+            ONS.region,
+            ONS.icb,
+            ONS.rural_urban_indicator_2011,
+            ONS.postcode,
```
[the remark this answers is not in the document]
james-francis-MT: Does it make sense to add it to a generic file when this import list is only going to be used in this script? Other scripts may have different import lists

jobs/clean_cqc_location_data.py
james-francis-MT: Seems less readable to me but happy to change it if this way is preferred

jobs/clean_cqc_location_data.py
james-francis-MT: It is dropped in the prepare step - it is needed to get the current postcodes

jobs/clean_cqc_location_data.py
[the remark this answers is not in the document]
garypolzin: True, just thinking that this could end up being 15 columns long so it'd be quite long

I was thinking in utils > column_names > cleaned_data_files > cqc_location_cleaned_values

so there would be 3 classes
- column names
- values
- column list for import

the column list might end up with 2 in there
cqc_cols_to_import = [....] - we'll probably reduce this file size by dropping cols at some point
ons_cols_to_import = [....]

jobs/clean_cqc_location_data.py
garypolzin: yea happy for you to leave as is

jobs/clean_cqc_location_data.py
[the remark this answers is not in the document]
james-francis-MT: Okay yeah, that makes sense to me

jobs/clean_cqc_location_data.py
garypolzin: import_date is the string version of the import date
after this function we'll have ONS_FORMATTED_IMPORT_DATE_COL which is the date formatted version

the prepare step can use the max ONS_FORMATTED_IMPORT_DATE_COL instead of the max of the string representation of the date

james-francis-MT: Not sure what the advantage is of dropping it here vs in the prepare function though but have made the change

garypolzin (approved):
all looks good thanks!

PR #300, opened 2024-03-01
Amending cols to import and identifiers

garypolzin:
# Description
I talked James around into putting the ons column list to import into the cleaned columns file which was a bad choice, so I've moved them back!!

Also there were some identifiers and column references which I moved into that file too

No code changes


MarionJHolloway (approved):
Thanks, this will really help with my geographies ticket!

PR #306, opened 2024-03-04
amended postcode match

garypolzin:
# Description
Previously I replaced one postcode in Wales with an another which is also in Wales, oops. Added a new match, which is in England


MarionJHolloway (approved):
Record for the smallest PR?

PR #309, opened 2024-03-05
Clean ons data job

garypolzin:
# Description
We were having issues with multiple postcode joins in one script so created a cleaning script which reformats the geography data to contain a lot less columns and only requires one join now to bring in all geography data, past and present.

All columns are still accessible too, in case we need them further down the pipeline

This will also make the unmatched postcodes easier to identify as there is only one join

Haven't tested in branch yet but the PR is a bit of a beast so opening the PR now

[Trello](https://trello.com/c/e6DLgcpH/293-epic-4-create-clean-ons-job-and-refactor-columns-must)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


jobs/clean_ons_data.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return refactored_df.withColumnRenamed(
+        ONSClean.ons_import_date, ONSClean.current_ons_import_date
+    )
+
+
+def refactor_columns_as_struct_with_alias(df: DataFrame, alias: str) -> DataFrame:
+    return df.select(
+        ONSClean.postcode,
+        ONSClean.ons_import_date,
```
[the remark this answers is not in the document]
garypolzin: this function is called twice, once for contemporary and once for current so it needs to be `ons_import_date`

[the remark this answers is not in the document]
garypolzin: the alias is where we assign current vs contemporary

tests/unit/test_clean_ons_data.py
```diff
@@ -0,0 +1,87 @@
...
+        super().setUp()
+
+    def test_only_most_recent_rows_for_max_date_are_kept(self):
+        returned_df = job.prepare_current_ons_data(
+            self.test_ons_postcode_directory_with_date_df
+        )
+        self.assertEqual(returned_df.count(), 3)
+
+        returned_data = returned_df.collect()
+        self.assertEqual(
```
MarionJHolloway, line 81: It looks like this is only testing one row of data, when it needs to check that none of the rows contain a different date

jobs/clean_ons_data.py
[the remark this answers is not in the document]
MarionJHolloway: In row 52 you rename the import date for the current data. I'm suggest that instead of having two date columns in the final dataframe called ons_import_date and current_ons_import_date, we have columns called contemporary_ons_import_date and current_ons_import_date to avoid confusion. You could do it as the last step in the job if that makes more sense.

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,98 @@
...
+
+    ons_df = cUtils.column_to_date(ons_df, Keys.import_date, ONSClean.ons_import_date)
+
+    refactored_ons_df = refactor_columns_as_struct_with_alias(
+        ons_df, ONSClean.contemporary
+    )
+
+    current_ons_df = prepare_current_ons_data(ons_df)
+
+    refactored_ons_with_current_ons_df = refactored_ons_df.join(
```
MarionJHolloway, line 31: This join is untested, worth refactoring into a function so you can test it?

tests/unit/test_clean_ons_data.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_refactor_columns_as_struct_returns_df_with_correct_number_of_rows(self):
+        self.assertIsInstance(self.returned_df, DataFrame)
+        self.assertEqual(self.returned_df.count(), 5)
+
+    def test_refactored_schema_matches_expected_schema(self):
+        self.assertEqual(self.returned_df.schema, Schema.expected_refactored_schema)
+
+
+class PrepareCurrentONSTests(CleanONSDatasetTests):
```
MarionJHolloway, line 70: This class had two tests in the locations script, are they not both needed now for some reason?

jobs/clean_cqc_location_data.py
```diff
@@ -125,53 +100,45 @@ def main(
...
 
-    new_ons_col_names = [STRING_TO_PREPEND + col for col in COLS_TO_RENAME]
 
-    for i in range(len(COLS_TO_RENAME)):
-        current_ons_df = current_ons_df.withColumnRenamed(
-            COLS_TO_RENAME[i], new_ons_col_names[i]
-        )
-
-    current_ons_df = current_ons_df.withColumnRenamed(ONS.postcode, CQCL.postcode)
+def join_ons_postcode_data_into_cqc_df(
```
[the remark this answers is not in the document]
garypolzin: possibly yes - it's hard to read in this format but we previously had 2 postcode join functions

`join_current_ons_postcode_data` - I deleted this

`join_contemporary_ons_postcode_data` - I pretty much removed the word contemporary and this became the new function. Then also added in the two postcode related tasks from main so all geography tasks contained in one function

So I can put it all into main but the same joins weren't in main before, I don't know the benefits/downsides of either approach so happy either way

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
```diff
@@ -20,18 +20,10 @@ class CqcLocationCleanedColumns(CqcLocationApiColumns):
     cqc_sector: str = CQCPClean.cqc_sector
     provider_name: str = "provider_name"
     cqc_location_import_date: str = "cqc_location_import_date"
-    current_cssr: str = "current_" + ONS.cssr
```
MarionJHolloway, line 23: These removed column names are used in later scripts and will want refactoring in everything downstream as well

jobs/clean_cqc_location_data.py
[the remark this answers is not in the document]
MarionJHolloway: The join function seems to be within this join function?

jobs/clean_ons_data.py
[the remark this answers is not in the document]
garypolzin: ah got you, I've done this

jobs/clean_cqc_location_data.py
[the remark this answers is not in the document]
garypolzin: all in one function now :)

tests/unit/test_clean_ons_data.py
garypolzin: yea wasn't sure how specific to go with tests

original df has 5 rows in - 3 have the latest date and the other 2 have older dates

so I checked the returned file has 3 rows in and that the date for one of them was the most recent date as a double check - I can add the other rows as checks too

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
garypolzin: thought I'd leave some work for another ticket!

tests/unit/test_clean_ons_data.py
garypolzin: yea the other was checking that columns had been renamed correctly but we aren't doing that step anymore

tests/unit/test_clean_ons_data.py
garypolzin: done :)

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
MarionJHolloway: We will need to bring that ticket into this sprint. I've added it to the top of sprint 4 so it doesn't get lost in new tickets

MarionJHolloway (commented):
Looks good to me, it would be good to get another set of eyes on this though, as it's a big PR, so haven't approved. Also haven't requested changes through the review process because I'm off tomorrow and don't want to block it merging.

jobs/clean_cqc_location_data.py
```diff
@@ -80,32 +71,16 @@ def main(
...
-
-    current_ons_postcode_directory_df = prepare_current_ons_data(
-        ons_postcode_directory_df
-    )
-
     cqc_location_df = cUtils.column_to_date(
         cqc_location_df, Keys.import_date, CQCLClean.cqc_location_import_date
     )
 
     cqc_location_df = remove_non_social_care_locations(cqc_location_df)
```
luke-a-williams, line 78: Can this logic be moved to line 63 since it is a where clause and would effectively reduce the amount of data the rest of the transformations would need to be performed on? I can have a look in Glue and see how much of a difference this might make if you aren't certain about moving it..

jobs/clean_cqc_location_data.py
```diff
@@ -80,32 +71,16 @@ def main(
...
 
-    cqc_location_df = utils.normalise_column_values(cqc_location_df, CQCL.postcode)
-
-    cqc_location_df = join_current_ons_postcode_data(
-        cqc_location_df, current_ons_postcode_directory_df
-    )
-
     cqc_location_df = join_cqc_provider_data(cqc_location_df, cqc_provider_df)
 
     cqc_location_df = add_list_of_services_offered(cqc_location_df)
```
luke-a-williams, line 86: This function and allocate_primary_service_type could be faster when ran before the joins, as it would be working on less data... unless it requires dependence on joined columns.
If there is dependence on joined columns, then the order of these functions matters and should be tested

luke-a-williams (commented):
Initial comments on a quick scan of the code

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,110 @@
...
+from utils.column_names.cleaned_data_files.ons_cleaned_values import (
+    OnsCleanedColumns as ONSClean,
+)
+
+
+onsPartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
+
+
+def main(ons_source: str, cleaned_ons_destination: str):
+    ons_df = utils.read_from_parquet(ons_source)
```
luke-a-williams, line 21: Is it necessary to read in ALL of the ons columns - or can a core subset be selected to significantly improve run-times?

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,110 @@
...
+            ONSClean.rural_urban_indicator_2011,
+            ONSClean.lower_super_output_area_2021,
+            ONSClean.middle_super_output_area_2021,
+            ONSClean.westminster_parliamentary_consitituency,
+        ).alias(alias),
+        Keys.year,
+        Keys.month,
+        Keys.day,
+        Keys.import_date,
+    )
```
luke-a-williams, line 89: Since this function's purpose seems to perform a select operation on the ONS data, can this logic be moved to follow immediately after the read operation for optimisation purposes?

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
garypolzin: just checked and we don't actually pull any geography columns into the merge script yet so even if these column names were added now it wouldn't run downstream as the columns don't exist yet. I've create a ticket

jobs/clean_cqc_location_data.py
garypolzin: I've reordered all the functions so all the location stuff happens first, then the two joins at the end of the job

jobs/clean_ons_data.py
garypolzin: I've refactored the ONS data so it now looks like below.

image

The schema is:
 |-- HYBRID_POSTCODE: string (nullable = true)
 |-- contemporary_ons_import_date: date (nullable = true)
 |-- contemporary: struct (nullable = false)
 |    |-- CSSR: string (nullable = true)
 |    |-- Region: string (nullable = true)
 |    |-- Sub_ICB: string (nullable = true)
 |    |-- ICB: string (nullable = true)
 |    |-- ICB_Region: string (nullable = true)
 |    |-- CCG: string (nullable = true)
 |    |-- lat: string (nullable = true)
 |    |-- long: string (nullable = true)
 |    |-- imd: string (nullable = true)
 |    |-- lsoa11: string (nullable = true)
 |    |-- msoa11: string (nullable = true)
 |    |-- ru11ind: string (nullable = true)
 |    |-- lsoa21: string (nullable = true)
 |    |-- msoa21: string (nullable = true)
 |    |-- pcon: string (nullable = true)
 |-- year: string (nullable = true)
 |-- month: string (nullable = true)
 |-- day: string (nullable = true)
 |-- import_date: string (nullable = true)
 |-- current_ons_import_date: date (nullable = true)
 |-- current: struct (nullable = true)
 |    |-- CSSR: string (nullable = true)
 |    |-- Region: string (nullable = true)
 |    |-- Sub_ICB: string (nullable = true)
 |    |-- ICB: string (nullable = true)
 |    |-- ICB_Region: string (nullable = true)
 |    |-- CCG: string (nullable = true)
 |    |-- lat: string (nullable = true)
 |    |-- long: string (nullable = true)
 |    |-- imd: string (nullable = true)
 |    |-- lsoa11: string (nullable = true)
 |    |-- msoa11: string (nullable = true)
 |    |-- ru11ind: string (nullable = true)
 |    |-- lsoa21: string (nullable = true)
 |    |-- msoa21: string (nullable = true)
 |    |-- pcon: string (nullable = true)

jobs/clean_ons_data.py
garypolzin: it is a select function but it doesn't drop any columns, it's the approach I've taken to refactor the df

we start with a df containing 21 columns
we finish with a df containing 7 column - all 15 geog columns are combined into one column which we can extract later

we can access the actual geogs we need later on by including the following
    df = df.withColumn(
        "current_cssr", df[ONSClean.current].cssr
    )

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,110 @@
...
+        (
+            "--ons_source",
+            "Source s3 directory for parquet ONS postcode directory dataset",
+        ),
+        (
+            "--cleaned_ons_destination",
+            "Destination s3 directory for cleaned parquet ONS postcode directory dataset",
+        ),
+    )
+    # Python logic ---> all in main
```
james-francis-MT, line 107: Probably no need for these comments

jobs/clean_ons_data.py
james-francis-MT: What is the reason for squashing columns like this if you need to extract them later anyway?

james-francis-MT (commented):
Would like to see this tested in branch fully to check it doesn't impact later jobs but otherwise looks good

luke-a-williams:
@james-francis-MT I've been running it through with Gary, didn't make a lot of comments on the PR as tests were failing and we got the tests working just now, so going to re-review

jobs/clean_ons_data.py
luke-a-williams: @james-francis-MT ^^ this image might help answer your other question as to why it is being squashed, it's being run twice and its a lot of columns that currently aren't being used but don't want to be lost... or so I understand.

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,116 @@
...
+
+    utils.write_to_parquet(
+        refactored_ons_with_current_ons_df,
+        cleaned_ons_destination,
+        mode="overwrite",
+        partitionKeys=onsPartitionKeys,
+    )
+
+
+def prepare_current_ons_data(df: DataFrame) -> DataFrame:
```
luke-a-williams, line 45: Can we get a docstring in here describing how this function is used and what it is doing?
My understanding is that you pass it a complete ons dataframe and it goes and finds the latest import date for each geography with the various fields of that geography, returning the df without partition keys

jobs/clean_ons_data.py
luke-a-williams: Optimisation I think, looking at the job it's running in 3 minutes even with a join being present which is pretty good. But the probable answer is "this is what was being done before"... Gary might be able to shine some more light on this.

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def join_current_ons_df_into_contemporary_df(
+    refactored_ons_df: DataFrame, current_ons_df: DataFrame
+) -> DataFrame:
+    return refactored_ons_df.join(current_ons_df, ONSClean.postcode, "left")
+
+
+if __name__ == "__main__":
+    # Where we tell Glue how to run the file, and what to print out
```
luke-a-williams, line 99: Don't need this comment

jobs/clean_ons_data.py
```diff
@@ -0,0 +1,116 @@
...
+        Keys.year,
+        Keys.month,
+        Keys.day,
+        Keys.import_date,
+    )
+
+
+def join_current_ons_df_into_contemporary_df(
+    refactored_ons_df: DataFrame, current_ons_df: DataFrame
+) -> DataFrame:
```
luke-a-williams, line 94: Can we get a docstring for this function? The order of the supplied dfs is important as it will still not error if they are put in the wrong way, so an explanation in the parameters of the docstring will help with this. If they are put in the wrong way, as we discussed, you'll just have a lot of empty rows where a newer geography doesn't have an older record.

luke-a-williams (changes requested):

terraform/pipeline/glue.tf
```diff
@@ -428,7 +442,7 @@ module "clean_cqc_location_data_job" {
   job_parameters = {
     "--cqc_location_source"              = "${module.datasets_bucket.bucket_uri}/domain=CQC/dataset=locations_api/"
     "--cleaned_cqc_provider_source"      = "${module.datasets_bucket.bucket_uri}/domain=CQC/dataset=providers_api_cleaned/"
-    "--ons_postcode_directory_source"    = "${module.datasets_bucket.bucket_uri}/domain=ONS/dataset=postcode_directory/"
+    "--ons_postcode_directory_source"    = "${module.datasets_bucket.bucket_uri}/domain=ONS/dataset=postcode_directory_cleaned/"
```
luke-a-williams, line 445: should we call this cleaned_ons_postcode_directory_source? So as not to cause confusion with the postcode directory source in other scripts.

jobs/clean_ons_data.py
garypolzin: yea geographies are a bit different to other cols. We probably won't use many geographies during the pipeline but we might want access to some/all of them at the end when splitting our estimated outputs out. There are 30 geog columns (15 contemporary, 15 current) so this reduces them into 2 cols whilst still being able to access them all at any point

this wasn't really the main objective of the ticket though. We were previously cleaning ONS data in the CQC locations job and having to do 2 joins - this new file means that we only need one join in CQC locations job

jobs/clean_cqc_location_data.py
```diff
@@ -275,7 +227,7 @@ def split_dataframe_into_registered_and_deregistered_rows(
             "Source s3 directory for cleaned parquet CQC provider dataset",
         ),
         (
-            "--ons_postcode_directory_source",
+            "--cleaned_ons_postcode_directory_source",
```
luke-a-williams, line 230: Relating to my other comment, since it is called cleaned_ons_postcode_directory_source here it won't work in your current terraform definitions

luke-a-williams (changes requested):
Mismatch in job parameters also causing failures in glue pipeline

jobs/clean_ons_data.py
garypolzin: yea just makes for a smaller file for joining - 9 columns instead of 27

jobs/clean_ons_data.py
garypolzin: yea I'll remove them - I copied clean_cqc_provider_data to get started so they've come from there

jobs/clean_cqc_location_data.py
```diff
@@ -125,58 +96,39 @@ def main(
...
-        formatted_ons_postcode_directory_df,
-        [CQCLClean.ons_import_date, CQCLClean.postcode],
-        "left",
-    )
-    return cqc_location_df
-
-
 def join_current_ons_postcode_data(cqc_loc_df: DataFrame, current_ons_df: DataFrame):
     return cqc_loc_df.join(
         current_ons_df,
```
luke-a-williams, line 134: This function isn't being used anymore, can you remove it?

terraform/pipeline/glue.tf
garypolzin: good spot, I'd missed this one

jobs/clean_ons_data.py
garypolzin: yep that's right, for each postcode throughout time it will add what the current set of geographies are now

need to drop the partition keys as these relate to the current data only, the partition keys we want to keep are already in the main df

garypolzin:
> Would like to see this tested in branch fully to check it doesn't impact later jobs but otherwise looks good

Will test this in branch but geographies aren't pulled into merge_ind_cqc yet, there's a separate ticket to include them
https://trello.com/c/kNSG2IRc/295-epic-7-bring-geographies-into-mergeindcqcdata-job-must

jobs/clean_cqc_location_data.py
```diff
@@ -63,55 +54,35 @@
...
         cqc_location_df, ons_postcode_directory_df
     )
 
-    cqc_location_df = utils.normalise_column_values(cqc_location_df, CQCL.postcode)
-
-    cqc_location_df = join_current_ons_postcode_data(
-        cqc_location_df, current_ons_postcode_directory_df
-    )
-
     cqc_location_df = join_cqc_provider_data(cqc_location_df, cqc_provider_df)
```
luke-a-williams, line 84: This doesn't work in the glue pipeline, giving an  "import_date is ambiguous" error meaning there is 2 columns with the same name. The problematic code is as follows:
"""
  File "clean_cqc_location_data.py", line 178, in join_cqc_provider_data
    locations_df[Keys.import_date]
"""
In the cleaned provider data there is a column called `cqc_provider_import_date` you could use instead to see if that alleviates this issue?

luke-a-williams: Or specify that it is the import_date that belongs to locations_df not just the key, maybe with an F.col(locations.import_date) or w.e. the column name is

luke-a-williams (changes requested):

jobs/clean_cqc_location_data.py
luke-a-williams: I don't know exactly where the problem occurs but the error seems to imply that at that point the location's df has two import_date columns. Maybe it could be coming from the join_ons_postcode_data_into_cqc_df function which does some stuff with import date too, but I don't fully understand exactly what's going on in that function without doing a deep dive myself!

garypolzin: only importing selected columns on the import now which should fix this

garypolzin: it was importing all columns from ons, so it was importing the ons partition keys as well as the ones already in the cqc df

jobs/clean_ons_data.py
james-francis-MT: Is the plan to use doc strings in functions that are just there to make the main function easier to read? I don't expect this function to be used anywhere else

luke-a-williams:
Re-running the job in the pipeline to see if the errors have been addressed

PR #312, opened 2024-03-06
Create prepare non res ind cqc features script

james-francis-MT:
# Description
Create script to prepare non res features

# How to test
Tested in branch and passing
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/job/jf-prepare_non_res_ind_cqc_features_job/run/jr_0e737a583b2e9c12aafc552290028a3cf04a1c405140028b135f8939cf86669c

Ticket: https://trello.com/c/K46ZiscX/233-epic-9-refactor-terraform-for-locationsnonresfeatureengineeringpy-must

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -46,6 +48,22 @@ class IndCqcColumns:
...
+    )
+    contemporary_rural_urban_indicator_2011: str = (
+        CQCLClean.contemporary_rural_urban_indicator_2011
+    )
+    job_count: str = "job_count"
+    job_count_unfiltered: str = "job_count_unfiltered"
+    job_count_unfiltered_source: str = "job_count_unfiltered_source"
+    service_count: str = "service_count"
+    date_diff: str = "date_diff"
+    care_home_features: str = "features"  # make this string more specific once it works
```
luke-a-williams, line 66: Is your comment or one that has been copied? Do we still need it?

luke-a-williams (approved):
Looking at ticket definition of what is to be done, this looks good. Not thoroughly checked any tests as assumed they too have just been renamed to match

utils/column_names/ind_cqc_pipeline_columns.py
james-francis-MT: This was copied from Marions branch. It is used for here code but I decided to copy everything so there were less conflicts

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
```diff
@@ -32,6 +32,7 @@ class CqcLocationCleanedValues:
     care_home_only: str = "Care home without nursing"
     non_residential: str = "non-residential"
     independent: str = CQCPValues.independent
+    local_authority: str = CQCPValues.local_authority
```
garypolzin, line 35: is there a reason why we've added this? I can't see that we use it anywhere

garypolzin: oh it might be from Marion's script, I'll check that first

jobs/prepare_non_res_ind_cqc_features.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def filter_locations_df_for_independent_non_res_care_home_data(
+    df: DataFrame, carehome_col_name: str, cqc_col_name: str
+) -> DataFrame:
+    care_home_data = df.filter(F.col(carehome_col_name) == "N")
+    independent_care_home_data = care_home_data.filter(
+        F.col(cqc_col_name) == "Independent"
+    )
+    return independent_care_home_data
```
garypolzin, line 157: we've already filtered to Independent sector so this function only needs to filter to care home == "N"

luke-a-williams: @james-francis-MT this might be enough to block the review, what do you think?

garypolzin: it's not wrong, it just isn't doing anything as it's filtering to something that it's already filtered to - it will have been like like this previously too, I'm pretty sure this isn't new code

garypolzin: I don't mind if you want to let this through and we could create another low priority ticket to remove that step

garypolzin: there's a ticket for this now so you can ignore this filter

https://trello.com/c/VqwJn2TZ/297-remove-filtering-to-independent-data-in-feature-engineering-jobs

utils/column_names/cleaned_data_files/cqc_location_cleaned_values.py
garypolzin: this will have been added because of the ind sector filter which we're getting rid out outside of the ticket so this is ok too

PR #313, opened 2024-03-06
changed all remaining instances of api folder from - to _

garypolzin:
# Description
The output of the cqc api scripts are still '-api' instead of '_api'.

I've changed all remaining instances I could find


luke-a-williams (approved):
Those are all of the places I can see too on a quick scan from a branch that isn't yours. GTM

PR #316, opened 2024-03-07
Update ons column names

MarionJHolloway:
# Description
https://trello.com/c/75x4YVbo/294-update-ons-column-names-downstream-in-pipeline-so-that-everything-still-works
Have removed the geographies struct and replaced with individual column names. 
Have created lists of geography columns for easy importing through the pipeline. 

# How to test
Unit tests passing
Whole pipeline running in branch successfully

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/clean_ons_data.py
```diff
@@ -50,43 +53,79 @@ def prepare_current_ons_data(df: DataFrame) -> DataFrame:
...
-            ONSClean.lower_super_output_area_2021,
-            ONSClean.middle_super_output_area_2021,
-            ONSClean.westminster_parliamentary_consitituency,
-        ).alias(alias),
-        Keys.year,
-        Keys.month,
-        Keys.day,
-        Keys.import_date,
-    )
+def refactor_columns_with_prefix(df: DataFrame, prefix: str) -> DataFrame:
```
garypolzin, line 61: not sure how I feel about this function, feels a bit misleading as the prefix isn't really a prefix and there's nothing to let you know when you call the function that it has to be either "_current" or "_contemporary"

might be better to split it into 2 select functions and a 'prepare' step which renames them without asking for a prefix

garypolzin: there already is a 'prepare_current_ons_data' function so the rename can sit in there

james-francis-MT (approved):

jobs/clean_ons_data.py
MarionJHolloway: Done. I think that's a better structure in the long run, in case we want to change something about the current/ contemporary columns, but not both

garypolzin (approved):
looks good :)

PR #320, opened 2024-03-12
Add error handling postcodes

luke-a-williams:
# Description
https://trello.com/c/X2rf20h6/179-epic-4-add-error-handling-for-postcodes-must

# How to test
Outside of the regular unit tests, pay attention to the changes to the main method as I had trouble doing anything more complicated than this, so see if this will suffice.
I also need to test this in Glue to make sure it errors out as expected, **but this is known about and we are expecting an error at this stage**

## Before: 
Screenshot 2024-03-12 at 14 23 36

## After:
Screenshot 2024-03-13 at 14 04 10
Successful run

# Developer checklist
- [x] Unit test
- [ ] Integration test
- [x] Linked to Trello ticket


jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+    """
+
+    COLUMNS_TO_FILTER = [
+        CQCLClean.postcode,
+        CQCLClean.location_id,
+    ]
+
+    if not column_to_check_for_nulls in cleaned_locations_df.columns:
+        raise AnalysisException(
+            f"ERROR: A column or function parameter with name {column_to_check_for_nulls} cannot be resolved."
```
garypolzin, line 252: maybe reword to be more specific? The function argument specifies a column so something like
f"ERROR: The column with name {column_to_check_for_nulls} cannot be found in the dataframe."

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+
+    sample_clean_null_df = cleaned_locations_df.filter(
+        F.col(column_to_check_for_nulls).isNull()
+    )
+    if not sample_clean_null_df.rdd.isEmpty():
+        list_of_tuples = []
+        data_to_log = (
+            sample_clean_null_df.select(COLUMNS_TO_FILTER)
+            .groupBy(COLUMNS_TO_FILTER)
+            .count()
```
garypolzin, line 263: I don't think the count will be necessary here - it will fail in the same way if there are 1 or 100 instances of a particular postcode failing. And you only need to fix that postcode once to fix all instances of it.

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
     return registered_df, deregistered_df
 
 
+def check_current_against_contemporary_geographies(
```
garypolzin, line 225: This function is to identify if the CQC postcode provided can be found in the ONS postcode file. If it is, a value will be provided. If not it will be a null entry. If it is null we want to raise an error

on that basis maybe we rename the function to be something like

def raise_error_if_postcode_not_found_in_ons_dataset():

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+    cleaned_locations_df: DataFrame,
+    column_to_check_for_nulls: str = CQCLClean.current_ons_import_date,
+) -> DataFrame:
+    """
+    Checks a cleaned locations df for any mismatches of contemporary versus current geographies based on the column_to_check_for_nulls,
+    where this column is a one column from a subset of columns resulting from a left join and so could contain null values.
```
garypolzin, line 231: Reword to something like:

Checks a cleaned locations df for any CQC postcodes which could not be found in the ONS postcode directory based on the column_to_check_for_nulls,
    where this column is a one column from a subset of columns resulting from a left join and so could contain null values.

jobs/clean_cqc_location_data.py
garypolzin: We should probably also add postcode and location_id into the AnalysisException as all 3 need to exist for this function to be successful

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+    Args:
+        cleaned_locations_df (DataFrame): A cleaned locations df that must contain at least the column_to_check_for_nulls, postcode and locationId.
+        column_to_check_for_nulls (str): Default - current_ons_import_date, should be any one of the left-joined columns to check for null entries.
+
+    Returns:
+        cleaned_locations_df (DataFrame): If the check does not find any null entries, it returns the original df. If it does find anything exceptions will be raised instead.
+
+    Raises:
+        AnalysisException: If column_to_check_for_nulls is mistyped or otherwise not present in cleaned_locations_df/
+        SystemExit: If the check finds mismatched postcodes, it will raise this and cause a glue job failure where the output should feature in Glue's logs.
```
garypolzin, line 242: could change to:

cause a glue job failure where **the unmatched postcodes and corresponding locationid** should feature in Glue's logs

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+            sample_clean_null_df.select(COLUMNS_TO_FILTER)
+            .groupBy(COLUMNS_TO_FILTER)
+            .count()
+            .collect()
+        )
+
+        for row in data_to_log:
+            list_of_tuples.append((row[0], row[1], f"count: {row[2]}"))
+        raise TypeError(
+            f"Error: Problem matching contemporary to current ons data. Here is a list of their {CQCL.postcode} and {CQCL.location_id}: {list_of_tuples}"
```
garypolzin, line 270: consider changing to:
f"Error: The following {CQCL.postcode}(s) and their corresponding {CQCL.location_id}(s) were not found in the ONS postcode data: {list_of_tuples}"

jobs/clean_cqc_location_data.py
```diff
@@ -217,6 +222,58 @@ def split_dataframe_into_registered_and_deregistered_rows(
...
+        )
+    else:
+        print("No mismatches found, returning original dataframe")
```
garypolzin, line 273: print("All postcodes were found in the ONS postcode file, returning original dataframe")

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -350,5 +357,61 @@ def test_join_ons_postcode_data_correctly_joins_data(self):
         self.assertCountEqual(returned_data, expected_data)
 
 
+class CheckCurrentAgainstContemporaryGeographies(CleanCQCLocationDatasetTests):
```
garypolzin, line 360: rename all test function names to match new function name

tests/test_file_data.py
```diff
@@ -1115,6 +1104,36 @@ class CQCLocationsData:
...
+            date(2019, 1, 1),
+            "PR19AB",
+            date(2020, 1, 1),
+            "loc-1",
+            "prov-1",
+            "Cardiff",
+            "Welsh Street",
+            date(2021, 1, 1),
+            "Some field in the hills",
+            "Welsh Region",
```
garypolzin, line 1122: "Some field in the hills" made me chuckle but unfortunately we don't cover Wales so maybe switch to England - otherwise we'll assume that this Welsh place shouldn't be a successful match!

utils/cqc_location_dictionaries.py
```diff
@@ -77,4 +77,7 @@ class InvalidPostcodes:
         "WF12 2SE": "WF13 2SE",
         "WR13 3JS": "WF13 3JS",
         "YO61 3FF": "YO61 3FN",
+        "CB10 4EJ": "CB10 1EJ",
+        "RG19 3DD": "RH19 3DD",
+        "BL11 5PE": "BB11 5PE",
```
garypolzin, line 82: postcode list is in alphabetical order, so if you can maintain that it would be great please :)

might get annoying long term but we'll maintain it for now

garypolzin (changes requested):

jobs/clean_cqc_location_data.py
luke-a-williams: So the count is following on from the groupby. This is being done to convert the groupby object back into a dataframe and condense the number of rows, effectively creating a row per postcode / locationid pair. Why I feel this is important is because IF as we saw there is more than one row being captured with the same postcode, currently this function is filtering out those duplicate rows into the Glue error.

I'm happy to remove the count, but what are your expectations around how you want the Glue Error to look?

jobs/clean_cqc_location_data.py
luke-a-williams: Happy to do this

jobs/clean_cqc_location_data.py
luke-a-williams: I also haven't updated the Raised value in the docstring, and can certainly mention the glue job

tests/test_file_data.py
luke-a-williams: Oh I didn't even think of that was just making up addresses. I can definitely make up another english address

utils/cqc_location_dictionaries.py
luke-a-williams: Good spot, will do this!

jobs/clean_cqc_location_data.py
garypolzin: yea I wasn't sure if this code was specifically so we had a count column or not, as I think you could just do 

sample_clean_null_df.select(COLUMNS_TO_FILTER).dropDuplicates()

but I guess it's the same thing

jobs/clean_cqc_location_data.py
luke-a-williams: Yes I can add those additional checks into the code and test them

garypolzin (approved):
I believe all comments have been addressed so happy to approve!

PR #325, opened 2024-03-13
Update non res features

garypolzin:
# Description
prepare_non_res_ind_cqc_features and prepare_care_home_ind_cqc_features are identical jobs apart from a filter at the start which filters one to care_home “Y” and the other to “N”

However the way they and the tests were written differed quite substantially

Also changed the way one was saved from "append" to "overwrite"

I've aligned both jobs and tests so everything is in the same order, tests are called the same, added tests missing from one script, data/schemas stored in the same place and generally speaking, when side by side, they look the same other than intentional differences

[Trello](https://trello.com/c/VqwJn2TZ/297-epic-9-fix-feature-engineering-jobs-must)

I've also created a Trello card to look at why we have 2 scripts 
[Trello](https://trello.com/c/Jlunp2Tu/305-epic-9-merge-features-into-one-job-should)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


luke-a-williams (approved):
Ran through with Gary, happy for this to be merged in to align the two 2 jobs for further testing of where the problem might be.

PR #330, opened 2024-03-15
reduced ons columns which are imported

garypolzin:
# Description
Only kept columns we use in the pipeline

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/jobs/merge_ind_cqc_data.py
+++ b/jobs/merge_ind_cqc_data.py
@@ -11,8 +11,7 @@ from utils.column_names.cleaned_data_files.cqc_location_cleaned_values import (
     CqcLocationCleanedValues as CQCLValues,
 )
 from utils.column_names.cleaned_data_files.ons_cleaned_values import (
-    current_geography_columns,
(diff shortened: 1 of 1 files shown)


jobs/merge_ind_cqc_data.py
```diff
@@ -43,8 +42,13 @@
...
     CQCLClean.primary_service_type,
-    *current_geography_columns,
-    *contemporary_geography_columns,
+    ONSClean.contemporary_ons_import_date,
+    ONSClean.contemporary_cssr,
+    ONSClean.contemporary_region,
+    ONSClean.current_ons_import_date,
+    ONSClean.current_cssr,
+    ONSClean.current_region,
+    ONSClean.current_rural_urban_ind_11,
```
luke-a-williams, line 51: ONS fields look fine, double checked and can see only rural_ind, and region being used, but might use cssr later

luke-a-williams (approved):
Should reduce the amount of data being processed by the job, lets see if this helps.

PR #333, opened 2024-03-18
Reorder clean cqc functions

garypolzin:
# Description
Currently the final step of the script is to split the dataset into registered and deregistered.

46% of the CQC dataset are deregistered locations so if we move this step earlier on in the script then it's a lot less data being included in the join functions (we don't need ONS or PIR data in the Deregistered dataset)

No change to the dataset being used and still runs fine in branch

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


jobs/clean_cqc_location_data.py
```diff
@@ -92,17 +91,21 @@ def main(
...
-    )
-
-    cqc_location_df = join_cqc_provider_data(cqc_location_df, cqc_provider_df)
-
     (
         registered_locations_df,
         deregistered_locations_df,
     ) = split_dataframe_into_registered_and_deregistered_rows(cqc_location_df)
 
+    registered_locations_df = add_list_of_services_offered(registered_locations_df)
```
MarionJHolloway, line 99: I thought these steps needed to happen to the deregistered locations as well? I'm sure when we were thinking early on about how the pipeline would interact with other processes, we decided they needed to be applied to both?

garypolzin: we only need the deregistered location dataset for reconciliation at the moment, and that doesn't request require any of the other data produced in this script

MarionJHolloway (approved):

PR #334, opened 2024-03-18
Provider api key

luke-a-williams:
# Description
https://trello.com/c/NBGWFkwP/24-epic-3-add-partner-code-to-api-calls-should
I have tried 2 different setups, shown in the commit history:
1) Having the Secret populated at build time. This is probably fine for this particular parameter, but since this sets a precedent and it is not known how visible circleci logs are for this parameter (security concerns around this) its not viable for more secure parameter requirements.
2) The **implementation shown in this PR**, which instead of loading the parameter at build time, loads the parameter at execution time. More secure, but slightly more involved in the terraform to implement (as it requires giving Glue the privilege to call from the parameter store)

# How to test
Runs have already been successful, however there hasn't been noticeable time saving. The ticket is to implement the partnerCode, and not to optimise the current script or debug why it hasn't sped it up.
I have however attached to the trello links to our slack thread discussion showing visuals on the before and after and how there doesn't appear to be change. Attached to the trello ticket is also an image to show the code is being pulled successfully

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/bulk_download_cqc_locations.py
```diff
@@ -28,22 +32,14 @@ def main(destination):
...
-
-    return args.destination_prefix
-
-
 if __name__ == "__main__":
-    destination_prefix = collect_arguments()
+    destination_prefix = utils.collect_arguments(
+        (
+            "--destination_prefix",
+            "Source s3 directory for parquet CQC providers dataset",
```
garypolzin, line 39: CQC **locations** dataset

utils/cqc_api.py
```diff
@@ -2,9 +2,13 @@
...
 
 import requests
+import boto3
+from botocore.exceptions import ClientError
 
 CQC_API_VERSION = "v1"
-RATE_LIMIT = 400  # Max ratelimit = 600 per minute.
+RATE_LIMIT = (
+    400  # Max ratelimit = 600 per minute without partnercode, 2000 with a partnercode
+)
```
garypolzin, line 11: main reason for the partnercode was to speed up the rate limit so just checking that 400 is what we're staying at?

MarionJHolloway (changes requested):
Looks like these are currently failing in branch?

utils/cqc_api.py
luke-a-williams: Marion and i spoke about this yesterday, we tried putting it up to 2000 and didn't notice any improvement at all, so tried putting it back to 400 to see if our own rate limiting would help reduce the amount of sleeping. If it has helped, its marginal especially on the full runs. There are some slack threads in chat where I have visually compared the logs for the 2 different rate limited parameters

luke-a-williams: I've collected them on the trello ticket: https://trello.com/c/NBGWFkwP/24-epic-3-add-partner-code-to-api-calls-should

luke-a-williams: I'm investigating putting it back to 2k now on a small set now I know the partnercode is formatted correctly this morning

luke-a-williams: Yeah it doesn't seem to have any affect

jobs/bulk_download_cqc_locations.py
```diff
@@ -8,13 +7,20 @@
...
     spark = utils.get_spark()
     df = None
     for paginated_locations in cqc.get_all_objects(
-        stream=True, object_type="locations", object_identifier=ColNames.location_id
+        stream=True,
+        object_type="locations",
+        object_identifier=ColNames.location_id,
+        partner_code=json.loads(
+            cqc.get_secret(secret_name="partner_code", region_name="eu-west-2")
+        )["partner_code"],
```
james-francis-MT, line 23: Can this function call be moved outside the parameters?

jobs/bulk_download_cqc_providers.py
```diff
@@ -8,13 +7,20 @@
...
     spark = utils.get_spark()
     df = None
     for paginated_providers in cqc.get_all_objects(
-        stream=True, object_type="providers", object_identifier=ColNames.provider_id
+        stream=True,
+        object_type="providers",
+        object_identifier=ColNames.provider_id,
+        partner_code=json.loads(
+            cqc.get_secret(secret_name="partner_code", region_name="eu-west-2")
+        )["partner_code"],
```
james-francis-MT, line 23: same here

utils/cqc_api.py
james-francis-MT: If it has made no difference then how do we know it is working at all?

utils/cqc_api.py
```diff
@@ -67,3 +79,51 @@ def get_object(cqc_location_id, object_type):
...
+    session = boto3.session.Session()
+    client = session.client(
+        service_name="secretsmanager",
+        region_name=region_name,
+    )
+
+    try:
+        get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "ResourceNotFoundException":
```
james-francis-MT, line 111: Perhaps a switch statement works best here

utils/cqc_api.py
```diff
@@ -67,3 +79,51 @@ def get_object(cqc_location_id, object_type):
...
+            )
+        elif e.response["Error"]["Code"] == "InternalServiceError":
+            print("An error occurred on service side:", e)
+    else:
+        if "SecretString" in get_secret_value_response:
+            text_secret_data = get_secret_value_response["SecretString"]
+            return text_secret_data
+        else:
+            binary_secret_data = get_secret_value_response["SecretBinary"]
+            return binary_secret_data
```
james-francis-MT, line 129: Probably worth writing some unit tests for this

jobs/bulk_download_cqc_locations.py
luke-a-williams: Certainly

jobs/bulk_download_cqc_providers.py
luke-a-williams: Actually that's a really good point, its retrieving it once per page with this setup so definitely makes sense to call it before entering the loop. Good spot.

utils/cqc_api.py
luke-a-williams: switch statements are a feature of python 3.10 and up. Our environment is on python version 3.9. Or did you mean something else?

utils/cqc_api.py
luke-a-williams: 100%, sorry it was taking me a while as I had to understand more about the kind of response packet that AWS could send back in this scenario. I have added tests in for this.

utils/cqc_api.py
james-francis-MT: wow 🤬 python is such a great language

luke-a-williams: There isn't sad laughing emoji to respond to your comment hah

utils/cqc_api.py
luke-a-williams: It's working on our end, in the sense that we are definitely sending a GET request to a url that now has a partnerCode parameter. Looking at the logs for key runs though, seems like the problem is their end. :(

james-francis-MT: Has someone taken an action to check with them?

james-francis-MT (approved):

MarionJHolloway (approved):

PR #338, opened 2024-03-20
Add care home model

garypolzin:
# Description
- Reintroduced the care home ML model into the estimates job

- Metrics changes
Previously we had a standalone job which created the r2 metric. This was called part way through the care_home_model job and the value was returned at the end and with the df. The estimates script got this returned r2 value and there was another function in the estimates script which imported the r2 metric and wrote to s3.

I've changed this so we now have one ml_model_metrics function which creates and writes the r2 within the model itself. It's generic so the non-res models can use this function too

Also simplified the metrics dataframe to just include the name and version of the model, r2 value and current timestamp. Previously also had latest import date (basically the same as timestamp but less detailed), job_run_id and job_name (both were set to "test" so didn't add any value)

- Removed any remaining references to the non-res pir models

- Renamed the test model folder (hence all the parquet files)

- Renamed the step function to be more descriptive/relevant

Side note - tidied some errors in test_data (double spaces or no spaces) and removed references to "(pseudo) Wales"

Whole pipeline ran fine in branch - [Pipeline](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:add-care-home-model-Ind-CQC-Filled-Post-Estimates-Pipeline:44f2e20c-e74f-422d-961e-a61e6666b78e)

[Trello](https://trello.com/c/Tk67t0mw/316-include-care-home-ml-model-in-estimates-script-must)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


tests/unit/test_model_care_home.py
```diff
@@ -31,22 +28,34 @@ def setUp(self):
...
 
-        self.assertEqual(df.count(), 5)
+        self.assertEqual(save_model_metrics.call_count, 1)
 
-    def test_model_care_homes_estimates_jobs_for_care_homes_only(self):
-        df, _ = model_care_homes(
+    @patch("utils.estimate_filled_posts.models.care_homes.save_model_metrics")
+    def test_model_care_homes_returns_expected_data(
+        self,
+        save_model_metrics: Mock,  # not using this but doesn't run without it, is this the best way?
```
MarionJHolloway, line 48: Yes, You are using it for the patch so it doesn't actually try to save for s3 - just because you're not calling the Mock, doesn't mean it's not being used in the background :)

MarionJHolloway: Worth removing the comment though!

MarionJHolloway (approved):
One minor change - otherwise looks good. I've approved so you can merge once the comment is removed :)

tests/unit/test_model_care_home.py
garypolzin: ha yea, forgot I'd put that in there but this was the reason why!
It's gone :)

PR #346, opened 2024-03-25
ML model - hard code the region list in features script

garypolzin:
# Description
Previously the regions were compiled as a dynamic list based on what values were in that column - changed this so the region list is set (same process we use for rural/urban column) - this ensures that the number and order of region columns will match the ML model features

`add_rui_data_data_frame` function has been made more generic as it's used for rui and region-> `convert_categorical_variable_to_binary_variables_based_on_a_dictionary`

`explode_column_from_distinct_values` and `format_strings` are redundant now so have been removed

### Test run
[Ran successfully in branch](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:use-new-ch-model-Ind-CQC-Filled-Post-Estimates-Pipeline:dc1c6a88-4525-4e27-90ae-6a640be7301b)

### Trello
[Trello](https://trello.com/c/IqXLVHD6/343-update-care-home-model-must)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


terraform/pipeline/glue.tf
```diff
@@ -438,7 +438,7 @@ module "estimate_ind_cqc_filled_posts_job" {
   job_parameters = {
     "--cleaned_ind_cqc_source"        = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=cleaned_ind_cqc_data/"
     "--care_home_features_source"     = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=care_home_ind_cqc_features/"
-    "--care_home_model_source"        = "${module.pipeline_resources.bucket_uri}/models/care_home_filled_posts_prediction/2.1.0/"
+    "--care_home_model_source"        = "${module.pipeline_resources.bucket_uri}/models/care_home_filled_posts_prediction/3.1.0/"
```
luke-a-williams, line 441: I noticed the test for this job doesn't have it's model updated... does that matter? It is currently set to 0.0.1?

garypolzin: the test has a dummy model which is saved as 0.0.1 - it's not testing the s3 file path is correct :)

luke-a-williams (commented):
Checked that version numbers have been updated as per ticket instructions initially, now will check the tests

jobs/prepare_care_home_ind_cqc_features.py
```diff
@@ -100,32 +100,26 @@ def main(
...
     utils.write_to_parquet(
-        features_df,
+        vectorised_features_df,
         care_home_ind_cqc_features_destination,
         mode="overwrite",
         partitionKeys=[Keys.year, Keys.month, Keys.day, Keys.import_date],
     )
 
+    return list_for_vectorisation
+
```
luke-a-williams, line 117: Why are you returning this? It's not being used anywhere within the script, and also the main method is defined to be returning a DataFrame not a list.

luke-a-williams: I can see that you use it in the test script, but the test is testing that main produces a dataframe, not a list? It's a bit confusing

tests/unit/test_prepare_care_home_ind_cqc_features.py
```diff
@@ -52,22 +52,81 @@ def test_main(
...
+            "indicator_1",
+            "indicator_10",
+            "indicator_2",
+            "indicator_3",
+            "indicator_4",
+            "indicator_5",
+            "indicator_6",
+            "indicator_7",
+            "indicator_8",
+            "indicator_9",
```
luke-a-williams, line 70: If the order is not important, you can achieve this in a much shorter way by using `*list(RURAL_URBAN_INDICATOR_LOOKUP.keys())` such as in the following example of list extension:
`["Hi", *list(RURAL_URBAN_INDICATOR_LOOKUP.keys())]`
Gives:
`['Hi',
 'indicator_1',
 'indicator_2',
 'indicator_3',
 'indicator_4',
 'indicator_5',
 'indicator_6',
 'indicator_7',
 'indicator_8',
 'indicator_9',
 'indicator_10']`

tests/unit/test_prepare_care_home_ind_cqc_features.py
```diff
@@ -52,22 +52,81 @@ def test_main(
...
+            IndCQC.number_of_beds,
+            "ons_east_midlands",
+            "ons_eastern",
+            "ons_london",
+            "ons_north_east",
+            "ons_north_west",
+            "ons_south_east",
+            "ons_south_west",
+            "ons_west_midlands",
+            "ons_yorkshire_and_the_humber",
```
luke-a-williams, line 80: As above, can be shortened with `*list(REGION_LOOKUP.keys())`

tests/unit/test_prepare_care_home_ind_cqc_features.py
```diff
@@ -52,22 +52,81 @@ def test_main(
...
+            "service_27",
+            "service_28",
+            "service_29",
+            "service_3",
+            "service_4",
+            "service_5",
+            "service_6",
+            "service_7",
+            "service_8",
+            "service_9",
```
luke-a-williams, line 109: As above, as long as no numbers are missing and the order isn't important, can be shortened with:
`*list(SERVICES_LOOKUP.keys())`

tests/unit/test_prepare_care_home_ind_cqc_features.py
```diff
@@ -52,22 +52,81 @@ def test_main(
...
+            IndCQC.service_count,
+        ]
+
+        returned_feature_list = job.main(
+            self.IND_FILLED_POSTS_CLEANED_DIR, self.CARE_HOME_FEATURES_DIR
+        )
 
-        job.main(self.IND_FILLED_POSTS_CLEANED_DIR, self.CARE_HOME_FEATURES_DIR)
+        self.assertEqual(len(returned_feature_list), 51)
```
luke-a-williams, line 117: The main function is defined to output a dataframe, not a list, so if this is changed then this step would need refactoring to get the length of the list of columns of the dataframe, I think.

tests/unit/test_prepare_care_home_ind_cqc_features.py
```diff
@@ -52,22 +52,81 @@ def test_main(
...
 
         result = write_to_parquet_mock.call_args[0][0].orderBy(
             F.col(IndCQC.location_id)
         )
 
-        self.assertTrue(result.filter(F.col("features").isNull()).count() == 0)
+        self.assertTrue(result.filter(F.col(IndCQC.features).isNull()).count() == 0)
         expected_features = SparseVector(
-            43, [8, 11, 12, 13, 42], [1.0, 10.0, 1.0, 1.0, 1.0]
+            51, [8, 11, 20, 21, 50], [1.0, 10.0, 1.0, 1.0, 1.0]
```
luke-a-williams, line 127: I'll admit I don't know much about this Vector at all, what's the significance of these number changes?

luke-a-williams (changes requested):
Review comments for the prepare_care_home_ind_cqc_features part of this PR

jobs/prepare_care_home_ind_cqc_features.py
```diff
@@ -34,39 +34,39 @@ def main(
...
-    data_with_rui = add_rui_data_data_frame(
-        df=data_with_expanded_services,
-        rui_col_name=IndCQC.current_rural_urban_indicator_2011,
-        lookup_dict=rural_urban_indicator_dict,
+    features_df = (
+        convert_categorical_variable_to_binary_variables_based_on_a_dictionary(
+            df=features_df,
+            categorical_col_name=IndCQC.current_rural_urban_indicator_2011,
+            lookup_dict=rural_urban_indicator_dict,
+        )
```
luke-a-williams, line 56: Why is this surrounded by paranthesis? Surely you can just do `features_df =  convert_categorical_variable_to_binary_variables_based_on_a_dictionary`

jobs/prepare_care_home_ind_cqc_features.py
```diff
@@ -34,39 +34,39 @@ def main(
...
 
-    distinct_regions = get_list_of_distinct_ons_regions(
-        df=data_with_rui,
+    regions = list(ons_region_dict.keys())
+    features_df = (
+        convert_categorical_variable_to_binary_variables_based_on_a_dictionary(
+            df=features_df,
+            categorical_col_name=IndCQC.current_region,
+            lookup_dict=ons_region_dict,
+        )
```
luke-a-williams, line 65: Why is this surrounded by paranthesis? Surely you can just do features_df =  convert_categorical_variable_to_binary_variables_based_on_a_dictionary

luke-a-williams (changes requested):

tests/unit/test_prepare_care_home_ind_cqc_features.py
luke-a-williams: 51 columns wide df, and this is saying that the 8th, 11th, 20th, 21st, 50th columns have the entries in the list to the right.

luke-a-williams: I can't track this backwards to the columns easily that aren't empty but the test passes.

jobs/prepare_non_res_ind_cqc_features.py
```diff
@@ -105,31 +105,25 @@ def main(
...
     utils.write_to_parquet(
-        features_df,
+        vectorised_features_df,
         non_res_ind_cqc_features_destination,
         mode="overwrite",
         partitionKeys=[Keys.year, Keys.month, Keys.day, Keys.import_date],
     )
 
+    return list_for_vectorisation
+
```
luke-a-williams, line 121: Same comment as for the other case, either return a dataframe or have the function defined to return a list.
It;s not being used in the script so you COULD do something only in the test to capture that list.

luke-a-williams (changes requested):

luke-a-williams:
Spoke to Gary, because the PR looks huge, and he has instructed me that most of the changes are re-names of scripts and adding and removing of tests based on the changed function calls... but the overall logic of the script hasn't changed much.

We resolved a lot of the things I'd reviewed already and conditional on the return type of main being up to scratch I'm happy with it as it is, since Gary wrote no new functionality

jobs/prepare_care_home_ind_cqc_features.py
garypolzin: not to note that this is auto formatted like this by black

jobs/prepare_care_home_ind_cqc_features.py
garypolzin: just to note that this is auto formatted like this by black

luke-a-williams (approved):
All of my comments have been addressed

PR #351, opened 2024-03-26
Configure circle CI to copy models into branch pipeline resources

MarionJHolloway:
# Description
Updated circle CI config and default folders in s3 to copy models into branch
https://trello.com/c/7fkylgyw/345-add-config-to-circle-ci-to-copy-models-into-branches

# How to test
Models copied successfully in this branch

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,7 +181,7 @@ jobs:
       
       - aws-s3/sync:
           aws-region: AWS_REGION
-          from: "s3://sfc-default-pipeline-resources/models/"
+          from: "s3://sfc-main-pipeline-resources/models/"
           to: "s3://$(cat terraform/pipeline/pipeline_resources_bucket_name)/models/"
 
   terraform-destroy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,11 @@ jobs:
           aws-region: AWS_REGION
           from: "s3://sfc-default-datasets/"
           to: "s3://$(cat terraform/pipeline/datasets_bucket_name)/"
+      
+      - aws-s3/sync:
+          aws-region: AWS_REGION
+          from: "s3://sfc-default-pipeline-resources/models/"
+          to: "s3://$(cat terraform/pipeline/pipeline_resources_bucket_name)/models/"
 
   terraform-destroy:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -178,6 +178,11 @@ jobs:
           aws-region: AWS_REGION
           from: "s3://sfc-default-datasets/"
           to: "s3://$(cat terraform/pipeline/datasets_bucket_name)/"
+      
+      - aws-s3/sync:
+          aws-region: AWS_REGION
+          from: "s3://sfc-default-pipeline-resources/models/"
```
garypolzin, line 184: I'm pretty sure the models don't auto-populate from sfc-main into sfc-default so you'll need to take them from main instead

james-francis-MT: They are in the default bucket now - I assume that was part of this ticket. It does mean you will have to keep that updated though

james-francis-MT (approved):

.circleci/config.yml
garypolzin: yea we need to manually create the s3 filepath in main whenever there's a new model and copy all the files in from where ever James' EMR script saves them

so for this to work we'll need to remember add new models into both main and default, which we'll probably forget. If we copied from main it only means we need to save them in one place, but I don't know if there's a downside to getting from main?

james-francis-MT: I guess this is a bit different to the data copy where we would want both all of the models in branch + the latest so I think copying from main instead will work

james-francis-MT: Happy either way - just whatever works best going forward

luke-a-williams (approved):

.circleci/config.yml
MarionJHolloway: I was thinking that we'd just keep the most recent model in the default bucket, as they don't change that often, but happy to point to main instead. I'll change it for now and maybe when the models stabilise a bit we can switch it back to default?

garypolzin: yea I guess it's a fairly small change either way so we can revisit if it' not working - it's already approved so happy for you to merge :)

PR #357, opened 2024-03-27
Refactor production of estimate_filled_posts and source so the order can be more easily changed

garypolzin:
# Description
In the estimate job we produce several models (known data, ML model, extrapolation etc). We then create one 'final' column 'estimate_filled_posts' (the value we want to use) and 'estimate_filled_posts_source' (text description of which model was used)

Currently this already exists, however this is a mix of doing it within the job and doing it within each individual model. This makes changing the order of the models very difficult as you need to reorder functions, and some functions need to be run before others.

This PR removes the population of 'estimate_filled_posts' and 'estimate_filled_posts_source' from within the models themselves and instead creates on function at the end, where you give the order you want as a list and it populates in that specific order.

`update_dataframe_with_identifying_rule` is also used in the ascwds filled post calculation scripts (now renamed to 'add_source_description_to_source_column'). Previously this was untested so I've added tests.

[Trello](https://trello.com/c/M2KlDVQt/348-refactor-production-of-estimatefilledposts-and-source-so-the-order-can-be-more-easily-changed)

[Successful branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:separate-model-run-order-to-fi-Ind-CQC-Filled-Post-Estimates-Pipeline:2b07b4f3-5ca6-4413-b48b-41f9eaeccd4f)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


jobs/estimate_ind_cqc_filled_posts.py
```diff
@@ -60,32 +55,21 @@ def main(
...
-) -> pyspark.sql.DataFrame:
+) -> DataFrame:
     print("Estimating independent CQC filled posts...")
 
     cleaned_ind_cqc_df = utils.read_from_parquet(
         cleaned_ind_cqc_source, cleaned_ind_cqc_columns
     )
     care_home_features_df = utils.read_from_parquet(care_home_features_source)
 
-    cleaned_ind_cqc_df = cleaned_ind_cqc_df.withColumn(
```
garypolzin, line 71: note: all deleted steps in this function still happen but now within the new function

tests/test_file_data.py
```diff
@@ -2081,116 +2094,17 @@ class ModelPrimaryServiceRollingAverage:
...
-            "1-000000004",
-            "2023-03-01",
-            1677628800,
-            None,
-            "non-residential",
-            None,
-            None,
-            50.7,
-        ),
+        ("1-000000001", "2023-01-01", 1672531200, 15.0, "Care home with nursing", 15.0),
```
garypolzin, line 2097: note, same data as above but without the last 2 cols, so it's formatted differently

tests/test_file_data.py
```diff
@@ -2520,33 +2434,19 @@ class MLModelMetrics:
...
-            "1-000000002",
-            date(2023, 1, 15),
-            1673740800,
-            20.0,
-            20.0,
-            "ascwds_filled_posts",
-        ),
-        ("1-000000002", date(2023, 1, 17), 1673913600, None, 21.0, "other_source"),
-        ("1-000000002", date(2023, 1, 19), 1674086400, None, None, None),
+        ("1-000000001", date(2023, 1, 1), 1672531200, None),
```
garypolzin, line 2437: note: same data as above but without the last 2 cols, so it's formatted differently

utils/estimate_filled_posts/insert_predictions_into_locations.py
```diff
@@ -26,18 +26,8 @@ def insert_predictions_into_locations(
...
-    )
-    locations_with_prediction_model_column = (
-        locations_with_prediction_model_column.withColumn(
-            IndCqc.estimate_filled_posts,
-            F.when(
-                F.col(IndCqc.estimate_filled_posts).isNotNull(),
-                F.col(IndCqc.estimate_filled_posts),
-            ).otherwise(F.col(IndCqc.prediction)),
-        )
+    locations_with_predictions = locations_with_predictions.withColumnRenamed(
```
garypolzin, line 29: note: previously added a new duplicate column then dropped the original - now it just renames it

utils/ind_cqc_filled_posts_utils/utils.py
```diff
@@ -1,15 +1,55 @@
...
+def populate_estimate_filled_posts_and_source_in_the_order_of_the_column_list(
+    df: DataFrame,
+    order_of_models_to_populate_estimate_filled_posts_with: list,
+) -> DataFrame:
+    df = df.withColumn(IndCQC.estimate_filled_posts, F.lit(None).cast(IntegerType()))
+    df = df.withColumn(
+        IndCQC.estimate_filled_posts_source, F.lit(None).cast(StringType())
+    )
+
+    for model_name in order_of_models_to_populate_estimate_filled_posts_with:
```
MarionJHolloway, line 36: So, having this for loop will slow down the job, as everything will process in series instead of in parallel. There is no better way to loop through columns, but we could avoid this by adapting this function. Instead of listing the order of the models, we could abstract the content of this loop into a function(something like populate_estimate_and source()  )which takes a model name, and then this wrapper function can control the order of multiple calls to populate_estimate_and source(). If we moved this function into the job in the short term, we could shuffle stuff around easily to compare outputs until we settle on an order. That would allow spark to paralellise as much as possible, whilst still making it fairly simple to update the order of models we use. What do you think? Happy to talk this through on Thurs if you like

MarionJHolloway (approved):

utils/ind_cqc_filled_posts_utils/utils.py
garypolzin: Just a note to say a trello ticket has been created to do this
https://trello.com/c/94jAj8cd

PR #358, opened 2024-03-28
Create tests for provider endpoint

MarionJHolloway:
# Description
Duplicate the location tests to cover provider data as well
https://trello.com/c/xMfk4bU9/337-cqc-api-create-tests-for-provider-endpoint-must

# How to test
 Unit & integration tests passing

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


tests/unit/test_cqc_api_new.py
```diff
@@ -18,47 +33,111 @@ def json(self):
...
+        self.location_object_type = "locations"
+        self.location_id_examples = ["location_1", "location_2", "location_3"]
+        self.test_locations = [
+            {CQCL.location_id: self.location_id_examples[0]},
+            {CQCL.location_id: self.location_id_examples[1]},
+            {CQCL.location_id: self.location_id_examples[2]},
+        ]
+
+    @patch("utils.cqc_api_new.call_api")
+    def test_get_object_when_getting_location(self, mock_call_api: Mock):
```
james-francis-MT, line 48: I'm not sure why these tests need to be duplicated - it doesn't matter if you set the identifier to 'location1' 'provider' or 'foo' - there is no logic in the function that will change

Also I know it isn't part of this ticket but perhaps the `cqc_location_id` argument should be something more generic as it isn't always location_id

tests/unit/test_cqc_api_new.py
```diff
@@ -18,47 +33,111 @@ def json(self):
...
+    @patch("utils.cqc_api_new.get_object")
+    def test_get_page_objects_when_getting_locations(
+        self, mock_get_object: Mock, mock_call_api: Mock
+    ):
+        mock_call_api.return_value = {self.location_object_type: self.test_locations}
+
+        mock_get_object.side_effect = self.test_locations
+
+        returned_page_objects = cqc.get_page_objects(
+            self.test_url, 1, self.location_object_type, CQCL.location_id, "PARTNERCODE"
```
james-francis-MT, line 66: Similar point to the other comment - in theory the object type and identifier aren't limited to these values. The function doesn't have logic that depends on the value of them so why is there a need to duplicate

the only tests that make sense to have both, in my opinion, are the integration tests as you are calling 2 different endpoints

MarionJHolloway:
This also now includes removing the partner code from the get query parameters. This was what made our new bulk download jobs fail. The integration tests were previously passing as they pointed to the old API. I have left the partner code parameter in the functions as we can rename it and use for the authorisation key soon. This will save time wiring that up in a few days.

tests/integration/test_cqc_api_new_integration.py
```diff
@@ -1,35 +1,106 @@
...
+            self.assertIsNotNone(location[CQCL.provider_id])
+
+
+class ProviderApiTests(CqcApiIntegrationTests):
+    def setUp(self) -> None:
+        super().setUp()
+        self.object_type = "providers"
+        self.example_object = {
+            CQCP.location_ids: ["1-10000792582"],
+            CQCP.provider_id: "1-9098203603",
```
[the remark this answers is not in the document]
MarionJHolloway: It's an integration test, so it's calling the API. If we gave it fake IDs it would return an error, so we do need genuine ones here. They are publicly available, so shouldn't be a problem.

PR #360, opened 2024-04-04
Amended duplication removal function to only look at workplaces with a locationid

garypolzin:
# Description
The function remove_locations_with_duplicates simply removes all workplaces in ASCWDS who have a duplicate ID.
That works fine for the CQC pipeline, but it doesn't work for other work as it removes every workplace in ASCWDS which doesn't have an ID (treats all None values as duplicates).

The amendment
- separates the data into 2 dfs - those with and those without a location id
- duplication step only applied to those with a location id
- appends that data back with those without a location id

Amended tests accordingly (added two blank location id workplaces and checked they were unaffected)

[successful branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:non-cqc-ascwds-fix-Ind-CQC-Filled-Post-Estimates-Pipeline:1440e732-8c3c-466e-b4e6-b5ec8fcd6a19)

# Developer checklist
- [X] Unit test

### DIFF
--- a/tests/unit/test_clean_ascwds_workplace_data.py
+++ b/tests/unit/test_clean_ascwds_workplace_data.py
@@ -235,12 +235,12 @@ class RemoveLocationsWithDuplicatesTests(IngestASCWDSWorkerDatasetTests):
     def setUp(self) -> None:
         super().setUp()
 
-        self.test_location_df = self.spark.createDataFrame(
+        self.input_df = self.spark.createDataFrame(
             Data.small_location_rows, Schemas.location_schema
         )
 
-        self.returned_df = job.remove_workplaces_with_duplicate_location_ids(
-            self.test_location_df
+        self.workplaces_with_duplicate_location_ids_removed_df = (
+            job.remove_workplaces_with_duplicate_location_ids(self.input_df)
         )
 
         self.test_duplicate_loc_df = self.spark.createDataFrame(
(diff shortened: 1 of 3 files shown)


tests/unit/test_clean_ascwds_workplace_data.py
```diff
@@ -223,7 +239,9 @@ def setUp(self) -> None:
             Data.small_location_rows, Schemas.location_schema
         )
 
-        self.returned_df = job.remove_locations_with_duplicates(self.test_location_df)
+        self.returned_df = job.remove_workplaces_with_duplicate_location_ids(
```
MarionJHolloway, line 242: This class wants renaming too

tests/unit/test_clean_ascwds_workplace_data.py
```diff
@@ -236,34 +254,45 @@ def test_returns_the_correct_columns(self):
...
-        filtered_df = job.remove_locations_with_duplicates(
+        filtered_df = job.remove_workplaces_with_duplicate_location_ids(
             locations_with_different_import_dates_df
-        ).orderBy(AWP.location_id)
+        )
 
         self.assertEqual(
-            filtered_df.collect(), locations_with_different_import_dates_df.collect()
+            filtered_df.sort(AWP.organisation_id).collect(),
+            locations_with_different_import_dates_df.sort(
```
MarionJHolloway, line 293: Can we rename this just to make it clear this is the expected output and filtered_df is what we're checking?

MarionJHolloway (approved):
Looks good!

PR #362, opened 2024-04-05
Setup reconciliation job

garypolzin:
# Description
Created job which reads the 2 required files. Added into step function

[Trello](https://trello.com/c/udqfUf31/39-epic-sfc-initial-set-up-of-reconciliation-job)

[Successful branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:setup-reconciliation-Ind-CQC-Filled-Post-Estimates-Pipeline:2ace2488-0a14-48a0-a9c8-3011104e3ef6)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/jobs/reconciliation.py
+++ b/jobs/reconciliation.py
@@ -69,11 +69,11 @@ if __name__ == "__main__":
         ),
         (
             "--reconciliation_single_and_subs_destination",
-            "Destination s3 directory for reconciliation parquet singles and subs dataset",
+            "Destination s3 directory for singles and subs reconciliation CSV dataset",
         ),
         (
             "--reconciliation_parents_destination",
-            "Destination s3 directory for reconciliation parquet parents dataset",
+            "Destination s3 directory for parents reconciliation CSV dataset",
         ),
     )
     main(
--- a/jobs/reconciliation.py
+++ b/jobs/reconciliation.py
@@ -1,0 +1,83 @@
+import sys
+
+from utils import utils
+from utils.column_names.cleaned_data_files.cqc_location_cleaned_values import (
+    CqcLocationCleanedColumns as CQCLClean,
+)
+from utils.column_names.cleaned_data_files.ascwds_workplace_cleaned_values import (
+    AscwdsWorkplaceCleanedColumns as AWPClean,
+)
+
+cqc_locations_columns_to_import = [
+    CQCLClean.import_date,
+    CQCLClean.location_id,
+    CQCLClean.registration_status,
+    CQCLClean.deregistration_date,
+]
+cleaned_ascwds_workplace_columns_to_import = [
+    AWPClean.ascwds_workplace_import_date,
+    AWPClean.establishment_id,
+    AWPClean.nmds_id,
+    AWPClean.is_parent,
+    AWPClean.parent_id,
+    AWPClean.organisation_id,
+    AWPClean.parent_permission,
+    AWPClean.establishment_type,
+    AWPClean.registration_type,
+    AWPClean.location_id,
+    AWPClean.main_service_id,
+    AWPClean.establishment_name,
+    AWPClean.region_id,
+]
+
+
+def main(
+    cqc_location_api_source: str,
+    cleaned_ascwds_workplace_source: str,
+    reconciliation_single_and_subs_destination: str,
+    reconciliation_parents_destination: str,
+):
+    cqc_location_df = utils.read_from_parquet(
+        cqc_location_api_source, cqc_locations_columns_to_import
+    )
+    ascwds_workplace_df = utils.read_from_parquet(
+        cleaned_ascwds_workplace_source,
+        selected_columns=cleaned_ascwds_workplace_columns_to_import,
+    )
+
+
+if __name__ == "__main__":
+    print("Spark job 'reconciliation' starting...")
+    print(f"Job parameters: {sys.argv}")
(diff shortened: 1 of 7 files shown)


jobs/reconciliation.py
```diff
@@ -0,0 +1,83 @@
+import sys
+
+from utils import utils
+from utils.column_names.cleaned_data_files.cqc_location_cleaned_values import (
```
MarionJHolloway, line 4: If you're importing the CQC api, should this be the uncleaned columns? It doesn't make a difference currently, but they could be different in the future.

jobs/reconciliation.py
```diff
@@ -0,0 +1,83 @@
...
+    reconciliation_single_and_subs_destination: str,
+    reconciliation_parents_destination: str,
+):
+    cqc_location_df = utils.read_from_parquet(
+        cqc_location_api_source, cqc_locations_columns_to_import
+    )
+    ascwds_workplace_df = utils.read_from_parquet(
+        cleaned_ascwds_workplace_source,
+        selected_columns=cleaned_ascwds_workplace_columns_to_import,
+    )
```
MarionJHolloway, line 46: Worth adding a write step just to check it's wired up correctly in terraform?

jobs/reconciliation.py
```diff
@@ -0,0 +1,83 @@
...
+            "--cqc_location_api_source",
+            "Source s3 directory for initial CQC location api dataset",
+        ),
+        (
+            "--cleaned_ascwds_workplace_source",
+            "Source s3 directory for cleaned parquet ASCWDS workplace dataset",
+        ),
+        (
+            "--reconciliation_single_and_subs_destination",
+            "Destination s3 directory for reconciliation parquet singles and subs dataset",
```
MarionJHolloway, line 69: Will it be saved as parquet or csv?

jobs/reconciliation.py
```diff
@@ -0,0 +1,83 @@
...
+        ),
+        (
+            "--reconciliation_parents_destination",
+            "Destination s3 directory for reconciliation parquet parents dataset",
```
MarionJHolloway, line 73: Again, parquet or csv?

tests/test_file_data.py
```diff
@@ -1770,6 +1770,50 @@ class CleanIndCQCData:
     ]
 
 
+@dataclass
+class ReconciliationData:
+    # fmt: off
+    input_ascwds_workplace_rows = [
+        (date(2024, 4, 1), "100", "A100", "0", None, "100", "2", "1", "1", None, "10", "Est Name 00", "1"),  # Single - not CQC regtype - INCLUDED
```
MarionJHolloway, line 1777: Good use of comments  :)

tests/unit/test_reconciliation.py
```diff
@@ -0,0 +1,55 @@
...
+            Schemas.input_cqc_location_api_schema,
+        )
+        self.test_clean_ascwds_workplace_df = self.spark.createDataFrame(
+            Data.input_ascwds_workplace_rows,
+            Schemas.input_ascwds_workplace_schema,
+        )
+
+        warnings.simplefilter("ignore", ResourceWarning)
+
+    def tearDown(self):
```
MarionJHolloway, line 32: Do we need this? I thought it was for the test file generator? If we can remove, there are imports that can go with it.

MarionJHolloway (approved):
Looks good. A few things to check before you merge though

jobs/reconciliation.py
garypolzin: import_date wasn't in the uncleaned list, but I've added from pipeline Keys instead and changed to unclean

jobs/reconciliation.py
garypolzin: not sure? it'll end up saving two CSV files but that'll take ages to save as it coalesces into one file

I could add a write to parquet but we'll have to remove it

jobs/reconciliation.py
garypolzin: updated :)

tests/test_file_data.py
garypolzin: took me absolutely ages to make this file!

jobs/reconciliation.py
MarionJHolloway: Fair enough, leave it for now then

tests/unit/test_reconciliation.py
garypolzin: oh I left in by accident but we might need

We have this in place for the test for write_to_parquet - I copied that for write to csv

garypolzin: I'll remove for now anyway, can review when I add the saving to csv step

MarionJHolloway: Yeah, that's because the test saves it to a temp directory. We might need it for the csv one, depends how we test it.

garypolzin: is there another way to test the write_to_csv works without doing this?

MarionJHolloway: potentially. We know the write function is tested in pyspark source code, so we are testing that it's going to the write place, so we could potentially mock/ stub it.

PR #363, opened 2024-04-05
Reconciliation prepare cqc data

garypolzin:
# Description
Reconciliation - preparing the CQC data ready for merging with ASCWDS

We need the most recent file and needs registered and deregistered locations so we can't use the 'cleaned' version (registered only)

We also need to auto define dates so that and data in the 'current' month is excluded and any data older than the previous month is also excluded from single/sub accounts

Also created a utils function to filter to the max value in a column - added to ONS script and also removed a test file generator which created an actual file

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -4,8 +4,7 @@ import shutil
 import unittest
 from io import BytesIO
 from enum import Enum
-from pyspark.sql import DataFrame
-from pyspark.sql import functions as F
+from pyspark.sql import DataFrame, functions as F
 from pyspark.sql.types import (
     StructField,
     StructType,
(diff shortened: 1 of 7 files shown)


tests/unit/test_utils.py
```diff
@@ -18,13 +17,10 @@
 import boto3
 from botocore.stub import Stubber
 from botocore.response import StreamingBody
-from tests.test_file_data import CQCPirCleanedData
-from tests.test_file_schemas import (
-    CQCPPIRCleanSchema,
-)
+from tests.test_file_data import UtilsData, CQCPirCleanedData
+from tests.test_file_schemas import UtilsSchema, CQCPPIRCleanSchema
```
MarionJHolloway, line 21: CQCPPIRCleanSchema looks like a typo?

MarionJHolloway (approved):

PR #366, opened 2024-04-05
Remove ascwds head office accounts without location ids

MarionJHolloway:
# Description
 Add testing for reconciliation function: remove ascwds head office accounts without locations ids

# How to test
unit test passing

# Developer checklist
- [x] Unit test


garypolzin (changes requested):
Sorry I've realised this function should be in the prepare ASCWDS step before joining to CQC data

Test format looks great though so I've copied it into my branch so this branch can be closed

PR #368, opened 2024-04-08
Create reconiliation output for ascwds single and sub accounts

MarionJHolloway:
# Description
Function refactored to allow for testing within each step
Tests written for each sub step

# How to test
Unit test passing

# Developer checklist
- [x] Unit test


tests/test_file_data.py
```diff
@@ -29,6 +29,9 @@
 from utils.reconciliation_utils.reconciliation_values import (
     ReconciliationValues as ReconValues,
 )
+from utils.reconciliation_utils.reconciliation_values import (
+    ReconciliationValues as ReconValues,
+)
```
garypolzin, line 34: probably merge conflict, duplicate of above

tests/unit/test_reconciliation.py
```diff
@@ -277,3 +280,198 @@ def test_filter_removes_rows_where_registration_status_is_deregistered_and_date_
...
+        )
+        self.assertEqual(
+            returned_data[0][ReconColumn.website], expected_data[0][ReconColumn.website]
+        )
+        self.assertEqual(
+            returned_data[0][ReconColumn.item], expected_data[0][ReconColumn.item]
+        )
+        self.assertEqual(
+            returned_data[0][ReconColumn.phone], expected_data[0][ReconColumn.phone]
+        )
```
garypolzin, line 420: what's the benefit of doing every one individually as opposed to the usual catch all
self.assertEqual(returned_data, expected_data)

tests/test_file_data.py
```diff
@@ -1898,6 +1906,247 @@ class ReconciliationData:
...
+            "",
+            "",
+            "",
+            "",
+            "",
+            "",
+            "",
+            "",
+            0,
+            "",
```
garypolzin, line 2138: possible just me but I tend to prefer removing the auto formatting for rows like this, find it must harder to see what's going on compared to seeing it in two sets of 4 rows

garypolzin (changes requested):
couple of minor things, might decide to not to action all of them

tests/unit/test_reconciliation.py
MarionJHolloway: I did each column individually because I thought it is not impossible that someone asks us to change the columns and then knowing which columns work/ break could be useful. If you think that's unlikely to be needed I can simplify

tests/test_file_data.py
MarionJHolloway: Ah sorry, I prefer it in this format because the rows often don't line up well. There's nothing to see here so I don't really mind though!

garypolzin: yea fair enough, just keep as this then

tests/unit/test_reconciliation.py
garypolzin: yea makes sense, probably unlikely but no harm in it being like this, it was more curiousity

garypolzin (approved):
Looks like there's a merge conflict but I'll approve on the basis that all my comments have been addressed :)

PR #371, opened 2024-04-10
Changed regionid replace to the cat label function

garypolzin:
# Developer checklist
- [X] Unit test


MarionJHolloway (commented):
Is there a reason this doesn't use the categorical labels function we already built and tested?

garypolzin:
> Is there a reason this doesn't use the categorical labels function we already built and tested?

probably laziness! Changed it to the cat labels function

jobs/reconciliation.py
```diff
@@ -122,7 +124,12 @@ def prepare_latest_cleaned_ascwds_workforce_data(
     df = utils.filter_df_to_maximum_value_in_column(
         ascwds_workplace_df, AWPClean.ascwds_workplace_import_date
     )
-    df = replace_region_id_labels_for_reconciliation(df)
+    df = cUtils.apply_categorical_labels(
+        df,
+        labels_dict,
```
MarionJHolloway, line 129: Might be worth giving this a more specific name? E.g. reconciliation_labels_dict. I'm being quite picky here though!

MarionJHolloway (approved):
Looks good!

PR #373, opened 2024-04-11
Initial updates on res model

garypolzin:
# Description
We're going to have 3 non res models so initial changes to makes the current model more descriptive and included dormancy as a feature - although the model isn't fully signed off yet so features will need more work in the future

[Branch run successful](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:non-res-model-Ind-CQC-Filled-Post-Estimates-Pipeline:b30e2fe1-a907-42fd-84eb-867e6f58bd08)

# Developer checklist
- [X] Unit test


MarionJHolloway (commented):
What happened to the vectorisation test for non res? It was previously our only way of locking in the column order. I think we want something to check that's coming out correctly, as it's easy to make a small change and mess up all the estimates without realising. Just something equivalent to what's in the care home script

garypolzin:
> What happened to the vectorisation test for non res? It was previously our only way of locking in the column order. I think we want something to check that's coming out correctly, as it's easy to make a small change and mess up all the estimates without realising. Just something equivalent to what's in the care home script

yea this needs to go back in eventually but currently we don't have a non-res model, so there is no feature list to match to. This features job currently isn't used at all, just getting things in place for when it's ready. When we have a model we can reintroduce the test to check that the feature lists match

MarionJHolloway (approved):

PR #374, opened 2024-04-11
Reshape old api data

MarionJHolloway:
# Description
This PR goes along with changes in s3 and scripts in EMR, which move the old API data to datasets with the "_old" suffix and save reformatted raw API data in the current buckets as version 2.0.0. All pipeline data will now be able to draw from this v2 dataset, which includes all our historical snapshots, effectively duplicating v0 and v1 data in the new v2 format.

The old bulk downloads still exist, but now point to the "_old" datasets, and can act as a back up data collection step in case we experience issues with the new job. 

# How to test
Unit tests are passing.
The new bulk download jobs have run successfully in branch, appending to the newly reformatted dataset and then been run in the pipeline with a full copy of the data.
I will run all bulk download jobs and the full pipeline in main after merge as these are significant changes.

# Developer checklist
- [x] Unit test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,10 +174,10 @@ jobs:
       - attach_workspace:
           at: .
 
-      #- aws-s3/sync:
-          #aws-region: AWS_REGION
-          #from: "s3://sfc-default-datasets/"
-          #to: "s3://$(cat terraform/pipeline/datasets_bucket_name)/"
+      - aws-s3/sync:
+          aws-region: AWS_REGION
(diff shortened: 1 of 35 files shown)


garypolzin (changes requested):
looks like you've got all the instances but I think it's worth renaming the scripts to match the domains

bulk_download_cqc_locations.py --> add '_old'
bulk_download_cqc_locations_new.py --> (remove '_new)

garypolzin (approved):
all good to me!

PR #380, opened 2024-04-22
Add cqc authentication (preliminary PR)

MarionJHolloway:
# Description
Using the code left behind from the partner code AWS secret, I've implemented the authentication for the bulk download jobs. these are running here: 
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/add-cqc-authentication-bulk_download_cqc_locations_job/runs 
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/add-cqc-authentication-bulk_download_cqc_providers_job/runs 

I have removed the old test scripts, as they were failing (and the old API has shut down so there is no point fixing them).
I have skipped the integration tests for now, as we need to decide whether to remove these or give circle CI permission to access our authentication key.
I will remove the old scripts and rename everything whilst the jobs run in branch (but wanted to see it working before I did this)

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


jobs/bulk_download_cqc_locations.py
```diff
@@ -1,27 +1,26 @@
+import json
 from datetime import date
 
-from utils import cqc_api_new as cqc
+from utils import cqc_api as cqc
 from utils import aws_secrets_manager_utilities as ars
 from utils import utils
 from schemas.cqc_location_schema import LOCATION_SCHEMA_NEW
```
garypolzin, line 7: should we drop _NEW from the schemas too?

jobs/bulk_download_cqc_locations.py
```diff
@@ -1,27 +1,26 @@
...
+    cqc_api_primary_key_value = json.loads(
+        ars.get_secret(secret_name="cqc_api_primary_key", region_name="eu-west-2")
+    )["Ocp-Apim-Subscription-Key"]
     for paginated_locations in cqc.get_all_objects(
         object_type="locations",
         object_identifier=ColNames.location_id,
-        partner_code=partner_code_value,
+        cqc_api_primary_key=cqc_api_primary_key_value,
     ):
         locations_df = spark.createDataFrame(paginated_locations, LOCATION_SCHEMA_NEW)
```
garypolzin, line 25: should we drop _NEW from the schemas too?

schemas/cqc_location_schema.py
```diff
@@ -14,7 +14,7 @@
     NewCqcLocationApiColumns as NewColNames,
```
garypolzin, line 14: sorry just spotted all these are called 'new' as well, also worth changing?

garypolzin (approved):
Approved as comment is only minor name changing

schemas/cqc_location_schema.py
MarionJHolloway: Agreed, but I haven't addressed this in this PR, just because I don't want to break it when we need it to run tonight. I will put in a separate PR to fix this that we can merge and test tomorrow :)

PR #381, opened 2024-04-22
Schedule cqc api calls

MarionJHolloway:
# Description
Set up a step function to run the cqc api calls and used event bridge scheduler to set it running.

# How to test
Unit tests passing
Scheduler will run in branch overnight - check here: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-2%3A344210435447%3AstateMachine%3Aschedule-cqc-api-calls-Bulk-Download-CQC-API-Pipeline

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/terraform/pipeline/eventbridge.tf
+++ b/terraform/pipeline/eventbridge.tf
@@ -27,7 +27,7 @@ resource "aws_iam_policy" "start_state_machines" {
       {
         Effect   = "Allow"
         Action   = ["states:StartExecution"]
-        Resource = "*"
+        Resource = [aws_sfn_state_machine.ingest_ascwds_state_machine.arn]
       }
     ]
   })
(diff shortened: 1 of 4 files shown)


terraform/pipeline/eventbridge.tf
```diff
@@ -52,6 +52,25 @@ resource "aws_iam_role" "start_state_machines" {
 EOF 
 }
 
+resource "aws_iam_role" "scheduler_execution_role" {
+  name = "${local.workspace_prefix}-scheduler_execution_role"
+
```
garypolzin, line 57: Might be worth adding a description to explain what this is doing (see row 4 of this script)

terraform/pipeline/eventbridge.tf
```diff
@@ -52,6 +52,25 @@ resource "aws_iam_role" "start_state_machines" {
 EOF 
 }
 
+resource "aws_iam_role" "scheduler_execution_role" {
+  name = "${local.workspace_prefix}-scheduler_execution_role"
```
garypolzin, line 56: I also think row 2 in this script might be useful here - I think that means that it only runs in main, not all branches

garypolzin (approved):
All looks good to me!

terraform/pipeline/eventbridge.tf
```diff
@@ -79,3 +79,55 @@ resource "aws_cloudwatch_event_target" "trigger_ingest_ascwds_state_machine" {
...
+    ]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "scheduler" {
+  policy_arn = aws_iam_policy.scheduler.arn
+  role       = aws_iam_role.scheduler.name
+}
+
+resource "aws_iam_policy" "scheduler" {
```
garypolzin, line 104: does this conflict with row 83?

garypolzin (changes requested):
sorry just spotted something

terraform/pipeline/eventbridge.tf
garypolzin: scrap that - I can't read - one is policy and one is role, and we've already spoke about that!

garypolzin (approved):
I was right first time, it's all good!

PR #384, opened 2024-04-25
Split pa filled posts into icb areas

RoyPriceCS50:
# Description
Created a new script which will eventually split pa filled posts into ICB areas. Script is not finished, it only has 1 function which creates are dataframe from the cleaned ons postcost directory.

Created a test script for the new function above, which is failing because of an "attribute error". The function type casts anobject as a dataframe, and has the import Dataframe line at the top... not sure how to solve.

Created some test data and a test schema for that data which isn't being used yet but will be used later to test making a list of ICB areas for each LA area.

Added new function to the glue script. My version is using 2023 and will need updating to 2024 later on.

Trello ticket 
https://trello.com/c/TZCqwuu5

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


MarionJHolloway (approved):
Looks good, well done!

PR #390, opened 2024-04-28
Count postcode by la area

RoyPriceCS50:
# Description

Created a function that adds a column with the count of the number of postcodes per LA area in the cleaned postcodes directory.

Function has two tests:
1 - check if the new column has been added and has the desired name.
2 - check the values in the new column are as expected (I didn't make a separate expected dataframe, I just indexed into the column and checked the value is what I expected.)

Link to trello ticket:
https://trello.com/c/cMlXwQ1v

# How to test

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -58,7 +58,7 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
 def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
     w = Window.partitionBy(
         ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr
-    ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
+    ).orderBy(ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr)
 
     postcode_directory_df = postcode_directory_df.withColumn(
         DPColNames.SUM_POSTCODES_PER_LA,
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -1,4 +1,9 @@
-from pyspark.sql import DataFrame
+from pyspark.sql import (
+    DataFrame,
+    Window,
+)
+
+import pyspark.sql.functions as F
 
 from utils import utils
 
@@ -20,6 +25,8 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
             ONSClean.contemporary_icb,
         ],
     )
+
+    postcode_directory_df = count_postcodes_per_la(postcode_directory_df)
 
     pa_filled_posts_df = utils.read_from_parquet(
         pa_filled_posts_source,
@@ -47,6 +54,18 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         destination,
         mode="overwrite",
         partitionKeys=[DPColNames.YEAR],
+    )
+
+
+def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
+    sum_postcodes_per_la: str = "sum_postcodes_per_la"
+
+    w = Window.partitionBy(
+        [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr]
+    ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
+
+    postcode_directory_df = postcode_directory_df.withColumn(
+        sum_postcodes_per_la, F.count(ONSClean.contemporary_ons_import_date).over(w)
     )
 
 
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -56,7 +56,6 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
 
 
 def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
-
     w = Window.partitionBy(
         ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr
     ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -61,7 +61,7 @@ def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
     ).orderBy(ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr)
 
     postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.COUNT_DISTINCT_POSTCODES_PER_LA,
+        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
         F.size(F.collect_set(ONSClean.postcode).over(w)),
     )
 
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -61,7 +61,8 @@ def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
     ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
 
     postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.SUM_POSTCODES_PER_LA, F.count(ONSClean.postcode).over(w)
+        DPColNames.SUM_POSTCODES_PER_LA,
+        F.size(F.collect_set(ONSClean.postcode).over(w)),
     )
 
     return postcode_directory_df
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -61,7 +61,7 @@ def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
     ).orderBy(ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr)
 
     postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.SUM_POSTCODES_PER_LA,
+        DPColNames.COUNT_DISTINCT_POSTCODES_PER_LA,
         F.size(F.collect_set(ONSClean.postcode).over(w)),
     )
 
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -1,9 +1,8 @@
 from pyspark.sql import (
     DataFrame,
     Window,
+    functions as F,
 )
-
-import pyspark.sql.functions as F
 
 from utils import utils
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -13,18 +13,24 @@
...
     TEST_PA_SOURCE = "some/directory"
     TEST_DESTINATION = "some/directory"
+    SUM_POSTCODES_PER_LA = "sum_postcodes_per_la"
 
     def setUp(self) -> None:
         self.spark = utils.get_spark()
         self.test_sample_ons_rows = self.spark.createDataFrame(
-            ONSTestData.ons_sample_rows_full, schema=ONSTestSchema.full_schema
+            ONSTestData.ons_sample_refactored_contemporary_rows,
+            schema=ONSTestSchema.expected_refactored_contemporary_schema,
```
garypolzin, line 33: we normally use the term 'expected' for known data that we want to compare the returned data to

Rather than creating it from scratch again, you can add the code below into test_file_schemas:
    your_new_schema_name = expected_refactored_contemporary_schema

so it's the same schema, but the name we see in the tests is more relevant

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -21,6 +26,8 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         ],
     )
 
+    postcode_directory_df = count_postcodes_per_la(postcode_directory_df)
```
garypolzin, line 29: good practice to have all of the read_from_parquet steps at the start so move this function (and replace?) # TODO 1 assuming this is that step

(not relevant here but also if you have multiple write_to_parquet function calls keep them all at the end)

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -1,4 +1,9 @@
-from pyspark.sql import DataFrame
+from pyspark.sql import (
+    DataFrame,
+    Window,
```
garypolzin, line 3: you can also add in `functions as F,` into this import as they are all from `pyspark.sql`

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -51,3 +57,30 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+        self.assertTrue(self.SUM_POSTCODES_PER_LA in returned_data.columns)
+
+    def test_count_postcodes_per_la_has_expected_values_in_new_column(
+        self,
+    ):
+        returned_data = job.count_postcodes_per_la(self.test_sample_ons_rows).collect()
+        column_index = job.count_postcodes_per_la(
+            self.test_sample_ons_rows
+        ).columns.index(self.SUM_POSTCODES_PER_LA)
+        print(column_index)
```
MarionJHolloway, line 79: remove?

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -51,3 +57,30 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+        self,
+    ):
+        returned_data = job.count_postcodes_per_la(self.test_sample_ons_rows)
+        self.assertTrue(self.SUM_POSTCODES_PER_LA in returned_data.columns)
+
+    def test_count_postcodes_per_la_has_expected_values_in_new_column(
+        self,
+    ):
+        returned_data = job.count_postcodes_per_la(self.test_sample_ons_rows).collect()
+        column_index = job.count_postcodes_per_la(
```
MarionJHolloway, line 76: You don't need this line - you can just use the column name string in square brackets to access the column. This will work, but is just extra code we don't need.

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -50,6 +57,20 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
 
+def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
+    sum_postcodes_per_la: str = "sum_postcodes_per_la"
+
+    w = Window.partitionBy(
+        [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr]
+    ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
```
garypolzin, line 65: I don't think you need the square brackets here

tests/test_file_data.py
```diff
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from datetime import date
+import datetime
```
garypolzin, line 3: already imported datetime (row above)

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -51,3 +57,30 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
             mode="overwrite",
             partitionKeys=[DPColNames.YEAR],
         )
+
+
+class CountPostcodesPerLA(SplitPAFilledPostsIntoICBAreas):
```
MarionJHolloway, line 62: You will also need a test to check behaviour if we have windows with duplicate postcodes. Currently, the function would count the duplicates, which is a behaviour we don't want (we'd be counting the same geographical area twice)

tests/test_file_data.py
```diff
@@ -616,6 +617,17 @@ class ONSData:
...
+        ("AB10AC", datetime.date(2024,1,1), "cssr1", "region1", "subicb1", "icb1", "icb_region1", "ccg1", "51.23456", "-.12345", "123", "E010123", "E020123", "Rural village", "E010123", "E020123", "pcon1", "2024", "01", "01", "20240101"),
+        ("AB10AA", datetime.date(2024,1,1), "cssr2", "region1", "subicb2", "icb2", "icb_region2", None, "51.23456", "-.12345", "123", "E010123", "E020123", "Rural village", "E010123", "E020123", "pcon1", "2024", "01", "01", "20240101"),
+        ("AB10AB", datetime.date(2024,1,1), "cssr2", "region1", "subicb2", "icb3", "icb_region2", None, "51.23456", "-.12345", "123", "E010123", "E020123", "Rural village", "E010123", "E020123", "pcon1", "2024", "01", "01", "20240101"),
+        ("AB10AC", datetime.date(2024,1,1), "cssr2", "region1", "subicb2", "icb3", "icb_region2", None, "51.23456", "-.12345", "123", "E010123", "E020123", "Rural village", "E010123", "E020123", "pcon1", "2024", "01", "01", "20240101"),
+        ("AB10AD", datetime.date(2024,1,1), "cssr2", "region1", "subicb2", "icb3", "icb_region2", None, "51.23456", "-.12345", "123", "E010123", "E020123", "Rural village", "E010123", "E020123", "pcon1", "2024", "01", "01", "20240101"),
```
garypolzin, line 629: as we've already got `from datetime import date` you can remove `datetime.` from these rows and just have `date(2024,1,1)`

MarionJHolloway (changes requested):
A really good first attempt! There's a bit more work to do to finish this off, but I'm really impressed with how far along you've managed to get in independently, well done!

Happy to talk through these comments on Tuesday if that would help.

garypolzin (commented):
added a couple of 'good practice' comments but on the whole it's a really impressive PR - it's way more complicated than the early PRs I was putting in and I defo had more comments back!

don't be daunted by the amount of comments either, they'll drop off really quickly :)

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
RoyPriceCS50: Is this expected data?
I put the cleaned postcode directory into a dataframe as "test data".
I'm not making a bunch of separate test_data rows with an an expected result from the function.
I copied stuff from test _clean_cqc_provider_data.

garypolzin: yea that's right, it's not expected data, but the schema has 'expected' in it so I wasn't sure if it was or not

give me a shout if this doesn't make sense!

RoyPriceCS50: I had to make an expected dataframe in the end, so i've got a test dataframe, a returned df and compare that to an expected df.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
RoyPriceCS50: Turns out the window function doesn't allow a count_distinct method.
I get a set from the window, then get the size of that set.

MarionJHolloway: I asked Roy to add it in because if we do have duplicates, I don't want the DPR stuff to break. We don't touch it very often and this will affect the weighting of our estimates. 

BTw - you're solution is very elegant Roy, well done!

RoyPriceCS50: google is my friend

garypolzin: duplicate postcodes will destroy our whole pipeline to be fair! but all good :)

MarionJHolloway: Yeah, I am itching to get the validaiton in!

garypolzin: slightly less important that one anyway as I clean that file before importing into s3

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -50,6 +55,19 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
 
+def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
+    w = Window.partitionBy(
+        ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr
+    ).orderBy([ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr])
+
+    postcode_directory_df = postcode_directory_df.withColumn(
+        DPColNames.SUM_POSTCODES_PER_LA,
```
MarionJHolloway, line 64: Can you rename this column to accurately reflect what's in it?

utils/direct_payments_utils/direct_payments_column_names.py
```diff
@@ -156,6 +158,9 @@ class DirectPaymentColumnNames:
 
     HISTORIC_RATIO: str = "historic_ratio"
 
+    # Split PA filled posts by ICB area
+    SUM_POSTCODES_PER_LA: str = "sum_postcodes_per_la"
```
MarionJHolloway, line 162: Column name will need updating here too

MarionJHolloway (changes requested):
Last few tweaks and you're good :)

utils/direct_payments_utils/direct_payments_column_names.py
RoyPriceCS50: I've renamed the column.

jobs/split_pa_filled_posts_into_icb_areas.py
RoyPriceCS50: Renamed it, Kept missing a few places in the code but passing tests now.

PR #391, opened 2024-04-29
Flatten cqc ratings prep historic data

MarionJHolloway:
# Description
Functionality to reshape the historic ratings data

# How to test
Unit tests passing
Running in branch here: https://github.com/NMDSdevopsServiceAdm/DataEngineering/compare/flatten-cqc-ratings-prep-historic-data?expand=1

### DIFF
--- a/tests/unit/test_flatten_cqc_ratings.py
+++ b/tests/unit/test_flatten_cqc_ratings.py
@@ -113,6 +113,7 @@ class FlattenCurrentRatings(FlattenCQCRatingsTests):
         expected_data = self.expected_df.collect()
         self.assertEqual(returned_data, expected_data)
 
+
 class FlattenHistoricRatings(FlattenCQCRatingsTests):
     def setUp(self) -> None:
         super().setUp()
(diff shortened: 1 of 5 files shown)


tests/unit/test_flatten_cqc_ratings.py
```diff
@@ -114,16 +114,46 @@ def test_flatten_current_ratings_returns_correct_values(self):
...
+
+    def test_flatten_historic_ratings_returns_correct_columns(self):
+        returned_columns = len(self.returned_df.columns)
+        expected_columns = len(self.expected_df.columns)
+        self.assertEqual(returned_columns, expected_columns)
+
+    def test_flatten_historic_ratings_returns_correct_rows(self):
+        returned_rows = self.returned_df.count()
+        expected_rows = self.expected_df.count()
+        self.assertEqual(returned_rows, expected_rows)
```
garypolzin, line 139: is it overkill to test columns and row count as well at the whole dataframe, or is this just to give nicer error messages?

if the columns or rows don't match, the df check will raise that (but granted, probably in a less clear way)

MarionJHolloway: I found it useful while getting the code to pass because the select/explode/select/explode is quite involved. Happy to remove or keep, but it did help with narrowing down errors.

garypolzin: it's not doing any harm to happy to keep, just wasn't sure if it classed as duplicated code/tests so thought I'd ask :)

MarionJHolloway: Well, it could be the right number of rows and columns but the wrong values, so they do test different things, it more like layers of testing than duplication, which is helpful for more complex stuff.

garypolzin (approved):

PR #393, opened 2024-04-29
Flatten cqc ratings ratings for benchmarks

MarionJHolloway:
# Description
Add functionality to create and save ratings for benchmarks dataset

# How to test
unit test passing
run in branch here: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/flatten-cqc-ratings-ratings-fo-flatten_cqc_ratings_job/runs

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


jobs/flatten_cqc_ratings.py
```diff
@@ -71,31 +72,33 @@ def main(
...
+def filter_to_monthly_import_date(df: DataFrame) -> DataFrame:
+    max_import_date = df.agg(F.max(df[Keys.import_date])).collect()[0][0]
     first_day_of_the_month = "01"
     month_and_year_of_import_date = max_import_date[0:6]
     monthly_import_date = month_and_year_of_import_date + first_day_of_the_month
-    cqc_location_df = cqc_location_df.where(
-        cqc_location_df[Keys.import_date] == monthly_import_date
-    )
-    return cqc_location_df
+    df = df.where(df[Keys.import_date] == monthly_import_date)
```
garypolzin, line 100: are we able to think of a more descriptive name for this function/variable - to me 'monthly' implies 'once a month' but it looks like this is filtering to the start of the most recent month only. Can't think of anything worthwhile to suggest unfortunately, best I can do is filter_to_start_of_most_recent_month

jobs/flatten_cqc_ratings.py
```diff
@@ -267,6 +270,61 @@ def create_standard_ratings_dataset(ratings_df: DataFrame) -> DataFrame:
...
+    benchmark_ratings_df = benchmark_ratings_df.withColumn(
+        CQCRatings.good_or_outstanding_flag,
+        F.when(
+            (benchmark_ratings_df[CQCRatings.overall_rating] == CQCRatingsValues.good)
+            | (
+                benchmark_ratings_df[CQCRatings.overall_rating]
+                == CQCRatingsValues.outstanding
+            ),
+            F.lit(1),
+        ).otherwise(F.lit(0)),
```
garypolzin, line 291: Do we want to keep as 1 and 0 or change to be text (assuming this is a SPSS hangover)?
Just conscious of us moving away from numbers towards text strings whilst also introducing 1/0's ourself

tests/test_file_data.py
```diff
@@ -3462,3 +3462,49 @@ class FlattenCQCRatings:
...
+    create_benchmark_ratings_dataset_rows = [
+        ("loc_1", "estab_1", 1, "Good", "2024-01-01", ""),
+        ("loc_2", "estab_2", 0, "Requires improvement", "2024-01-01", ""),
+        ("loc_3", None, 1, "Good", "2024-01-01", ""),
+        ("loc_4", "estab_2", 0, None, "2024-01-01", ""),
+        ("loc_5", None, 0, None, "2024-01-01", ""),
+    ]
+    expected_create_benchmark_ratings_dataset_rows = [
+        ("loc_1", "estab_1", 1, "Good", "2024-01-01"),
+        ("loc_2", "estab_2", 0, "Requires improvement", "2024-01-01"),
```
garypolzin, line 3509: should these dates still be in `import_date` format (without dashes)? shouldn't change how it works at all, just couldn't see where the format changed, or if it should have

tests/unit/test_flatten_cqc_ratings.py
```diff
@@ -52,10 +52,22 @@ def test_main(self, read_from_parquet_patch: Mock, write_to_parquet_patch: Mock)
...
+            ),
+            call(
+                ANY,
+                self.TEST_BENCHMARK_RATINGS_DESTINATION,
+                mode="overwrite",
+            ),
+        ]
+        write_to_parquet_patch.assert_has_calls(
+            expected_write_to_parquet_calls,
+            any_order=True,
```
garypolzin, line 70: this is nice, like it!

tests/unit/test_flatten_cqc_ratings.py
```diff
@@ -335,5 +349,107 @@ def test_create_standard_ratings_dataset_deduplicates_rows(self):
...
+            Data.expected_join_establishment_ids_rows,
+            Schema.expected_join_establishment_ids_schema,
+        )
+        self.returned_df = job.join_establishment_ids(
+            self.test_ratings_df, self.test_ascwds_df
+        )
+        self.expected_df.show()
+        self.expected_df.printSchema()
+        self.returned_df.show()
+        self.returned_df.printSchema()
```
garypolzin, line 422: assume these can go now?

utils/cqc_ratings_utils/cqc_ratings_values.py
```diff
@@ -14,9 +14,16 @@ class CQCRatingsColumns:
     rating_sequence: str = "Rating_sequence"
     reversed_rating_sequence: str = "Reversed_rating_sequence"
     latest_rating_flag: str = "Latest_rating_flag"
+    good_or_outstanding_flag: str = "flag_good_or_outstanding_current_overall_rating"
+    inspection_date: str = "inspection_date"
+    benchmarks_location_id: str = "cqc_location_id"
+    benchmarks_establishment_id: str = "ascwds_establishment_id"
+    benchmarks_overall_rating: str = "overall_rating"
```
garypolzin, line 21: is it ok to have both `"overall_rating"` and `"Overall_rating"` (row 7)?
I'm assuming they don't appear in the same datasets at the same time so probably ok

garypolzin (approved):
I don't massively understand the process so some comments you can probably ignore :)

None of them are enough to block the PR though so in the interest of time I'll approve now and you can choose which to implement/ignore

jobs/flatten_cqc_ratings.py
MarionJHolloway: The output files have a specific structure because they are used in other applications. Changing them might create extra work for other people

tests/test_file_data.py
MarionJHolloway: The raw data has dashes.

utils/cqc_ratings_utils/cqc_ratings_values.py
MarionJHolloway: The two different exports have different naming requirements

PR #401, opened 2024-05-03
Count postcodes per hybrid area

RoyPriceCS50:
# Description
Refactored the count postcodes by la function within job to be more generic. Instead of counting postcodes by just the la column, the function takes a list as an argument so it can be passed one or more columns to group by.

All tests passing.

Link to Trello ticket:    https://trello.com/c/2akbtsSB

# How to test

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -71,13 +71,14 @@ def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
 def create_hybrid_area_column(postcode_directory_df: DataFrame) -> DataFrame:
     postcode_directory_df = postcode_directory_df.withColumn(
         DPColNames.HYBRID_AREA_LA_ICB,
-        F.concat(ONSClean.contemporary_cssr, F.lit("-"), ONSClean.contemporary_icb),
+        F.concat(ONSClean.contemporary_cssr, F.lit(" - "), ONSClean.contemporary_icb),
     )
 
     return postcode_directory_df
 
 
-def count_postcodes_per_hybrid_area(postcode_directory_df: DataFrame) -> DataFrame: ...
+def count_postcodes_per_hybrid_area(postcode_directory_df: DataFrame) -> DataFrame:
+    ...
 
 
 if __name__ == "__main__":
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -35,9 +35,22 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
     # TODO 1 - Create column with count of postcodes by LA.
-    postcode_directory_df = count_postcodes_per_la(postcode_directory_df)
+    postcode_directory_df = count_postcodes_per_list_of_columns(
+        postcode_directory_df,
+        [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
+        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
+    )
 
     # TODO 2 - Create column with count of postcodes by hybrid area.
+    postcode_directory_df = count_postcodes_per_list_of_columns(
+        postcode_directory_df,
+        [
+            ONSClean.contemporary_ons_import_date,
+            ONSClean.contemporary_cssr,
+            ONSClean.contemporary_icb,
+        ],
+        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA,
+    )
 
     # TODO 3 - Create column with ratio.
 
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -55,44 +55,19 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
 
-def count_postcodes_per_la(postcode_directory_df: DataFrame) -> DataFrame:
-    w = Window.partitionBy(
-        ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr
-    ).orderBy(ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr)
+def count_postcodes_per_list_of_columns(
+    postcode_directory_df: DataFrame,
+    list_of_columns_to_group_by: list,
+    new_column_name: str,
+) -> DataFrame:
+    w = Window.partitionBy(list_of_columns_to_group_by).orderBy(
+        list_of_columns_to_group_by
+    )
 
     postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
+        new_column_name,
         F.size(F.collect_set(ONSClean.postcode).over(w)),
     )
-
-    return postcode_directory_df
-
-
-def create_hybrid_area_column(postcode_directory_df: DataFrame) -> DataFrame:
-    postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.HYBRID_AREA_LA_ICB,
-        F.concat(ONSClean.contemporary_cssr, F.lit(" - "), ONSClean.contemporary_icb),
-    )
-
-    return postcode_directory_df
-
-
-def count_postcodes_per_hybrid_area(postcode_directory_df: DataFrame) -> DataFrame:
-    postcode_directory_df = create_hybrid_area_column(postcode_directory_df)
-    w = Window.partitionBy(
-        ONSClean.contemporary_ons_import_date,
-        DPColNames.HYBRID_AREA_LA_ICB,
-    ).orderBy(
-        ONSClean.contemporary_ons_import_date,
-        DPColNames.HYBRID_AREA_LA_ICB,
-    )
-
-    postcode_directory_df = postcode_directory_df.withColumn(
-        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA,
-        F.size(F.collect_set(ONSClean.postcode).over(w)),
-    )
-
-    postcode_directory_df.show()
 
     return postcode_directory_df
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+            in job.count_postcodes_per_list_of_columns(
+                self.test_sample_ons_rows,
+                [
+                    ONSClean.contemporary_ons_import_date,
+                    ONSClean.contemporary_cssr,
+                    ONSClean.contemporary_icb,
+                ],
+                DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA,
+            ).columns
+        )
```
garypolzin, line 96: I think you can avoid two sets of tests here by making the tests a bit more generic 

Test 1 is good, it's checking that the column you want to appear, does appear. You can cut the name a bit shorter to `test_count_postcodes_per_list_of_columns_adds_new_column_with_given_name`

You can use whatever column name you want for this test, any of these would be completely fine
- DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA
- DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA
- "new_column_name"

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
         )
 
 
-class CountPostcodesPerLA(SplitPAFilledPostsIntoICBAreas):
+class CountPostcodesPerListOfColumns(SplitPAFilledPostsIntoICBAreas):
```
garypolzin, line 66: you can run the job in the setUp and then call it in your tests. In setUp you'd put

```
self.df = job.count_postcodes_per_list_of_columns(
                self.test_sample_ons_rows,
                [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
                DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
            )
```

and then in your tests below, instead of running the job each time you can just refer to `self.df`, such as `self.df.columns`

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -28,9 +28,13 @@ def setUp(self) -> None:
...
-        self.exptected_ons_rows = self.spark.createDataFrame(
-            TestData.expected_ons_sample_contemporary_rows,
-            schema=TestSchema.expected_ons_sample_contemporary_schema,
+        self.expected_postcode_count_per_la_rows = self.spark.createDataFrame(
+            TestData.expected_postcode_count_per_la_rows,
+            schema=TestSchema.expected_postcode_count_per_la_schema,
+        )
+        self.expected_postcode_count_per_la_icb_rows = self.spark.createDataFrame(
+            TestData.expected_postcode_count_per_la_icb_rows,
+            schema=TestSchema.expected_postcode_count_per_la_icb_schema,
```
garypolzin, line 37: usually best to create the 'expected' df in the specific test they are used in, or if they are used multiple times (column check, row lenth, data match) you can put them in the setUp for that test class

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+            )
+            .sort([ONSClean.postcode, ONSClean.contemporary_ons_import_date])
+            .collect(),
+            self.expected_postcode_count_per_la_rows.sort(
+                [ONSClean.postcode, ONSClean.contemporary_ons_import_date]
+            ).collect(),
         )
 
-    def test_count_postcodes_per_la_has_expected_values_in_new_column(
+    def test_count_postcodes_per_list_of_columns_has_expected_values_when_grouped_by_la_and_icb(
```
garypolzin, line 114: as above, the test before this one tests the functionality that the postcode count is as expected when it's given a list so I don't think you need this test, it's not testing anything new

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+                [
+                    ONSClean.contemporary_ons_import_date,
+                    ONSClean.contemporary_cssr,
+                    ONSClean.contemporary_icb,
+                ],
+                DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA,
+            ).columns
+        )
+
+    def test_count_postcodes_per_list_of_columns_has_expected_values_when_grouped_by_la_only(
```
garypolzin, line 98: can rename this to something more generic (such as test_count_postcodes_per_list_of_columns_has_expected_values) and drop the test below

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -35,9 +35,22 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
     # TODO 1 - Create column with count of postcodes by LA.
```
garypolzin, line 37: probably best to remove these comments once you've added the code

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -35,9 +35,22 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
...
-    postcode_directory_df = count_postcodes_per_la(postcode_directory_df)
+    postcode_directory_df = count_postcodes_per_list_of_columns(
+        postcode_directory_df,
+        [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
+        DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
+    )
 
     # TODO 2 - Create column with count of postcodes by hybrid area.
```
garypolzin, line 44: as above

garypolzin (commented):
Looks good, quite a few comments but they are all things which I don't think are necessarily needed as opposed to any actual issues

saying that, might be worth waiting to see if Marion agrees with my comments before actioning anything just in case!

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
+                self.test_sample_ons_rows,
+                [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
+                DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
+            ).columns
+        )
+
+    def test_count_postcodes_per_list_of_columns_adds_new_column_with_given_name_when_given_la_icb_name(
+        self,
+    ):
+        self.assertTrue(
```
MarionJHolloway, line 85: Whilst you can code like this in python, it's generally bad practice to get one line of code to do so much heavy lifting. It makes it much harder to read for someone who is looking at your code for the first time. Try to make the lines of code with the assertions just test the assertion and do the other step(s) elsewhere. You can look at our other tests to see what I mean. Happy to to chat this through with you more if you want.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -59,26 +63,70 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
...
             DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA
-            in job.count_postcodes_per_la(self.test_sample_ons_rows).columns
+            in job.count_postcodes_per_list_of_columns(
+                self.test_sample_ons_rows,
+                [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
+                DPColNames.COUNT_OF_DISTINCT_POSTCODES_PER_LA,
+            ).columns
+        )
+
+    def test_count_postcodes_per_list_of_columns_adds_new_column_with_given_name_when_given_la_icb_name(
```
MarionJHolloway, line 82: You don't need to test this twice. You generally don't need to test things more than once, except when the change in input might affect the change in output (e.g. if you're function uses branching logic, or your data has multiple features you want to test. I suggest you make one of these tests more generic and remove the other.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
MarionJHolloway, line 98: For the expected values, you do need two tests as you are passing it a different number of columns to group by, so keep both of these tests.

MarionJHolloway (commented):
Looking good! Just a bit more work needed to tidy up these tests.

I've not locked this PR for changes because I'm off the rest of the week and don't want to block it being merged.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
MarionJHolloway: I think it's worth having at least two tests here, given that the list is a different length.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
MarionJHolloway: See my comment above. Just spotted that this does need renaming though because you are grouping by import date and la, not la only

jobs/split_pa_filled_posts_into_icb_areas.py
MarionJHolloway: Agreed, but if it helps you to keep them until you've completed all the steps, that's fine for now.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
garypolzin: fair enough, I'll mark as resolved

jobs/split_pa_filled_posts_into_icb_areas.py
RoyPriceCS50: I'll leave these until the end.
Easier for me to read when the steps are blocked out for now.

MarionJHolloway (approved):

PR #403, opened 2024-05-07
Create ratio hybrid area to la area postcode counts

RoyPriceCS50:
# Description
Added functionality to job to add a new column which is count of hybrid area postcodes divided by count of LA area postcodes.

All tests passing.

Link to Trello:   https://trello.com/c/FYqVnvAZ

# How to test

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -53,6 +53,12 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
     # TODO 3 - Create column with ratio.
+    postcode_directory_df = create_ratio_between_columns(
```
MarionJHolloway, line 56: Are you planning on using this function in another context?

RoyPriceCS50: No.
I've made the function more specific to this job in latest commit.

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -85,6 +93,20 @@ def count_postcodes_per_list_of_columns(
     return postcode_directory_df
 
 
+def create_ratio_between_hybrid_area_and_la_area_postcode_counts(
+    postcode_directory_df: DataFrame,
+    hybrid_area_postcode_count: str,
```
MarionJHolloway, line 98: This is still not specific to these columns. Someone could put two different columns here and still get the function to run (although it might not work as expected). You need to remove these arguments so they are fixed to make it into a specific function.

RoyPriceCS50: I've actioned this comment in latest commit

tests/test_file_data.py
```diff
@@ -672,6 +672,40 @@ class PAFilledPostsByICBArea:
...
+        ("AB10AB", date(2024,1,1), "cssr2", "icb3",4,3,0.75000), 
+        ("AB10AC", date(2024,1,1), "cssr2", "icb3",4,3,0.75000), 
+        ("AB10AD", date(2024,1,1), "cssr2", "icb3",4,3,0.75000), 
+        ("AB10AA", date(2023,1,1), "cssr1", "icb1",3,3,1.00000),
+        ("AB10AB", date(2023,1,1), "cssr1", "icb1",3,3,1.00000),
+        ("AB10AC", date(2023,1,1), "cssr1", "icb1",3,3,1.00000),
+        ("AB10AA", date(2023,1,1), "cssr2", "icb2",3,1,0.33333), 
+        ("AB10AB", date(2023,1,1), "cssr2", "icb3",3,2,0.66666), 
+        ("AB10AC", date(2023,1,1), "cssr2", "icb3",3,2,0.66666), 
+        ("AB10AC", date(2023,1,1), "cssr2", "icb3",3,2,0.66666),
```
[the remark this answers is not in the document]
garypolzin: you can probably leave it is now as it's testing ok but it's often best to makes test files as simple as possible. in this case it's dividing one col by another, so you only need those 2 columns in the test data (the date, postcode, cssr and icb area columns don't contribute to this test data in anyway

there's also a quirk you might not have come across yet where circleci tests reorder your data for absolutely no reason. You'll see this in most of our test data but we get around this by always including an ID column, and then sorting the data on the ID before asserting that they match

your test data isn't wrong though, it just makes creating test data way quicker for you if your data is just smaller. These 3 rows will test the ratio is working in the same way

```
rows = [
    ('loc1', 3, 3)
    ('loc2', 4, 1)
    ('loc3', 4, 3)
]
```

```
expected_rows = [
    ('loc1', 3, 3, 1.00)
    ('loc2', 4, 1, 0.25)
    ('loc3', 4, 3, 0.75)
]
```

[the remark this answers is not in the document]
garypolzin: although it's actually worth adding the ID column as circleci is weird with the re-sorting so it might randomly start failing this test at some point if it starts sorting it on postcode or whatever itself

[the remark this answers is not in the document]
garypolzin: just noticed you do sort later on based on postcode and date, however these aren't unique (see two rows below) so these could come out in either order

```
        ("AB10AC", date(2023,1,1), "cssr1", "icb1",3,3,1.00000),
        ("AB10AC", date(2023,1,1), "cssr2", "icb3",3,2,0.66666),
```

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -110,3 +110,61 @@ def test_count_postcodes_per_list_of_columns_has_expected_values_when_grouped_by
...
+class CreateRatioBetweenColumns(SplitPAFilledPostsIntoICBAreas):
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.test_sample_rows_with_la_and_hybrid_area_postcode_counts = self.spark.createDataFrame(
+            TestData.sample_rows_with_la_and_hybrid_area_postcode_counts,
+            schema=TestSchema.sample_rows_with_la_and_hybrid_area_postcode_counts_schema,
+        )
+
+        self.returned_ratio_between_hybrid_area_and_la_area_counts = (
```
garypolzin, line 124: as above, end name with '_df'

garypolzin: as we have test classes set up now you coule just call this returned_df if you wanted too

utils/direct_payments_utils/direct_payments_column_names.py
```diff
@@ -164,6 +164,9 @@ class DirectPaymentColumnNames:
     COUNT_OF_DISTINCT_POSTCODES_PER_HYBRID_AREA: str = (
         "count_of_distinct_postcodes_per_hybrid_area"
     )
+    RATIO_HYBRID_AREA_TO_LA_AREA_POSTCODES: str = (
+        "ratio_hybrid_area_to_la_area_postcodes"
+    )
```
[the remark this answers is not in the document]
garypolzin: Not one to correct in this PR but I'm not sure why we capitalise column names in DPR work but not in our main pipeline

Any thoughts @MarionJHolloway?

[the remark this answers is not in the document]
RoyPriceCS50: Fixed these issues in latest commit.

garypolzin (approved):
all looks good to me now and I think all of Marions comments are addressed too, good work!

PR #404, opened 2024-05-07
Write tests for validation utils

MarionJHolloway:
# Description
I've refactored the testing for validation so that the general functionality is tested in as part of the utils script and the job only tests that the job runs. This will make testing subsequent validation jobs easier and will keep the number of tests we need to a minimum.

In a later PR I plan to handle and test the size of dataset calculations.

I have removed the parsing of failures as we will probably not want or need this behaviour generally.

# How to test
Unit tests passing
Run in branch here: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:write-tests-for-validation-uti-Ind-CQC-Filled-Post-Estimates-Pipeline:1644e368-84de-47fb-9a0d-2a622c9a372e

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -3786,3 +3786,7 @@ class FlattenCQCRatings:
         ("loc_1", "estab_1", 1, "Good", "2024-01-01"),
         ("loc_2", "estab_2", 0, "Requires improvement", "2024-01-01"),
     ]
+
+@dataclass
+class ValidationUtils:
+    schema = []
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -3792,12 +3792,32 @@ class FlattenCQCRatings:
 @dataclass
 class ValidationUtils:
     size_of_dataset_rule = {RuleName.size_of_dataset: 3}
-    size_of_dataset_rows = [
+    size_of_dataset_success_rows = [
         ("loc_1",),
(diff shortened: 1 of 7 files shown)


tests/test_file_data.py
```diff
@@ -36,6 +36,7 @@
 from utils.cqc_ratings_utils.cqc_ratings_values import (
     CQCRatingsValues,
 )
+from utils.validation.validation_rule_names import RuleNames as RuleName
```
garypolzin, line 39: is this required?

utils/validation/validation_utils.py
```diff
@@ -1,3 +1,8 @@
+import os
+import sys
```
garypolzin, line 2: is sys required?

garypolzin (approved):
couple of imports I'm not sure are required but otherwise all looks good!

tests/test_file_data.py
MarionJHolloway: Yes

utils/validation/validation_utils.py
MarionJHolloway: Nope, removed thanks!

PR #405, opened 2024-05-09
Drop duplicates from hybrid and la postcode count table

RoyPriceCS50:
# Description
Need to drop duplicate rows after creating a proportion of postcodes for each ICB within each LA area.
So the output is a table which has each ICB within an LA area and that ICB's % of the total postcodes in the LA.

Link to Trello:    https://trello.com/c/JBJFgqqR

# How to test

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -109,7 +109,7 @@ def create_ratio_between_hybrid_area_and_la_area_postcode_counts(
 
 def deduplicate_ratio_between_hybrid_area_and_la_area_postcode_counts(
     postcode_directory_df: DataFrame,
-):
+) -> DataFrame:
     postcode_directory_df = postcode_directory_df.drop(
         ONSClean.postcode,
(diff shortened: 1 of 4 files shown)


jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -60,6 +60,11 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
     # TODO 4 - Drop duplicates.
+    postcode_directory_df = (
+        deduplicate_ratio_between_hybrid_area_and_la_area_postcode_counts(
```
garypolzin, line 64: sorry should have picked this up on the last PR but we changed the name of the column from 'RATIO....' to 'PROPORTION_OF...' so we should also change the function name so they are in sync

also applies to the function on row 98/57

garypolzin (changes requested):
Good PR this one Roy!
Only a few comments, one is something I should have noticed last time round but all minor stuff :)

RoyPriceCS50:
I've fixed the issues in latest commit

garypolzin (approved):
All good with me!

PR #410, opened 2024-05-13
Add check for categorical value columns

MarionJHolloway:
# Description
Add check and tests for categorical value columns
Add categorical value rules to Merged Ind CQC validation

# How to test
Unit tests passing
Run in branch here: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:validation-add-is-contained-in-Ind-CQC-Filled-Post-Estimates-Pipeline:c16a17f7-f036-4d74-9a7a-33a11d727b6b

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


utils/validation/categorical_column_values.py
```diff
@@ -0,0 +1,184 @@
...
+        CQCLValues.non_residential,
+    ]
+    care_home_values = ["Y", "N"]
+    dormancy_values = ["Y", "N"]
+
+
+class ONSCategoricalValues:
+    rural_urban_indicators = list(RURAL_URBAN_INDICATOR_LOOKUP.values())
+    regions = list(REGION_LOOKUP.values())
+    cssrs = [
```
garypolzin, line 26: I think categorical values are best done at source and this job just needs to check that columns are fully populated?

I'm just thinking about next time a new LA is created or if we accidentally change from '&' to 'and', the merge process feels like the wrong time to find that out / we'll never think to update the merge validation when new geographies appear

garypolzin: appreciate those validations don't exist yet though, so maybe the thinking is that they exist here until every job has it's own validations, then we move out of here?

MarionJHolloway: Yes, I haven't worked out where these will eventually and I think we need a "single source of truth" that we can update and it will update the validation and the features jobs as well. THere's a few bits like this I plan to refactor later on in this process - this script is just so I don't have a mess as I build out the other validation jobs. These validation checks will run on each dataset, so will be checked earlier in the pipeline too and we could also run them on the raw data if we wanted. 

In short, yes this is a WIP script that will probably be refactored at a later date but I think is good enough for now if you agree?

garypolzin (approved):

PR #413, opened 2024-05-17
Apply ratio to calc icb filled posts

RoyPriceCS50:
# Description
Added a function to create a new column with the estimated number of PA's in each ICB. The column is the estimate of PA's in each CSSR multiplied by the proportion of postcodes per ICB within each CSSR.

There is an issue with the join, leading to blank rows in the joined columns. The issue is caused by the CSSR names being different in the PA estimates to the cleaned postcode file.
One of the joined columns is used as a partition key when writing to parquet, and because this has blanks, this causes the job to run very slowly (but succeeds).

# How to test
Tests are passing.
It runs successfully in branch, but takes ~25 minutes.
Link to run:
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/apply-ratio-to-calc-icb-filled-split_pa_filled_posts_into_icb_areas_job/runs

Link to Trello:    https://trello.com/c/CiBeY4ja

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -61,7 +61,7 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         )
     )
 
-    postcode_directory_df = (
+    proportion_of_postcodes_per_hybrid_area_df = (
         deduplicate_proportion_between_hybrid_area_and_la_area_postcode_counts(
             postcode_directory_df
         )
@@ -70,16 +70,21 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     pa_filled_posts_df = create_date_column_from_year_in_pa_estimates(
         pa_filled_posts_df
     )
-    postcode_directory_df = join_pa_filled_posts_to_hybrid_area_proportions(
-        postcode_directory_df, pa_filled_posts_df
+
+    proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df = (
+        join_pa_filled_posts_to_hybrid_area_proportions(
+            proportion_of_postcodes_per_hybrid_area_df, pa_filled_posts_df
+        )
     )
 
-    postcode_directory_df = apply_icb_proportions_to_pa_filled_posts(
-        postcode_directory_df
+    proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df = (
+        apply_icb_proportions_to_pa_filled_posts(
+            proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df
+        )
     )
 
     utils.write_to_parquet(
-        postcode_directory_df,
+        proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df,
         destination,
         mode="overwrite",
         partitionKeys=[ONSClean.contemporary_cssr],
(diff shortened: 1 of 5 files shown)


tests/test_file_data.py
```diff
@@ -754,6 +754,20 @@ class PAFilledPostsByICBArea:
         (date(2022,5,1), "Leeds", "icb1", 1.00000, 300.3, "2023"),
         (date(2022, 5, 1), "Barking & Dagenham", "icb4", 1.00000, None, None),
     ]
+
+    sample_proportions_and_pa_filled_posts_rows = [
+        (1.00000, 100.2),
```
MarionJHolloway, line 759: What are you testing that is unique with each of the top three rows of this data frame? You might want to consider testing the output if none exists in the first column or both columns.

jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -57,35 +55,34 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
...
     )
 
-    # TODO 6 - Apply ratio to calculate ICB filled posts.
+    postcode_directory_df = apply_icb_proportions_to_pa_filled_posts(
+        postcode_directory_df
+    )
 
     utils.write_to_parquet(
-        pa_filled_posts_df,
+        postcode_directory_df,
```
MarionJHolloway, line 82: At some point, this stops really being the postcode directory anymore! Now that most of the job is coded up, have a think about how you want to name your variables in main. Try to create a variable name that describes what you are actually outputting here.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -51,7 +51,7 @@ def test_main(self, read_from_parquet_mock: Mock, write_to_parquet_mock: Mock):
             ANY,
             self.TEST_DESTINATION,
             mode="overwrite",
-            partitionKeys=[DPColNames.YEAR],
+            partitionKeys=[ONSClean.contemporary_cssr],
```
MarionJHolloway, line 54: Note - this is a temporary change to allow the data to save currently. This will be reverted in a later PR.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -328,3 +328,64 @@ def test_join_pa_filled_posts_to_hybrid_area_proportions_has_expected_values(
             ).collect()
         )
         self.assertEqual(returned_rows, expected_rows)
+
+
+class ApplyIcbProportionsToPAEstimates(SplitPAFilledPostsIntoICBAreas):
```
MarionJHolloway, line 333: Standardise capitalisation (Icb vs. ICB)

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
```diff
@@ -328,3 +328,64 @@ def test_join_pa_filled_posts_to_hybrid_area_proportions_has_expected_values(
...
+        self,
+    ):
+        returned_rows = (
+            self.returned_apply_icb_proportions_to_pa_filled_posts_df.collect()
+        )
+        expected_rows = (
+            self.expected_apply_icb_proportions_to_pa_filled_posts_df.collect()
+        )
+
+        for i in range(len(returned_rows)):
```
MarionJHolloway, line 381: Why did you choose to write this as a loop over lines instead of comparing one data frame to another?

MarionJHolloway (changes requested):
A few small changes to make and a couple of questions about your code :) This was generally really good though, well done!

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
RoyPriceCS50: It was how I tested during previous approved PR on CreateRatioBetweenColumns.
I've changed it in this test, but not in the previous test in CreateRatioBetweenColumns.
I tried it in the previous test, but kept getting a 'failed with errors' because unsupported types list and list, even though in both tests they use lists.

RoyPriceCS50:
I've actioned the comments in my latest commit.

tests/unit/test_dpr_split_pa_filled_posts_into_icb_areas.py
MarionJHolloway: It's not wrong to use the for loop. If you want to be able to check each row separately, it works quite well, particularly if you made the string into an f-string that told you which row was failing. Have a think about both tests and decide if checking each row separately is useful in each case. If you're getting errors let me know and we can debug together.

PR #414, opened 2024-05-20
Fix bug with multiple checks

MarionJHolloway:
# Description
These changes fix a bug where adding multiple checks of the same type was producing incorrect results.

The bug occurred when constraints of the same type, but with different integer values were added to checks and produced different results depending on the order in which they were checked. By adding each constraint as a separate check to the run, I was able to avoid this problem.

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/fix-bug-with-multiple-checks-validate_merged_ind_cqc_data_job/runs

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -4289,6 +4289,13 @@ class ValidationUtils:
         }
     }
 
+    distinct_values_multiple_columns_rule = {
+        RuleName.distinct_values: {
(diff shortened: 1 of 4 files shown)


garypolzin (approved):
keeping me on my toes with the show functions!
otherwise all good!

PR #443, opened 2024-06-07
Fix raw worker data

MarionJHolloway:
# Description
Add function to fix problem with duplicated raw worker data
Remove unique indexes from raw worker data rules

# How to test
Unit tests passing
Run successfully in branch: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/fix-raw-worker-data-validate_ascwds_worker_cleaned_data_job/runs
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/fix-raw-worker-data-validate_ascwds_worker_raw_data_job/runs

NB: remaining validation errors are due to new job roles that are available but not in the data yet


garypolzin (approved):
looks good, easy to follow

only thing I'm unsure of is if we need to make a note of why we're removing this specific entry, but not sure what the best way to do that is?

PR #445, opened 2024-06-07
Fix raw pir data

MarionJHolloway:
# Description
Fix problems with pir data relating to index columns and people directly employed values

# How to test
Unit tests passing
Run successfully in branch


tests/unit/test_raw_data_adjustments.py
```diff
@@ -58,3 +58,50 @@ def test_remove_duplicate_worker_in_worker_data_does_not_remove_rows_when_they_d
...
+        )
+        self.test_with_single_row_to_remove_df = self.spark.createDataFrame(
+            Data.pir_data_with_single_row_to_remove, Schemas.pir_data_schema
+        )
+        self.test_without_rows_to_remove_df = self.spark.createDataFrame(
+            Data.pir_data_without_rows_to_remove, Schemas.pir_data_schema
+        )
+        self.expected_df = self.spark.createDataFrame(
+            Data.expected_pir_data, Schemas.pir_data_schema
+        )
```
RoyPriceCS50, line 77: I dont understand why this data/schema combo works.
The schema is all stringtype, and the expected_pir_data has an integer/None as the second to last element in it's rows.

tests/unit/test_clean_cqc_pir_data.py
```diff
@@ -63,6 +66,23 @@ def test_main(
...
+            Data.remove_rows_missing_people_directly_employed,
+            Schemas.remove_rows_missing_people_directly_employed_schema,
+        )
+        expected_df = self.spark.createDataFrame(
+            Data.expected_remove_rows_missing_people_directly_employed,
+            Schemas.remove_rows_missing_people_directly_employed_schema,
+        )
+        returned_df = job.remove_rows_without_people_directly_employed(test_df)
+        expected_df.show()
+        returned_df.show()
```
garypolzin, line 82: remove :)

tests/unit/test_raw_data_adjustments.py
MarionJHolloway: Good spot! That was a mistake (it should be a string type in both). THe reson it won't have failed is because the integer will have been coerced into a string when the dataframe was created.

garypolzin (approved):

PR #447, opened 2024-06-11
Improve dpr to icb job

RoyPriceCS50:
# Description
Made two quality changes to this job.

Changed the column name for the PA estimates by ICB to be "by hybrid area" which is the correct name. The final dataframe isn't a unique list of ICB's, it's hybrid area.

Dropped the original column of PA estimates by LA area. It has duplicate values and it's no longer needed.

Changed the LA name fixing function to use a dict instead of two lists. 

Links to Trello:
https://trello.com/c/KTaqnAbR
https://trello.com/c/KhAh7ncX

# How to test
Runs successfully in branch in 2 minutes.
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/improve-dpr-to-icb-job-split_pa_filled_posts_into_icb_areas_job/runs

I checked that the correct column had been dropped in Athena.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -86,12 +86,6 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         )
     )
 
-    proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df = (
-        proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df.drop(
-            DPColNames.ESTIMATED_TOTAL_PERSONAL_ASSISTANT_FILLED_POSTS
-        )
-    )
-
     utils.write_to_parquet(
         proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df,
         destination,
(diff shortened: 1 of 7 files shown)


jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -86,6 +86,12 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         )
     )
 
+    proportion_of_postcodes_per_hybrid_area_with_pa_filled_posts_df = (
```
MarionJHolloway, line 89: I agree this is worth doing, but can you make it part of the function `apply_icb_proportions_to_pa_filled_posts`? It should be easy to adjust the unit tests to account for this and that way it' included in test coverage.

utils/direct_payments_utils/direct_payments_configuration.py
```diff
@@ -61,20 +61,11 @@ class EstimatePeriodAsDate:
...
-
-    LIST_OF_CORRECT_LA_NAMES = [
-        "Bath and North East Somerset",
-        "Blackburn with Darwen",
-        "Bournemouth Christchurch and Poole",
-        "East Riding of Yorkshire",
-        "Medway",
-        "Southend on Sea",
-    ]
+    DICT_TO_CORRECT_LA_NAMES = {
```
MarionJHolloway, line 64: Thanks - the other advantage this way is that you can link up multiple incorrect keys to the same correct value, which might be useful in the future.

MarionJHolloway (changes requested):
One small tweak and it's good to go :)

PR #454, opened 2024-06-14
Impute missing provider data

MarionJHolloway:
# Description
Add a function that will fill in blanks in provider api columns after they have been joined into locations data. These exist because the location appears in the location api earlier than the corresponding provider appears in the provider api.

The function finds the earliest value and imputes that, in case the value has changed over time.
https://trello.com/c/wQm9L4vZ/498-impute-missing-early-provider-data-in-cleaned-locations-dataset

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/impute-missing-provider-data-clean_cqc_location_data_job/runs

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -251,13 +251,16 @@ def allocate_primary_service_type(df: DataFrame):
 def impute_missing_data_from_provider_dataset(
     locations_df: DataFrame, column_name: str
 ) -> DataFrame:
+    w = (
+        Window.partitionBy(CQCL.provider_id)
+        .orderBy(CQCLClean.cqc_location_import_date)
+        .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+    )
     locations_df = locations_df.withColumn(
         column_name,
         F.when(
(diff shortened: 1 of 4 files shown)


jobs/clean_cqc_location_data.py
```diff
@@ -104,6 +104,13 @@ def main(
         registered_locations_df, cqc_provider_df
     )
 
+    registered_locations_df = impute_missing_data_from_provider_dataset(
+        registered_locations_df, CQCLClean.provider_name
+    )
+    registered_locations_df = impute_missing_data_from_provider_dataset(
+        registered_locations_df, CQCLClean.cqc_sector
+    )
```
garypolzin, line 112: what causes these functions to be needed?
is it because they weren't in the original CQC provider dataset so are blank after the join?

MarionJHolloway: They were in the original CQC provider dataset, but at a later import date than they are in the CQC locations dataset, so they are blank after the join

garypolzin (approved):

PR #460, opened 2024-06-14
Update ascwds source to bounded values

MarionJHolloway:
# Description
ASCWDS filled posts source checks were failing because the list of values referred to the raw total staff and worker records columns instead of bounded values. This change fixes this.

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/update-ascwds-source-to-bounde-validate_cleaned_ind_cqc_data_job/runs


garypolzin (approved):
should we specify that the other two also use the bounded values too?

PR #467, opened 2024-06-18
Changed postcount function to use count instead of approx count disti…

RoyPriceCS50:
# Description
This function used to use approx_count_distinct to count postcodes per LA then per LA/ICB hybrid. Then make percentages.
Because it's approximate, the percentages often don't sum to 100%. Which means, when PA filled posts get broken down, their sum is not the same as the original overall value.

These need to be exactly the same.

Approx count distinct was used because there may be duplicates in the postcode file. But it's not expected to have duplicates in the postcode directory, and if there were then they would have the same la/icb area allocated to them, so it won't affect the proportions.

https://trello.com/c/7PmtU4dH/501-fix-issue-with-dpr-split-by-icb-count-over-window-is-not-accurate-enough

# How to test
Unit tests passing.
Successful run in Glue ( runtime: 2m14s):
https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/accurate-dpr-to-icb-job-split_pa_filled_posts_into_icb_areas_job/runs

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/split_pa_filled_posts_into_icb_areas.py
+++ b/jobs/split_pa_filled_posts_into_icb_areas.py
@@ -42,6 +42,8 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
         ],
     )
 
+    check_for_duplicate_postcodes(postcode_directory_df)
+
     postcode_directory_df = count_postcodes_per_list_of_columns(
         postcode_directory_df,
         [ONSClean.contemporary_ons_import_date, ONSClean.contemporary_cssr],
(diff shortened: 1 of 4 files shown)


jobs/split_pa_filled_posts_into_icb_areas.py
```diff
@@ -94,6 +94,19 @@ def main(postcode_directory_source, pa_filled_posts_source, destination):
     )
 
 
+def check_for_duplicate_postcodes(
```
MarionJHolloway, line 97: Great function, you will want to call it in your job though! :)

MarionJHolloway (changes requested):
One small tweak...

MarionJHolloway (approved):

PR #482, opened 2024-07-12
Fix registration date

MarionJHolloway:
# Description
There was a bug with the raw data where registration dates in the original csv files were null in the parquet file. This PR is part of the work to fix this problem.

As well as this PR, a new version of the locations_api dataset has been created which now has these registration dates reinstated. The code for this is in my EMR workspace in the Fix Registration Date folder. This dataset is currently saved in this branch. The script will need to be rerun with the latest data and added to main when this PR is merged.

Within this PR:
- Added a new step in imputing registration dates to remove registration dates that are later than the earliest import date. This was causing issues with the models as time_registered had negative values.
- Unit test for the new step (above)
- Updated references to the version of the locations api dataset to version 2.1.0

# How to test
Unit tests passing
Run in branch successfully with full datasets here: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:fix-reg-date-Bulk-Download-CQC-API-Pipeline:6214c3a1-c8fd-4e09-8e73-1b7a4ec5eeb8

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/Rftft0C2/664-join-in-registration-dates-from-raw-csv-files


.circleci/config.yml
```diff
@@ -174,10 +174,11 @@ jobs:
       - attach_workspace:
           at: .
 
-      - aws-s3/sync:
-          aws-region: AWS_REGION
-          from: "s3://sfc-default-datasets/"
-          to: "s3://$(cat terraform/pipeline/datasets_bucket_name)/"
+      # THIS IS ONLY FOR TESTING
```
RoyPriceCS50, line 177: Remove these comments

jobs/clean_cqc_location_data.py
```diff
@@ -160,20 +160,91 @@ def fill_missing_provider_ids_from_other_rows(cqc_df: DataFrame) -> DataFrame:
...
+
+    This function changes registration dates from the imupted_registration_date column to None when they are greater than the import date.
+
+    Args:
+        df (DataFrame): A dataframe of CQC locations data with the column imputed_registration_date containing cleaned registration dates.
+
+    Returns:
+        DataFrame: A dataframe of CQC locations data with the column imputed_registration_date containing registration dates that are less than or equal to than the import date.
+    """
+    min_import_date = "min_import_date"
```
RoyPriceCS50, line 214: Put this in column names script

RoyPriceCS50 (changes requested):
Also spotted that the reconciliation job in IndCQCFilledPost... step function is not version controlled. 
Make a separate ticket for it?

jobs/clean_cqc_location_data.py
MarionJHolloway: This column is only used within this function and then dropped before returning the dataframe, so I kept it out of the main definition of columns in our dataset. We have done similar elsewhere in the code, when the column is only required as a step in a process and not the end result.

RoyPriceCS50 (approved):
Removing the commented code in config.yml is only bit left

PR #484, opened 2024-07-15
Remove specialist colleges

MarionJHolloway:
# Description
This PR adds a function to the clean CQC locations job which removes rows where the primary service type is specialist college.
Unit tests added.
Updates made to validation row count.

# How to test
Unit tests passing
Run in branch here: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:remove-specialist-colleges-Ind-CQC-Filled-Post-Estimates-Pipeline:e59f1b1f-70c2-4c86-af9e-4e013f2bf387

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/BKn6NzwK/496-epic-7-remove-specialist-colleges-from-data-must

### DIFF
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -265,17 +265,17 @@ def allocate_primary_service_type(df: DataFrame):
 
 def remove_specialist_colleges(df: DataFrame):
     """
-    Removes rows where 'Specialist college service' is the first service listed
+    Removes rows where 'Specialist college service' is the only service listed
     in 'services_offered'.
 
-    We do not include locations which are primarily specialist colleges in our
+    We do not include locations which are only specialist colleges in our
     estimates. This function identifies and removes the ones listed in the locations dataset.
 
     Args:
         df (DataFrame): A cleaned locations dataframe with the services_offered column already created.
 
     Returns:
-        (DataFrame): A cleaned locations dataframe with locations which are primarily specialist colleges removed.
+        (DataFrame): A cleaned locations dataframe with locations which are only specialist colleges removed.
     """
     df = df.where(
         (df[CQCLClean.services_offered][0] != Services.specialist_college_service)
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -279,7 +279,7 @@ def remove_specialist_colleges(df: DataFrame):
     """
     df = df.where(
         (df[CQCLClean.services_offered][0] != Services.specialist_college_service)
-        | (F.size(df[CQCLClean.services_offered]) == 0)
+        | (F.size(df[CQCLClean.services_offered]) != 1)
         | (df[CQCLClean.services_offered].isNull())
     )
     return df
--- a/jobs/clean_cqc_location_data.py
+++ b/jobs/clean_cqc_location_data.py
@@ -22,6 +22,7 @@ from utils.column_values.categorical_column_values import (
     LocationType,
     PrimaryServiceType,
     RegistrationStatus,
+    Services,
 )
 from utils.column_names.cleaned_data_files.ons_cleaned import (
     OnsCleanedColumns as ONSClean,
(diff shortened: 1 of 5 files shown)


garypolzin (changes requested):
The current rules for specialist college are that they are removed if it is the 'only' service they offer - I don't know if the ordering of the gac_service_types column implies how much of something they offer (that's not how we allocate primary_service_type anyway)

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -272,16 +275,66 @@ def test_allocate_primary_service_type(self):
...
+        self,
+    ):
+        self.assertLess(
+            self.returned_df.count(), self.test_with_specialist_colleges_df.count()
+        )
+
+    def test_remove_specialist_colleges_does_not_drop_rows_where_specialist_college_is_the_first_service_listed(
+        self,
+    ):
+        self.assertEqual(self.returned_df.collect(), self.expected_df.collect())
```
garypolzin, line 309: probably safest to sort these

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -272,16 +275,66 @@ def test_allocate_primary_service_type(self):
...
+        test_df = self.spark.createDataFrame(
+            Data.test_with_specialist_colleges_not_listed_first_rows,
+            Schemas.remove_specialist_colleges_schema,
+        )
+        returned_df = job.remove_specialist_colleges(test_df)
+        expected_df = self.spark.createDataFrame(
+            Data.expected_with_specialist_colleges_not_listed_first_rows,
+            Schemas.remove_specialist_colleges_schema,
+        )
+        self.assertEqual(returned_df.collect(), expected_df.collect())
```
garypolzin, line 323: sort

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -272,16 +275,66 @@ def test_allocate_primary_service_type(self):
...
+            Data.expected_with_specialist_colleges_rows,
+            Schemas.remove_specialist_colleges_schema,
+        )
+
+    def test_remove_specialist_colleges_has_fewer_rows_when_specialist_colleges_are_present(
+        self,
+    ):
+        self.assertLess(
+            self.returned_df.count(), self.test_with_specialist_colleges_df.count()
+        )
```
garypolzin, line 304: if the function deletes every single row then it will have less rows so it's quite ambiguous, better to test that returned cols match expected like we usually do

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -272,16 +275,66 @@ def test_allocate_primary_service_type(self):
...
+
+    def test_remove_specialist_colleges_does_not_drop_rows_where_specialist_college_is_the_first_service_listed(
```
garypolzin, line 306: might need to specify that this is only relevant when they have multiple services - if they only have one service then college will be the first, and we want to remove them

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -272,16 +275,66 @@ def test_allocate_primary_service_type(self):
...
+        self,
+    ):
+        self.assertEqual(self.returned_df.collect(), self.expected_df.collect())
+
+    def test_remove_specialist_colleges_has_the_same_number_of_rows_where_specialist_college_is_listed_but_not_the_first_service_listed(
+        self,
+    ):
```
garypolzin, line 313: instead of first/not-first, we're actually wanting to know if it's the only/not only

jobs/clean_cqc_location_data.py
```diff
@@ -261,6 +263,28 @@ def allocate_primary_service_type(df: DataFrame):
...
+    Args:
+        df (DataFrame): A cleaned locations dataframe with the services_offered column already created.
+
+    Returns:
+        (DataFrame): A cleaned locations dataframe with locations which are only specialist colleges removed.
+    """
+    df = df.where(
+        (df[CQCLClean.services_offered][0] != Services.specialist_college_service)
+        | (F.size(df[CQCLClean.services_offered]) != 1)
+        | (df[CQCLClean.services_offered].isNull())
```
garypolzin, line 283: this does work but I personally find the tests a bit confusing refering to whether it's the first or not

might be able to change the test wording and keep this function as it is, or change to the below
(keep if 1 service and it's not a college, or keep if more than service, or null)

```
        ((F.size(df[CQCLClean.services_offered]) = 1) & (df[CQCLClean.services_offered][0] != Services.specialist_college_service))
        | (F.size(df[CQCLClean.services_offered]) > 1)
        | (df[CQCLClean.services_offered].isNull())
```

garypolzin (changes requested):
feel free to give me a call if the below doesn't make sense, I clicked something by accident and I can't see half of the comments I made so hopefully they all make sense!

jobs/clean_cqc_location_data.py
MarionJHolloway: Leaving this as is and changing tests for more clarity

garypolzin: yea happy to keep as it is, and this is also the benefit of the doc string isn't it!

garypolzin (approved):
looks great now thanks!

PR #488, opened 2024-07-18
Glue script for estimate diagnostics

RoyPriceCS50:
# Description
Added the diagnostics job to the glue script so I can run it in Glue.

Also added functionality to the diagnostics job to filter to known values for any given dataframe and given column.
Added to the skeleton test for the filter function and added additonal tests.

Trello ticket: https://trello.com/c/buWnc72r

# How to test
Unit tests passing.
Link to sucessful run in Glue:  https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/glue-script-for-estimate-diagn-diagnostics_on_known_filled_posts_job/runs

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


tests/test_file_data.py
```diff
@@ -5164,5 +5164,21 @@ class DiagnosticsOnKnownFilledPostsData:
             10.0,
         ),
     ]
-    filter_to_known_values_rows = []
-    expected_filter_to_known_values_rows = []
+
+    filter_to_known_values_rows = [
+        (
```
MarionJHolloway, line 5169: You probably want to add a row where there is a null value in a different column to demonstrate that this function only filters on the specified column

jobs/diagnostics_on_known_filled_posts.py
```diff
@@ -39,6 +39,9 @@ def main(
     )
 
     # filter to where ascwds clean is not null
+    filled_posts_df = filter_to_known_values(
+        filled_posts_df, IndCQC.ascwds_filled_posts_clean
+    )
```
MarionJHolloway, line 44: Remember that spark uses lazy evaluation, so unless you ask it to give you an output, it will plan the job in glue but not execute it. This isn't something you need to fix here, unless you want to. Just understand that just because you get the green tick in glue, it doesn't mean this has actually run your code on the data!

MarionJHolloway (changes requested):
Looks good, just one row of data I would suggest you add and then this can go through :)

tests/test_file_data.py
MarionJHolloway: Fixed

PR #490, opened 2024-07-19
remove show functions and print schema from tests

garypolzin:
# Description
Spotted some data appearing in circleci so quick PR to remove show functions and print schemas


MarionJHolloway (approved):
Oops!

PR #492, opened 2024-07-19
Create test classes for filter tests

garypolzin:
# Description
Currently the tests for this job are under one class, and the data and schema are in the class for another job.
Haven't changed any functionality other than creating classes for each job and set of tests

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -3034,6 +3034,9 @@ class FilterAscwdsFilledPostsData:
         ("03", date(2023, 1, 1), "Y", 25, 3.0),
     ]
 
+
+@dataclass
+class CareHomeFilledPostsPerBedFilterData:
     # fmt: off
     care_home_filled_posts_per_bed_rows = [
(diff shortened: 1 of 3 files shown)


tests/test_file_data.py
```diff
@@ -3034,6 +3034,9 @@ class FilterAscwdsFilledPostsData:
         ("03", date(2023, 1, 1), "Y", 25, 3.0),
     ]
 
+
+@dataclass
+class CareHomeFilledPostsPerBedFilterData:
```
MarionJHolloway, line 3039: I generally try to name the classes here according to the script they are in (so one unit test script pulls in one dataclass of data and one of schemas) and then use the variable name to describe which test they relate to. You might need to rename the script and unit test script or the dataclasses, depending on which is more clear, but at the moment they look quite different to me.

MarionJHolloway (changes requested):

MarionJHolloway (approved):

PR #494, opened 2024-07-19
Refactor calculate percentile function

garypolzin:
# Description
refactored two functions calculate_standardised_residual_cutoffs and calculate_percentile into one function called calculate_lower_and_upper_standardised_residual_percentile_cutoffs, which does exactly the same thing, but cleaner (I hope!)

Added doc string to new function

Changed from 'care home' to separate primary service types

Added an error if percentile >= 1 (trying to filter out 100%+ of data)

Added many tests, think all eventualities are covered!

[branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:refactor-calculate-percentile--Ind-CQC-Filled-Post-Estimates-Pipeline:a7fd3f1b-4120-4fd6-8257-e2f23f59bf4b)


MarionJHolloway (changes requested):

MarionJHolloway (approved):
I think this looks good, just two small recommendations:
1) seeing as you changed "remove" to "null" in some o the function names, would it make sense to do the same to the scripts themselves as well?
2) probably worth rerunning in branch before merging to check everything is wired up right after changing names/ imports etc.

I've approved so you can merge easily after :)

garypolzin:
> I think this looks good, just two small recommendations:
> 
> 1. seeing as you changed "remove" to "null" in some o the function names, would it make sense to do the same to the scripts themselves as well?
> 2. probably worth rerunning in branch before merging to check everything is wired up right after changing names/ imports etc.
> 
> I've approved so you can merge easily after :)

I've got a follow up PR which renames the scripts themselves and rerunning the branch now before merging :)

PR #509, opened 2024-08-05
Increase outlier removal percentage from 5% to 10%

garypolzin:
# Description
Increased the percentage of values removed from 5% to 10%.
Added tests for the actual numerical values as they were missing

Side effect was the test data only now needs 20 rows to test, not 40


MarionJHolloway (approved):
Looks good :)

PR #510, opened 2024-08-05
Dedup before clean

MarionJHolloway:
# Description
Move the deduplication step before the cleaning step
Remove old diagnostics job

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:dedup-before-clean-Ind-CQC-Filled-Post-Estimates-Pipeline:efb2fb32-cb49-4860-923f-5494103ff0ef

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/JgUVBAE1/706-epic-2-refactor-clean-ind-cqc-filled-posts-so-we-deduplicate-before-cleaning


jobs/diagnostics_on_known_filled_posts.py
```diff
@@ -25,7 +25,7 @@
 estimate_filled_posts_columns: list = [
     IndCQC.location_id,
     IndCQC.cqc_location_import_date,
-    IndCQC.ascwds_filled_posts_clean,
+    IndCQC.ascwds_filled_posts_dedup_clean,
     IndCQC.ascwds_filled_posts_dedup_clean,
```
garypolzin, line 29: this column is included twice

tests/test_file_schemas.py
```diff
@@ -3572,7 +3400,7 @@ class ValidateCleanedIndCqcData:
             StructField(IndCQC.worker_records_bounded, IntegerType(), True),
             StructField(IndCQC.ascwds_filled_posts_source, StringType(), True),
             StructField(IndCQC.ascwds_filled_posts, DoubleType(), True),
-            StructField(IndCQC.ascwds_filled_posts_clean, DoubleType(), True),
+            StructField(IndCQC.ascwds_filled_posts_dedup_clean, DoubleType(), True),
```
garypolzin, line 3403: column included twice

utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -26,15 +26,15 @@ class PartitionKeys:
 class IndCqcColumns:
     absolute_residual: str = "absolute_residual"
     ascwds_filled_posts: str = "ascwds_filled_posts"
-    ascwds_filled_posts_clean: str = ascwds_filled_posts + "_clean"
+    ascwds_filled_posts_dedup: str = ascwds_filled_posts + "_deduplicated"
     ascwds_filled_posts_dedup_clean: str = ascwds_filled_posts + "_clean_deduplicated"
```
garypolzin, line 30: can we flip the last 2 words of the string to be the below please (which matches the order of how it happens and also the order of the reference which we use)

ascwds_filled_posts + "_deduplicated_clean"

utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -26,15 +26,15 @@ class PartitionKeys:
...
     ascwds_filled_posts_source: str = ascwds_filled_posts + "_source"
     ascwds_filtering_rule: str = "ascwds_filtering_rule"
     ascwds_workplace_import_date: str = AWPClean.ascwds_workplace_import_date
     average_absolute_residual: str = "average_absolute_residual"
     average_percentage_residual: str = "average_percentage_residual"
-    avg_residuals_ascwds_filled_posts_clean_dedup_non_res_pir: str = (
-        "avg_residuals_ascwds_filled_posts_clean_dedup_non_res_pir"
+    avg_residuals_ascwds_filled_posts_dedup_clean_dedup_non_res_pir: str = (
+        "avg_residuals_ascwds_filled_posts_dedup_clean_dedup_non_res_pir"
```
garypolzin, line 37: remove the second 'dedup' so it's "dedup_clean" instead of "dedup_clean_dedup"

garypolzin (approved):
Some minor edits but otherwise looks great thanks

utils/column_names/ind_cqc_pipeline_columns.py
MarionJHolloway: We can, but we will need to rebuild the models in emr as they contain this string. Whilst I can update it in the test models, it will break the pipeline models.

garypolzin: I think it's only the training of the models with will need amending, the actual models themselves don't use this variable

and we're still tinkering with the model training so I can sort that bit out

MarionJHolloway: Tested in pipeline and seems to work

PR #515, opened 2024-08-07
Refactor rolling average function to take column names as arguments

garypolzin:
# Description
Refactored the model to:
- take another argument which will be the name of the column to create a rolling average of
- take another argument which will be the name of the final rolling average column name
- instead of creating a flag and then summing the flag, changed to be a rolling count
- created a separate function to do the average calculation (previously an untested step)
- include general improvements to the code (docstrings, test data/tests)

[Trello card](https://trello.com/c/40iS4UdC/715-refactor-refactor-rolling-average-model-so-it-can-be-run-using-any-variable-must)

[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:refactor-rolling-avg-Ind-CQC-Filled-Post-Estimates-Pipeline:8b71a707-f4c4-4ef0-9b04-277650a6cd07)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -8,65 +8,135 @@
 
 
 def model_primary_service_rolling_average(
-    df: DataFrame, number_of_days: int
+    df: DataFrame, column_to_average: str, number_of_days: int, model_column_name: str
```
MarionJHolloway, line 11: Do we want to generate the model column name automatically as part of the model? We have this for extrapolation, so any extrapolation column is named "extrapolation_" + column_name. Probably best to keep consistant across models whichever we choose.

utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -8,65 +8,135 @@
...
+        ),
+    )
+    return df
+
+
 def define_window_specifications(unix_date_col: str, number_of_days: int) -> Window:
+    """
+    Define the Window specification partitioned by primary service column.
+
+    Define the Window specification partitioned by primary service column.
```
MarionJHolloway, line 107: If you don't need a more detailed description, you can just have the summary description, no need to repeat yourself. :)

utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -8,65 +8,135 @@
...
+
+    Args:
+        unix_date_col (str): The name of the column containing unix timestamps.
```
MarionJHolloway, line 110: Do we need this to be an argument? We only have one unix date column anyway, so could hard code it.

tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -54,144 +55,122 @@ def test_only_one_additional_column_returned(self):
...
+        self.rolling_sum_df = self.spark.createDataFrame(
+            Data.rolling_sum_rows, Schemas.rolling_sum_schema
         )
-        self.assertEqual(
-            self.returned_row_object[4][IndCqc.rolling_average_model],
-            self.expected_row_object[4][IndCqc.rolling_average_model],
+        self.returned_rolling_sum_df = job.calculate_rolling_sum(
+            self.rolling_sum_df,
+            IndCqc.ascwds_filled_posts_dedup_clean,
+            88,
```
MarionJHolloway, line 78: Define in test class setup

tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -54,144 +55,122 @@ def test_only_one_additional_column_returned(self):
...
+        self.rolling_count_df = self.spark.createDataFrame(
+            Data.rolling_count_rows, Schemas.rolling_count_schema
         )
-        self.assertEqual(
-            self.returned_row_object[16][IndCqc.rolling_average_model],
-            self.expected_row_object[16][IndCqc.rolling_average_model],
+        self.returned_rolling_count_df = job.calculate_rolling_count(
+            self.rolling_count_df,
+            IndCqc.ascwds_filled_posts_dedup_clean,
+            88,
```
MarionJHolloway, line 115: Define in test class setup

tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -10,29 +10,30 @@
...
 
-
-class RollingAverageModelTests(TestModelPrimaryServiceRollingAverage):
-    def setUp(self):
-        super().setUp()
         self.returned_df = job.model_primary_service_rolling_average(
-            self.estimates_df, 88
+            self.estimates_df,
+            IndCqc.ascwds_filled_posts_dedup_clean,
+            88,
```
MarionJHolloway, line 26: Define in test class setup

tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -54,144 +55,122 @@ def test_only_one_additional_column_returned(self):
...
-        )
-        self.assertEqual(
-            self.returned_row_object[15][IndCqc.rolling_average_model],
-            self.expected_row_object[15][IndCqc.rolling_average_model],
-        )
+        for i in range(len(self.returned_row_object)):
+            self.assertEqual(
+                self.returned_row_object[i][IndCqc.rolling_average_model],
+                self.expected_row_object[i][IndCqc.rolling_average_model],
+                f"Returned row {i} does not match expected",
```
MarionJHolloway, line 65: Nice!

MarionJHolloway (changes requested):
A few suggestions for improvements on this one. Generally looks a lot cleaner than before though! :)

utils/estimate_filled_posts/models/primary_service_rolling_average.py
garypolzin: @MarionJHolloway hmm good point, I'll change it

I'm currently looking into non-res extrapolation based on length of time registered for alongside the current 'rolling average'. The name for that might need to be more descriptive of the process such as 'extrapolation_time_open_for' as opposed to 'extrapolation_ascwds_filled_posts_dedup_clean', but if/when required we can change it so the name is optional - by default it follows the current naming approach but if required you can enter something else

utils/estimate_filled_posts/models/primary_service_rolling_average.py
garypolzin: this was the only surviving function I didn't change haha
good spot

tests/unit/test_model_primary_service_rolling_average.py
garypolzin: @MarionJHolloway I think this is ok to stay as it is? the value is to test the calculations are working as expected, it's not tied to the actual number we choose. These tests should only pass/fail based on the code in the function, the actual number of days we use in the pipeline should be tested in the script where this function is used

tests/unit/test_model_primary_service_rolling_average.py
garypolzin: haha yea, didn't think I needed it until it failed and I had no idea where!

tests/unit/test_model_primary_service_rolling_average.py
garypolzin: or do you just mean define 'days = 88' in the test class setup, not related to the actual job?

garypolzin: I think you meant the latter, I'll change to that!

MarionJHolloway (approved):
Looks good, thanks!

PR #519, opened 2024-08-09
Move interpolation column names to indcqc pipeline list

garypolzin:
# Description
Added the interpolation column references to the main cqc pipeline cols so they are all together


MarionJHolloway (approved):
Looks good, can you just clean up the PR name before you merge so we can understand it in 6 months time :)

PR #527, opened 2024-08-15
Replace sum, count and division with average in rolling average calculation

garypolzin:
# Description
Replaced sum, count and the division of sum/count with the average function
Removed tests/data/schemas for functions no longer required.

# How to test
No new tests required, previous tests are sufficient (end product is the same)
[Run in branch](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:refactor-rolling-avg-Ind-CQC-Filled-Post-Estimates-Pipeline:9b283c3e-aa5f-4886-b30a-d96f94e66915)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -113,25 +51,3 @@ def define_window_specifications(number_of_days: int) -> Window:
         .orderBy(F.col(IndCqc.unix_time).cast("long"))
```
MarionJHolloway, line 51: Could you remove the conversion to long type here as well. Pretty sure we don't need such big numbers for unix time

MarionJHolloway (approved):
Looks good, one addition

PR #529, opened 2024-08-16
Changed date formatting in raw_date_format to yyyyMMdd and removed argument from write to parquet call

RoyPriceCS50:
Changed date formatting in raw_date_format to yyyyMMdd
Removed mode argument from write to parquet call as its redundant.

# Description

# How to test
No unit test but ran sucessfully in branch and the date column created by the function is formatted like this 2020-04-01

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


MarionJHolloway (approved):
Thanks!

PR #540, opened 2024-08-29
Null 999 filled posts

MarionJHolloway:
# Description
Add filter to null locations which have 999 ascwds filled posts
Add unit tests

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:null-999-filled-posts-Ind-CQC-Filled-Post-Estimates-Pipeline:79e345eb-7335-429c-bb78-71a160879c08

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/80m4h0h4/413-filtering-example-workplaces-entering-999


utils/ind_cqc_filled_posts_utils/null_ascwds_filled_post_outliers/null_filled_posts_where_locations_use_missing_data_code.py
```diff
@@ -0,0 +1,32 @@
...
+from utils.ind_cqc_filled_posts_utils.null_ascwds_filled_post_outliers.ascwds_filtering_utils import (
+    update_filtering_rule,
+)
+
+MISSING_DATA_CODE = 999.0
+
+
+def null_filled_posts_where_locations_use_missing_data_code(df: DataFrame) -> DataFrame:
+    """
+    Null rows where locations have 999 in ascwds_filled_posts_dedup_clean.
```
garypolzin, line 16: maybe add a bit more info into the docsting for this as it's a bit more rogue
'999' is used elsewhere in ASCWDS to represent not known, That isn't the case for these variables, however upon investigation, all 999 entries at the time of investigation do not look to be an accurate reflection so we are removing them as DQ

tests/unit/test_null_filled_posts_where_locations_use_missing_data_code.py
```diff
@@ -0,0 +1,48 @@
...
+            Schemas.null_filled_posts_using_missing_data_code_schema,
+        )
+
+    def test_null_filled_posts_where_locations_use_missing_data_code_returns_correct_values(
+        self,
+    ):
+        self.assertEqual(
+            self.expected_df.collect(),
+            self.returned_df.sort(IndCQC.location_id).collect(),
+        )
```
garypolzin, line 48: add a test that missing code is 999, just in case someone changes it by accident

utils/ind_cqc_filled_posts_utils/null_ascwds_filled_post_outliers/null_filled_posts_where_locations_use_missing_data_code.py
```diff
@@ -0,0 +1,32 @@
...
+
+from utils.column_names.ind_cqc_pipeline_columns import (
+    IndCqcColumns as IndCQC,
+)
+from utils.column_values.categorical_column_values import AscwdsFilteringRule
+from utils.ind_cqc_filled_posts_utils.null_ascwds_filled_post_outliers.ascwds_filtering_utils import (
+    update_filtering_rule,
+)
+
+MISSING_DATA_CODE = 999.0
```
garypolzin, line 11: just wondering if there's a clearer name for this, 999 isn't actually the missing data code they should use (it should be -1 or -2), we just suspect that some are using this

proxy_missing_data_code / user_inputted_missing_data_code / invalid_missing_data_code

garypolzin: do we need an explanation next to this 999 to say why it's 999, or possibly move it inside the function which calls it as it's already described in there?

utils/column_values/categorical_column_values.py
```diff
@@ -527,3 +527,4 @@ class AscwdsFilteringRule(ColumnValues):
     filtered_care_home_filled_posts_to_bed_ratio_outlier: str = (
         "filtered_care_home_filled_posts_bed_to_ratio_outlier"
     )
+    contained_missing_data_code: str = "contained_missing_data_code"
```
garypolzin, line 530: you might see this before the other comment but change this if we rename

garypolzin (changes requested):
couple of small tweaks on this, it's more of a rogue rule this one so think we need to make it a bit more explanatory as to why we're doing it

I'm also wondering if not just saying 999 is the right decision, or whether we say 999. This rule will only ever be in place specifically for 999 entries so do we need to make it generic?

utils/ind_cqc_filled_posts_utils/null_ascwds_filled_post_outliers/null_filled_posts_where_locations_use_missing_data_code.py
MarionJHolloway: I can't move this into the function and still be able to test the value (as it will only exist when the function runs). I've added a brief comment instead.

utils/ind_cqc_filled_posts_utils/null_ascwds_filled_post_outliers/null_filled_posts_where_locations_use_missing_data_code.py
```diff
@@ -0,0 +1,39 @@
...
+INVALID_MISSING_DATA_CODE = (
+    999.0  # '999' is used elsewhere in ASCWDS to represent not known.
+)
+
+
+def null_filled_posts_where_locations_use_missing_data_code(df: DataFrame) -> DataFrame:
+    """
+    Null rows where locations have 999 in ascwds_filled_posts_dedup_clean.
+
+    '999' is used elsewhere in ASCWDS to represent not known. Whilst that isn't the case for filled posts (i.e. the raw variables total staff and worker records), upon investigation, all 999 entries at the time of investigation do not look to be an accurate reflection of that location's staff so we are removing them for data quality.
```
garypolzin, line 20: remove `and worker records` from inside the brackets, this is only relevant for totalstaff
(they wouldn't enter 999 individual worker records to imply they don't know, they'd just type '999' into total staff)

tests/unit/test_null_filled_posts_where_locations_use_missing_data_code.py
```diff
@@ -46,3 +46,13 @@ def test_null_filled_posts_where_locations_use_missing_data_code_returns_correct
             self.expected_df.collect(),
             self.returned_df.sort(IndCQC.location_id).collect(),
         )
+
+
+class MissingDataCodeTests(NullFilledPostsUsingMissingDataCodeTests):
```
garypolzin, line 51: update file names (job and test), this class and function name to be 'invalid' missing data code so it all aligns

garypolzin (changes requested):
feel free to push back on adding 'invalid' to all the different places it should appear but it'd be good for everything to be consistent and more explanatory

PR #543, opened 2024-09-04
Change outlier removal percentage from 10% to 5%

garypolzin:
# Description
Following filter investigation we are reducing the filtered values from 10% to 5%

Tests/data updated in line with changes


MarionJHolloway (approved):
Looks good!

PR #551, opened 2024-09-13
Replace PIR column with RUI column in final estimates job

garypolzin:
# Description
We don't use the non-deduped version of PIR and it would be useful to have RUI in the estimates data, so replacing that column

[Job ran successfully against main](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/inc-additional-cols-in-estimat-estimate_ind_cqc_filled_posts_job/runs)


MarionJHolloway (approved):
Looks good

PR #556, opened 2024-09-16
Remove original pir col from estimates job

garypolzin:
# Description
Removed the validation in the estimates job for people_directly_employed as that column is no longer in the dataset (we still validate for people_directly_employed_dedup)

Also alphabetised list, just find it easier!


MarionJHolloway (approved):
Thanks!

PR #559, opened 2024-09-17
Change non-res validation to check size of dataset using carehome instead of primary_service_type

garypolzin:
# Description
Validation for non-res currently filters on primary_service_type but the job (and equiv CH job) all filter on carehome column - amended for consistency

# How to test
[validate non-res exc dormancy job ran ok](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/change-nr-validation-to-careho-validate_non_res_ascwds_without_dormancy_ind_cqc_features_data_job/runs)
validate non-res inc dormancy job is currently failing as I ran against main but we haven't run main since the new features were added so it can't find them

# Developer checklist
- [X] Unit test


MarionJHolloway (approved):
Looks good - don't forget to run main as soon as it's deployed :)

PR #564, opened 2024-09-18
Amend non res feature validation size check to match the job filtering

garypolzin:
# Description
Branch data currently failing validations as services which have never been known are all nulled, and the features script removes null rows from the data. This additional non-null filter ensure that the size will match the expected size

Branch now completing validations

# How to test
[branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:amend-non-res-validation-size-Ind-CQC-Filled-Post-Estimates-Pipeline:d43f3c00-7814-4ec0-b676-b130f2bc50f1)

[trello ticket](https://trello.com/c/vxXoP7fa/778-filter-null-services-from-non-res-validation-size-check)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/jobs/validate_non_res_ascwds_inc_dormancy_ind_cqc_features_data.py
+++ b/jobs/validate_non_res_ascwds_inc_dormancy_ind_cqc_features_data.py
@@ -26,6 +26,7 @@ cleaned_ind_cqc_columns_to_import = [
     IndCQC.location_id,
     IndCQC.care_home,
     IndCQC.dormancy,
+    IndCQC.imputed_gac_service_types,
 ]
 
 
@@ -76,6 +77,7 @@ def calculate_expected_size_of_non_res_ascwds_inc_dormancy_ind_cqc_features_data
     expected_size = cleaned_ind_cqc_df.where(
         (cleaned_ind_cqc_df[IndCQC.care_home] == CareHome.not_care_home)
         & (cleaned_ind_cqc_df[IndCQC.dormancy].isNotNull())
+        & (cleaned_ind_cqc_df[IndCQC.imputed_gac_service_types].isNotNull())
     ).count()
     return expected_size
 
--- a/jobs/validate_non_res_ascwds_inc_dormancy_ind_cqc_features_data.py
+++ b/jobs/validate_non_res_ascwds_inc_dormancy_ind_cqc_features_data.py
@@ -1,8 +1,6 @@
 import os
 import sys
 
-os.environ["SPARK_VERSION"] = "3.3"
-
 from pyspark.sql.dataframe import DataFrame
 
 from utils import utils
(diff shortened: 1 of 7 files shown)


jobs/validate_non_res_ascwds_inc_dormancy_ind_cqc_features_data.py
```diff
@@ -22,13 +18,15 @@
 )
 from utils.validation.validation_rule_names import RuleNames as RuleName
 
+os.environ["SPARK_VERSION"] = "3.3"
```
MarionJHolloway, line 21: So this actually needs to be where it was. SPARK_VERSION needs to be set before pydeequ is imported, else it won't work. The reason everything still works here is because it's also called in the validation utils script, which is above where you've put it. In a way this line is redundant, because it's in the validation utils script. However, I kept it in the jobs because setting up pydeequ was a nightmare that took me and James about a week to figure out! If we change the way these jobs work in the future, having this line in the code will mean they can be edited easily, rather than someone having to go through the same process of figuring out what is wrong (and the error messages are cryptic).

jobs/validate_non_res_ascwds_without_dormancy_ind_cqc_features_data.py
```diff
@@ -22,13 +18,15 @@
 )
 from utils.validation.validation_rule_names import RuleNames as RuleName
 
+os.environ["SPARK_VERSION"] = "3.3"
```
MarionJHolloway, line 21: See other comment. Please move back

tests/test_file_data.py
```diff
@@ -5913,18 +5913,18 @@ class ValidateNonResASCWDSIncDormancyIndCqcFeaturesData:
...
-        ("1-000000002", date(2024, 1, 9), None, "Y"),
-        ("1-000000003", date(2024, 1, 1), CareHome.care_home, "N"),
-        ("1-000000004", date(2024, 1, 1), CareHome.care_home, "N"),
-        ("1-000000003", date(2024, 1, 9), CareHome.not_care_home, "N"),
-        ("1-000000004", date(2024, 1, 9), None, "N"),
-        ("1-000000005", date(2024, 1, 1), CareHome.care_home, None),
-        ("1-000000006", date(2024, 1, 1), CareHome.care_home, None),
-        ("1-000000005", date(2024, 1, 9), CareHome.not_care_home, None),
-        ("1-000000006", date(2024, 1, 9), None, None),
+        ("1-000000001", date(2024, 1, 1), CareHome.care_home, "Y", [{"name": "Name", "description": "Desc"}]),
```
MarionJHolloway, line 5916: Can you re-order these rows so that each column is changing in turn? It just makes it easier to see that every possibility is covered. Currently the penultimate column is changing before the last column.

tests/test_file_data.py
```diff
@@ -5947,18 +5947,18 @@ class ValidateNonResASCWDSWithoutDormancyIndCqcFeaturesData:
...
```
MarionJHolloway, line 5950: Can you re-order these rows so that each column is changing in turn? It just makes it easier to see that every possibility is covered. Currently the penultimate column is changing before the last column.

MarionJHolloway (changes requested):
Just a couple of tweaks

MarionJHolloway (approved):
Looks good, thanks

PR #573, opened 2024-09-30
Backfill relationships

garypolzin:
# Description
Backfills the relationship column to all historical years (pre-API) as imputed relationships and then uses this imputed column to populate the related location column.

# How to test
Outputs in Athena return expected results

[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:backfill-relationships-Ind-CQC-Filled-Post-Estimates-Pipeline:2f9d19be-376b-44d3-a661-82f6d12ee070)

[Trello](https://trello.com/c/mGIMTbzC/795-backdate-relationships-related-locationid-data)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


tests/unit/test_clean_cqc_location_data.py
```diff
@@ -228,6 +228,96 @@ def test_amend_invalid_postcodes(self):
...
+
+    def test_original_relationships_remains_unchanged(self):
+        for i in range(len(self.returned_data)):
+            self.assertEqual(
+                self.returned_data[i][CQCL.relationships],
+                self.expected_data[i][CQCL.relationships],
+                f"Returned value in row {i} does not match original",
+            )
+
+    def test_impute_relationships_returns_expected_imputed_data(self):
```
MarionJHolloway, line 269: Because you are using branching logic in the function, it is worth splitting this test (and it's corresponding data) into one test per branch. The test names can then be something like:  test_impute_relationships_returns_[what is expected]_when_[branch condition]

Splitting it up like this helps make sure that all possible outcomes are covered in the branches. It also helps if this function needs refactoring or breaks at some point in the future to see what cases are working and what aren't.

tests/test_file_data.py
```diff
@@ -1221,6 +1221,550 @@ class CQCLocationsData:
...
+        ("1-001", date(2024, 1, 1), RegistrationStatus.deregistered, None, None),
+        (
+            "1-002",
+            date(2024, 2, 1),
+            RegistrationStatus.deregistered,
+            [
+                {
+                    CQCL.related_location_id: "1-0021",
+                    CQCL.related_location_name: "Name after",
+                    CQCL.type: "HSCA Successor",
```
MarionJHolloway, line 1700: Is this test data correct? It seems to still contain successor entries.

jobs/clean_cqc_location_data.py
```diff
@@ -313,6 +314,96 @@ def amend_invalid_postcodes(df: DataFrame) -> DataFrame:
...
+
+    exploded_df = distinct_df.withColumn(
+        CQCLClean.relationships_exploded,
+        F.explode(CQCLClean.first_known_relationships),
+    )
+    predecessors_df = exploded_df.filter(
+        F.col(f"{CQCLClean.relationships_exploded}.{CQCL.type}") == "HSCA Predecessor"
+    )
+    aggregated_predecessors_df = predecessors_df.groupby(CQCL.location_id).agg(
+        F.collect_set(CQCLClean.relationships_exploded).alias(
```
MarionJHolloway, line 397: Is there a reason to use collect_set over collect_list? collect_set will deduplicate, collect_list won't - is that something we need here?

jobs/clean_cqc_location_data.py
```diff
@@ -453,21 +544,20 @@ def realign_carehome_column_with_primary_service(df: DataFrame):
...
     """
     df = df.withColumn(
         CQCLClean.related_location,
         F.when(
-            F.size(df[CQCL.relationships]) > 0,
+            F.size(F.col(CQCLClean.imputed_relationships)) > 0,
             F.lit(RelatedLocation.has_related_location),
-        ).when(
-            F.size(df[CQCL.relationships]) == 0,
+        ).otherwise(
```
MarionJHolloway, line 560: Does this change affect how nulls are populated? Previously we wanted nulls to be recorded differently to an empty list.

MarionJHolloway (changes requested):
I wasn't entirely sure of the intended behaviour in places. So some of these comments may need actioning, some are just for my understanding :)

jobs/clean_cqc_location_data.py
garypolzin: this is list of previous locations so there shouldn't be any duplicates, but if there are then deduping is fine

a duplicate would be saying "this location used to be location 1-001 and it also used to be location 1-001" which is the same as saying "this location used to be location 1-001"

jobs/clean_cqc_location_data.py
garypolzin: previously an empty list meant 'no relationships' and null meant (we don't know as we didn't collect this data prior to API), but now we can backdate all values then empty and null both mean the same

I did think about imputing an empty array historically - would need to define the schema of the imputed_relationship column and then replace nulls with `F.lit([]).cast(schema)` or something like that

at present we only have a plan to use this column for the Y/N linked location, and we can define what 'definitely related' looks like, so be default everything else is 'definitely not related'

tests/test_file_data.py
garypolzin: splitting the test data out should help clear this up

Deregistered locations can have succesors so it's correct for this to be in the imputed column

tests/unit/test_clean_cqc_location_data.py
garypolzin: done :)

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -228,6 +228,160 @@ def test_amend_invalid_postcodes(self):
         )
 
 
+class ImputeHistoricRelationshipsTests(CleanCQCLocationDatasetTests):
+    def setUp(self) -> None:
+        super().setUp()
+
+    def test_impute_historic_relationships_where_type_is_none_returns_none(self):
```
MarionJHolloway, line 235: Small tweak, but it's clearer to use "when" (think F.when()) than "where" (think F.where()) when referring to branching.

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -228,6 +228,160 @@ def test_amend_invalid_postcodes(self):
...
+            Schemas.impute_historic_relationships_schema,
+        )
+        returned_df = job.impute_historic_relationships(test_df)
+        expected_df = self.spark.createDataFrame(
+            Data.expected_impute_historic_relationships_where_type_is_none_returns_none_rows,
+            Schemas.expected_impute_historic_relationships_schema,
+        )
+        self.assertEqual(returned_df.collect(), expected_df.collect())
+
+    def test_impute_historic_relationships_where_where_type_is_predecessor_returns_predecessor(
```
MarionJHolloway, line 247: "where_where"

jobs/clean_cqc_location_data.py
```diff
@@ -313,6 +314,96 @@ def amend_invalid_postcodes(df: DataFrame) -> DataFrame:
...
+    """
+    Imputes historic relationships for locations in the given DataFrame.
+
+    This function performs the following steps:
+    1. Creates a window specification to partition by location_id and order by cqc_location_import_date.
+    2. Adds a column 'first_known_relationships' with the first non-null relationship for each location.
+    3. Filters the relationships to include only those of type 'HSCA Predecessor'.
+    4. Adds a column 'imputed_relationships' based on the following conditions:
+       - If 'relationships' is not null, use 'relationships'.
+       - If 'registration_status' is 'deregistered', use 'first_known_relationships'.
```
MarionJHolloway, line 327: Could you add a note here to say this includes predecessors and successors, so it's clear that successors are inly dropped from registered locations? It wasn't immediately obvious to me on the first review, but it makes sense now.

MarionJHolloway (approved):
Couple of small tweaks, but I've approved so as to not slow you down

PR #578, opened 2024-10-02
Move get_selected_value to utils

garypolzin:
# Description
This function will be used by both extrapolation and interpolation so moved to the utils script. Not currently used by the pipeline

# How to test
Same functionality/tests as before

# Developer checklist
- [X] Unit test


tests/test_file_data.py
```diff
@@ -4,7 +4,6 @@
 from pyspark.ml.linalg import Vectors
 
 from utils.column_names.capacity_tracker_columns import (
-    CapacityTrackerCareHomeColumns as CTCH,
```
MarionJHolloway, line 7: I assume this isn't being used?

MarionJHolloway (approved):
Looks good

tests/test_file_data.py
garypolzin: correct :)

PR #582, opened 2024-10-04
Create one imputation model column instead of separate extrapolation and interpolation

garypolzin:
# Description
Creates one imputation column which contains the name of the column with nulls and the model used to imputate from

Extrapolation and interpolation are dropped, so instead of having unique names they can just be generic 'interpolation_model' and 'extrapolation_model'.

image

# How to test
Tests passing, not in pipeline yet

# Developer checklist
- [X] Unit test


MarionJHolloway (approved):
Looks good

PR #592, opened 2024-10-17
Remove extrapolation

MarionJHolloway:
# Description
Remove old extrapolation code and associated tests and functions

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:remove-extrapolation-Ind-CQC-Filled-Post-Estimates-Pipeline:39e421eb-1183-4268-8ba8-3863d9e04bbc

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/ASeM40hw/814-remove-old-versions-of-extrapolation-inc-test-data


utils/validation/validation_rules/estimated_ind_cqc_filled_posts_validation_rules.py
```diff
@@ -68,7 +68,7 @@ class EstimatedIndCqcFilledPostsValidationRules:
             IndCqcColumns.current_cssr: CatValues.current_cssr_column_values.count_of_categorical_values,
             IndCqcColumns.current_region: CatValues.current_region_column_values.count_of_categorical_values,
             IndCqcColumns.ascwds_filled_posts_source: CatValues.ascwds_filled_posts_source_column_values.count_of_categorical_values,
-            # IndCqcColumns.estimate_filled_posts_source: CatValues.estimate_filled_posts_source_column_values.count_of_categorical_values,  # temporarily removed as extrapolation rolling model is always used before extrapolation ML model, so that model is never chosen (only finds 8 of the 9 distinct options)
+            # IndCqcColumns.estimate_filled_posts_source: CatValues.estimate_filled_posts_source_column_values.count_of_categorical_values, # temporarily removed whilst working on non res models - imputed_posts_non_res_with_dormancy_model not currently being selected
```
garypolzin, line 71: initial comment is still true (although imputation instead of extrapolation), but the new comment is also true - there's a chance we'll settle on rolling average model first though as it's the smoothest, so might end up needing to edit this rule so that it's checks that there are no invalid options, but it still passes if an allowable categorical option isn't actually used

garypolzin (approved):

utils/validation/validation_rules/estimated_ind_cqc_filled_posts_validation_rules.py
MarionJHolloway: What you're talking about are actually two different rules. One rule is checking if the values are in the list of permitted values, another rule (this one) is checking that all permitted values are present. Hence why this rule is temporarily commented out. We have a trello ticket to reinstate rules we've temporarily removed whilst the models are a WIP once the models are "good enough".

garypolzin: yea so might be the case that we end up deleting this check but too early to know

PR #593, opened 2024-10-17
Impute people directly employed

MarionJHolloway:
# Description
Call imputation for people directly employed column.

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:impute-people-directly-employe-Ind-CQC-Filled-Post-Estimates-Pipeline:5f25e58a-8495-4e59-bd78-4e9b0754a8e3

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/mh8lqvpZ/810-add-imputation-for-pir-data


jobs/estimate_missing_ascwds_ind_cqc_filled_posts.py
```diff
@@ -64,12 +64,19 @@ def main(
...
         care_home=True,
     )
-    """
+
+    estimate_missing_ascwds_df = model_imputation_with_extrapolation_and_interpolation(
+        estimate_missing_ascwds_df,
+        IndCQC.people_directly_employed_dedup,
+        IndCQC.rolling_average_model,
+        IndCQC.imputed_people_directly_employed,
+        care_home=True,
```
garypolzin, line 73: this will only populate for care_homes, I'm guessing we'll want the same but for `care_home=False` too?

it's better to keep them separate instead of changing of the function to bypass the care_home option as that will reintroduce the issue when places switch between CH/NR, as NR avg is a lot higher so it causes a bit dip/rise in posts when they switch. This separated method avoids that happening

garypolzin: might need a separate function to merge the PIR CH and PIR NR together though

garypolzin (changes requested):

jobs/estimate_missing_ascwds_ind_cqc_filled_posts.py
MarionJHolloway: I think I'd intended to only do this for non-res to start with. I must have copied the function but forgot to change that parameter! Seeing as we don't use PIR for care homes, do we need to add that in too?

garypolzin: ah ok, yea just change to be NR then and we can add CH later if we need to (but unlikely)

garypolzin: might be worth putting nonres in the name or something so we know?
`IndCQC.imputed_non_res_people_directly_employed` or something like that?

garypolzin (approved):

PR #595, opened 2024-10-18
Add basic pir filter

MarionJHolloway:
# Description
Sets up code structure for creating filters for PIR data using simple example.
Adds unit testing

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:add-basic-pir-filter-Ind-CQC-Filled-Post-Estimates-Pipeline:f4e9946a-be2f-47f6-bc35-1c9444c18c67

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/lQgXb020/817-add-basic-filter-based-on-pir-ascwds-data


garypolzin (commented):
I'm not sure if we're in a position to implement this yet with some proper investigations

for non-res it's quite important to have data, especially if they are different to the norm

If ASCWDS if 550 and PIR says 650 then both agree that this is a really large place. Similarly if ASCWDS is 4 and PIR is 6 then both agree it's a really small place.

If we simply remove both sets of data then we'll estimate them and they'll end up with around 80 posts

Not sure what the rules will end up as, but think this needs a bit more investigation before implementing

MarionJHolloway:
> I'm not sure if we're in a position to implement this yet without more thorough investigations
> 
> for non-res it's quite important to have data, especially if they are different to the norm
> 
> If ASCWDS if 550 and PIR says 650 then both agree that this is a really large place. Similarly if ASCWDS is 4 and PIR is 6 then both agree it's a really small place.
> 
> If we simply remove both sets of data then we'll estimate them and they'll end up with around 80 posts
> 
> Not sure what the rules will end up as, but think this needs a bit more investigation before implementing

MarionJHolloway:
I'm happy to change the rule to make it into a dummy rule for now that doesn't filter anything? The main thing was getting the code set up so we can implement rules easily for testing. The rule itself was arbitrary.

PR #598, opened 2024-10-21
Remove only permitted value

garypolzin:
# Description
Tightening up the rules on ASCWDS creation to only trust inputted values where totalstaff and worker record count are similar, the rules around keeping a value if only one is provided have been removed

# How to test
[branch run successful](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:remove-only-permitted-value-Ind-CQC-Filled-Post-Estimates-Pipeline:416d4262-181e-4c17-9a0e-0554dd629fe0)

# Developer checklist
- [X] Unit test


tests/test_file_data.py
```diff
@@ -3773,11 +3770,11 @@ class CleanIndCQCData:
         ("1-000001", 0, None, None, None,),
```
MarionJHolloway, line 3770: Can we move these into an appropriate test class (same for the schema)? I realise this was already wrong, but it should be a quick fix. The tests for this need an overhaul anyway, but moving this into a new test class would reduce the risk of unintended interdependency in the future. I think the rest of the data is in the tests themselves, so unlikely for anything outside to reference it.

MarionJHolloway (approved):
I think this looks okay. Do we want to see how much it affects our estimates before we merge?

tests/test_file_data.py
garypolzin: yea, always tricky knowing how much old stuff to change in a PR, can end up being massive!

garypolzin:
> I think this looks okay. Do we want to see how much it affects our estimates before we merge?

yea I'll set running against main data so we can see

garypolzin:
models look good so merging in

PR #607, opened 2024-10-28
Utils function to filter rows to non null values only

garypolzin:
# Description
Created a generic utils function to filter rows to non null values in a column

# How to test
Tests passing for both string and numeric value columns

# Developer checklist
- [X] Unit test

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -794,33 +794,19 @@ class SelectRowsWithNonNullValueTests(UtilsTests):
             UtilsSchema.select_rows_with_non_null_values_schema,
         )
 
-    def test_select_rows_returns_expected_non_null_string_rows(self):
-        returned_string_df = utils.select_rows_with_non_null_value(
-            self.test_df, "column_with_string_nulls"
+    def test_select_rows_returns_expected_non_null_rows(self):
+        returned_df = utils.select_rows_with_non_null_value(
+            self.test_df, "column_with_nulls"
         )
-        expected_string_df = self.spark.createDataFrame(
-            UtilsData.expected_select_rows_with_non_null_string_values_rows,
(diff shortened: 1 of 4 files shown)


tests/unit/test_utils.py
```diff
@@ -786,5 +786,42 @@ def test_select_rows_with_value_does_not_change_columns(self):
...
+    def test_select_rows_returns_expected_non_null_string_rows(self):
+        returned_string_df = utils.select_rows_with_non_null_value(
+            self.test_df, "column_with_string_nulls"
+        )
+        expected_string_df = self.spark.createDataFrame(
+            UtilsData.expected_select_rows_with_non_null_string_values_rows,
+            UtilsSchema.select_rows_with_non_null_values_schema,
+        )
+
+        returned_string_data = returned_string_df.sort("ID").collect()
```
MarionJHolloway, line 806: ID is lower case in the schema.

tests/unit/test_utils.py
```diff
@@ -786,5 +786,42 @@ def test_select_rows_with_value_does_not_change_columns(self):
...
 
+class SelectRowsWithNonNullValueTests(UtilsTests):
+    def setUp(self) -> None:
+        super().setUp()
+        self.test_df = self.spark.createDataFrame(
+            UtilsData.select_rows_with_non_null_values_rows,
+            UtilsSchema.select_rows_with_non_null_values_schema,
+        )
+
+    def test_select_rows_returns_expected_non_null_string_rows(self):
```
MarionJHolloway, line 797: I don't think there's any need to check different datatypes here as that behaviour sits inside "isNotNull()"

MarionJHolloway (approved):
A couple of small changes, but I have approved so as not to slow you down

PR #616, opened 2024-10-31
Create archiving job

MarionJHolloway:
# Description
Set up template for job and tests
Terraform job and add to pipeline

# How to test
Unit tests passing
Branch run: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:create-archiving-job-Ind-CQC-Filled-Post-Estimates-Pipeline:0b5e3a5c-147d-4634-b244-439e28c091bf

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/pnQ6meKl/840-create-archive-file-for-estimates

### DIFF
--- a/terraform/pipeline/glue.tf
+++ b/terraform/pipeline/glue.tf
@@ -179,6 +179,20 @@ module "clean_ons_data_job" {
   }
 }
 
+module "archive_filled_posts_estimates_job" {
+  source          = "../modules/glue-job"
+  script_name     = "archive_filled_posts_estimates.py"
+  glue_role       = aws_iam_role.sfc_glue_service_iam_role
+  resource_bucket = module.pipeline_resources
+  datasets_bucket = module.datasets_bucket
+
+  job_parameters = {
+    "--estimate_ind_cqc_filled_posts_source"     = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=estimated_ind_cqc_filled_posts/"
+    "--monthly_filled_posts_archive_destination" = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=monthly_filled_posts_archive/"
+    "--annual_filled_posts_archive_destination"  = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=annual_filled_posts_archive/"
+  }
+}
+
 
 module "prepare_features_non_res_ascwds_ind_cqc_job" {
   source          = "../modules/glue-job"
--- a/terraform/pipeline/glue.tf
+++ b/terraform/pipeline/glue.tf
@@ -188,8 +188,8 @@ module "archive_filled_posts_estimates_job" {
 
   job_parameters = {
     "--estimate_ind_cqc_filled_posts_source"     = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=estimated_ind_cqc_filled_posts/"
-    "--monthly_filled_posts_archive_destination" = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=monthly_filled_posts_archive/"
-    "--annual_filled_posts_archive_destination"  = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=annual_filled_posts_archive/"
+    "--monthly_filled_posts_archive_destination" = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=archived_monthly_filled_posts/"
+    "--annual_filled_posts_archive_destination"  = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=archived_annual_filled_posts/"
   }
 }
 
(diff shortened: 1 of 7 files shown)


terraform/pipeline/step-functions/IndCqcFilledPostEstimatePipeline-StepFunction.json
```diff
@@ -278,88 +278,122 @@
                   "--ml_model_metrics_destination": "${dataset_bucket_uri}/domain=ind_cqc_filled_posts/dataset=ml_model_metrics/"
                 }
               },
-              "Next": "Run estimate ind cqc filled posts by job role job"
+              "Next": "Run diagnotics and archive data"
```
garypolzin, line 281: amend spelling of diagnotics

terraform/pipeline/step-functions/IndCqcFilledPostEstimatePipeline-StepFunction.json
```diff
@@ -278,88 +278,122 @@
...
+                              "Type": "Task",
+                              "Resource": "arn:aws:states:::states:startExecution.sync:2",
+                              "Parameters": {
+                                "StateMachineArn": "${run_gold_validation_state_machine_arn}",
+                                "Input": {
+                                  "AWS_STEP_FUNCTIONS_STARTED_BY_EXECUTION_ID.$": "$$.Execution.Id"
+                                }
+                              },
+                              "End": true
+                            }
```
garypolzin, line 337: 'Validation and crawler' shouldn't start until 'filled posts by job role' are finished - one of the gold validations is for that job

terraform/pipeline/glue.tf
```diff
@@ -179,6 +179,20 @@ module "clean_ons_data_job" {
...
+  source          = "../modules/glue-job"
+  script_name     = "archive_filled_posts_estimates.py"
+  glue_role       = aws_iam_role.sfc_glue_service_iam_role
+  resource_bucket = module.pipeline_resources
+  datasets_bucket = module.datasets_bucket
+
+  job_parameters = {
+    "--estimate_ind_cqc_filled_posts_source"     = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=estimated_ind_cqc_filled_posts/"
+    "--monthly_filled_posts_archive_destination" = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=monthly_filled_posts_archive/"
+    "--annual_filled_posts_archive_destination"  = "${module.datasets_bucket.bucket_uri}/domain=ind_cqc_filled_posts/dataset=annual_filled_posts_archive/"
```
garypolzin, line 192: do you think we create a new domain folder for archived/published figures

might make stuff easier to point at for those linking tableaus etc if the archived folder contains all the stuff we've published, less chance of linking to the wrong dataset hopefully

garypolzin (changes requested):
Couple of bits to look at, some of them questions :)

terraform/pipeline/glue.tf
MarionJHolloway: We can also delete tables that we no longer use anymore and still keep the data in s3

terraform/pipeline/step-functions/IndCqcFilledPostEstimatePipeline-StepFunction.json
```diff
@@ -291,75 +291,104 @@
...
+                  "States": {
+                    "Run archive filled posts estimates": {
+                      "Type": "Task",
+                      "Resource": "arn:aws:states:::glue:startJobRun.sync",
+                      "Parameters": {
+                        "JobName": "${archive_filled_posts_estimates_job_name}",
+                        "Arguments": {
+                          "--estimate_ind_cqc_filled_posts_source": "${dataset_bucket_uri}/domain=ind_cqc_filled_posts/dataset=estimated_ind_cqc_filled_posts/",
+                          "--monthly_filled_posts_archive_destination": "${dataset_bucket_uri}/domain=ind_cqc_filled_posts/dataset=archived_monthly_filled_posts/",
+                          "--annual_filled_posts_archive_destination": "${dataset_bucket_uri}/domain=ind_cqc_filled_posts/dataset=archived_annual_filled_posts/"
```
garypolzin, line 364: image

if the diagnostics job finishes before the archive job then the crawler won't run on the newly archived data

garypolzin (changes requested):
think the step function needs a little tweak but everything else looks good

terraform/pipeline/step-functions/IndCqcFilledPostEstimatePipeline-StepFunction.json
garypolzin: actually come to think of it, do we want the archiving step as part of the main step function?
The archive is supposed to be copies of the published data so maybe for a while at least we specifically decide when to archive data

Otherwise if we run main several times a day for whatever reason then everyone of them goes into archive. Or if the first run of the month wasn't great, we tweak it and run again, then the 2nd run of the month is the archived version we want to use

if we do include in this pipeline, we'd want it to be after the gold validation (at the moment, we are archiving data regardless of whether there validation data issues or not)

MarionJHolloway: Yeah, good point. I think if we rerun, especially early on, we probably want to store both sets of data temporarily, so maybe deleting the data we don't want afterwards would be best? I can add a timestamp column instead of date column as a partition so it will be easy to delete.

garypolzin: I think the idea of the archived folder is that it's all data we've published

Your idea about saving files is a good one though, but I think that's a separate location - and that can be more short term, save all runs in the past month maybe? that way we can say 'the run from 3 days ago was better than this one so lets go back to that'

garypolzin: the archived folder could even work from that location - the job becomes to copy run X into the archived folder

garypolzin (approved):

PR #618, opened 2024-10-31
Update dv for ch metrics

MarionJHolloway:
# Description
Update the r2 calculation to use the correct column for care homes

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:update-dv-for-ch-metrics-Ind-CQC-Filled-Post-Estimates-Pipeline:fb65961d-f264-4217-b03a-5227cefecded
r2 on test data is now 0.56


tests/test_file_data.py
```diff
@@ -4706,6 +4712,7 @@ class CareHomeFeaturesData:
             "Urban city and town",
             25.0,
             2.5,
+            None,
```
garypolzin, line 4715: I don't suppose you can change this None to 2.5 please so it looks ok if we ever need to run a show function on it?

garypolzin (approved):
One very minor change so I'll approve now :)

PR #621, opened 2024-11-01
Add location id hash

MarionJHolloway:
# Description
Add a hashed location ID to the cqc ratings for data requests

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/add-location-id-hash-flatten_cqc_ratings_job/runs
Column appears in athena

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/pyxv5axJ/839-add-hashed-identifier-to-the-cqc-ratings-file


tests/unit/test_flatten_cqc_ratings.py
```diff
@@ -369,6 +369,87 @@ def test_create_standard_ratings_dataset_deduplicates_rows(self):
...
+        expected_columns = self.expected_df.columns
+        self.assertEqual(returned_columns, expected_columns)
+
+    def test_add_location_id_hash_adds_a_column_with_expected_hashed_ids_when_location_id_has_spaces(
+        self,
+    ):
+        self.assertEqual(
+            self.returned_df.sort(CQCL.location_id).collect(),
+            self.expected_df.collect(),
+        )
```
garypolzin, line 396: We don't need this test (if we did want this as a test, we'd want to remove the space before hashing, which would return a different hashed ID to the one we are testing)

currently the SPSS version needs a step with removes spaces as you need to specify a string width for the column. I've set it to 12 which is the max locationid length, but that means that a locationid of length 10 gets 2 spaces at the end to make it 12 characters. So I need to removed these spaces before hashing

but we don't have that situation in aws so we don't need to remove blanks

garypolzin (approved):
I'll approve now as pretty easy the remove the unrequired test - all the functionality is good :)

PR #627, opened 2024-11-06
Remove duplicate locations in ascwds and cqc for coverage data

garypolzin:
# Description
Creates a generic removes duplicates function which takes a list of columns used to highlight duplicates, plus sorts on another column in descending order (keeps the first value it finds).

In terms of ASCWDS data, this will keeps the most recently updated one (based on highest mupddate date)
In terms of CQC data, this will keep the one which is 'in ascwds' (in ASCWDS=1, not in ASCWDS =0).

If either of them are completely the same (same mupddate or both in/not in ASCWDS, then it doesn't matter which we keep)

# How to test
None of the problematic locationids identified from another branch appear as duplicates anymore

[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:remove-dupes-from-coverage-Ind-CQC-Filled-Post-Estimates-Pipeline:ae1d8854-9cc8-4145-8130-532bd7bf993e)

[Trello care](https://trello.com/c/tV8foJBH/851-remove-ascwds-locationid-duplicates)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -3698,6 +3698,19 @@ class MergeCoverageData:
     ]
     # fmt: on
 
+    remove_duplicate_locationids_rows = [
+        (date(2024, 1, 1), "1-001", date(2023, 1, 1)),
+        (date(2024, 1, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+    ]
+    expected_remove_duplicate_locationids_rows = [
+        (date(2024, 1, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+    ]
+
     # fmt: off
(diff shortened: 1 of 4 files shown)


tests/test_file_data.py
```diff
@@ -3673,45 +3673,54 @@ class MergeIndCQCData:
...
+        (date(2024, 2, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+    ]
+    expected_remove_duplicate_locationids_rows = [
+        (date(2024, 1, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-001", date(2023, 2, 1)),
+        (date(2024, 2, 1), "1-002", date(2023, 2, 1)),
+    ]
+
```
RoyPriceCS50, line 3725: This test data is already sorted on master update (last element), so it's not testing the sort part of the function?
It's dropping the correct rows, so could you just swap the position of first and second rows in "remove_duplicate_locationids_rows".

garypolzin: the sort in the function is in descending order, so for location "1-001" it keeps the second row, not the first

RoyPriceCS50 (approved):

PR #641, opened 2024-11-28
Refactor rolling average

garypolzin:
# Description
Refactored rolling average to use one singular column which is rolling averaged then split out to help with follow up work.

# How to test
No change to functionality, previous tests still passing and [branch still running ok](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:rename-rolling-avg-variables-Ind-CQC-Filled-Post-Estimates-Pipeline:24f569ab-a68a-4f40-87e0-59e15da5d4e6)

# Developer checklist
- [X] Unit test
- [] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_model_primary_service_rolling_average.py
+++ b/tests/unit/test_model_primary_service_rolling_average.py
@@ -94,18 +94,54 @@ class CreateSingleColumnToAverageTests(ModelPrimaryServiceRollingAverageTests):
         self.returned_data = self.returned_df.collect()
         self.expected_data = self.expected_df.sort(IndCqc.location_id).collect()
 
-    def test_model_primary_service_rolling_average_returns_expected_columns(self):
+    def test_create_single_column_to_average_returns_expected_columns(self):
         self.assertEqual(
             sorted(self.returned_df.columns),
             sorted(self.expected_df.columns),
         )
 
-    def test_returned_rolling_average_model_values_match_expected(
+    def test_returned_column_to_average_values_match_expected(
         self,
     ):
         for i in range(len(self.returned_data)):
             self.assertEqual(
-                self.returned_data[i][IndCqc.column_to_average],
-                self.expected_data[i][IndCqc.column_to_average],
+                self.returned_data[i][IndCqc.temp_column_to_average],
+                self.expected_data[i][IndCqc.temp_column_to_average],
+                f"Returned row {i} does not match expected",
+            )
+
+
+class CalculateRollingAverageTests(ModelPrimaryServiceRollingAverageTests):
+    def setUp(self) -> None:
+        super().setUp()
+
+        number_of_days: int = 3
+
+        test_df = self.spark.createDataFrame(
+            Data.calculate_rolling_average_rows,
+            Schemas.calculate_rolling_average_schema,
+        )
+        self.returned_df = job.calculate_rolling_average(test_df, number_of_days)
+        self.expected_df = self.spark.createDataFrame(
+            Data.expected_calculate_rolling_average_rows,
+            Schemas.expected_calculate_rolling_average_schema,
+        )
+        self.returned_data = self.returned_df.collect()
+        self.expected_data = self.expected_df.sort(IndCqc.location_id).collect()
+
+    def test_calculate_rolling_average_returns_expected_columns(self):
+        self.assertEqual(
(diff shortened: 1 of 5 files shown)


tests/unit/test_model_primary_service_rolling_average.py
```diff
@@ -17,7 +17,12 @@ def setUp(self):
         warnings.filterwarnings("ignore", category=ResourceWarning)
         warnings.filterwarnings("ignore", category=DeprecationWarning)
 
-        number_of_days = 88
+
+class MainTests(ModelPrimaryServiceRollingAverageTests):
+    def setUp(self) -> None:
+        super().setUp()
+
+        number_of_days = 89
```
MarionJHolloway, line 25: integer?

utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -1,131 +1,147 @@
...
-    df: DataFrame,
-    care_home_column_to_average: str,
-    non_res_column_to_average: str,
-    window: Window,
-    temp_col: str,
-) -> DataFrame:
+def calculate_rolling_average(df: DataFrame, number_of_days: int) -> DataFrame:
     """
-    Calculates the rolling average of a specified column over a given window of days for care home and non residential separately.
+    Calculates the rolling average of a specified column over a given window of days.
```
MarionJHolloway, line 85: This is still calculating separately by service type - worth mentioning here?

utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -1,131 +1,147 @@
...
 
-    A temporary column is used so that the window functions can be run simultaneously to improve spark performance.
+    Calculates the rolling average of a specified column over a given window of days.
+    One day is removed from the provided number_of_days value to reflect that the range is inclusive.
```
MarionJHolloway, line 88: "One day is removed from the provided number_of_days value because the pyspark range between function is inclusive at both the start and end point. We only want it to be inclusive of the current day."

Think this might be a bit clearer, else it sounds like you want it to be inclusive at both ends.

MarionJHolloway: Also an option to just reduce the number of days we set by one and make a comment there that it is inclusive at both ends instead.

utils/estimate_filled_posts/models/primary_service_rolling_average.py
```diff
@@ -1,131 +1,147 @@
...
-        .rangeBetween(-convert_days_to_unix_time(number_of_days), 0)
+    df = df.withColumn(
+        ratio_rolling_average_model_column_name,
+        F.when(
+            F.col(IndCqc.care_home) == CareHome.care_home,
+            F.col(TempCol.temp_rolling_average),
+        ),
     )
+    df = df.withColumn(
+        posts_rolling_average_model_column_name,
```
MarionJHolloway, line 140: So, doing this will create the column whether we want this to be calculated or not. I don't know if you made these changes to support your rolling average work or my use of the function in the CT pipeline, but I just wanted to be clear that this will cause problems with my current CT PR by adding an unwanted and empty column that I then have to remove (or change this function again!).

MarionJHolloway (changes requested):
A few things to consider before I approve this :)

utils/estimate_filled_posts/models/primary_service_rolling_average.py
garypolzin: this looks like a change on github but it's the same as it was previously (apart from col name changes), the ratio is run first, then the filled posts. You've named them both the same so the second will still overwrite the first

The main branch is:

```
    df = df.withColumn(
        care_home_ratio_column_name,
        F.when(F.col(IndCqc.care_home) == CareHome.care_home, (F.col(temp_col))),
    )
    df = df.withColumn(
        model_column_name,
        F.when(
            F.col(IndCqc.care_home) == CareHome.care_home,
            (F.col(temp_col)) * F.col(IndCqc.number_of_beds),
        ).otherwise(F.col(temp_col)),
    )
```

MarionJHolloway (approved):
Looks good, thanks!

PR #642, opened 2024-11-28
Step 0 - Standardise rolling avg column names

garypolzin:
# Description
Renamed columns to be more standardised

# How to test
[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:rename-rolling-avg-cols-Ind-CQC-Filled-Post-Estimates-Pipeline:c7ffc839-5719-48b9-8f4f-cb432c5c6c04)


tests/unit/test_prepare_features_care_home_ind_cqc.py
```diff
@@ -86,7 +86,7 @@ def test_main_produces_dataframe_with_features(
 
         self.assertTrue(result.filter(F.col(IndCQC.features).isNull()).count() == 0)
         expected_features = SparseVector(
-            51, [0, 9, 10, 17, 24, 31], [10.0, 1.0, 2.5, 1.0, 1.0, 1.0]
+            51, [0, 1, 10, 17, 24, 31], [10.0, 2.5, 1.0, 1.0, 1.0, 1.0]
```
MarionJHolloway, line 89: How come these have changed?

garypolzin: if we don't end up going with the new method I can yea. The imputed ones are called `imputed_posts_rolling_average_model` and `imputed_ratio_rolling_average_model` so I used those names without 'imputed_'

so if I flip them it makes sense to flip the imputed ones too

Will pause for now but that's the solution if not

MarionJHolloway: Did this get addressed in the end?

garypolzin: we're addressing them as we speak by creating new models with the new features

garypolzin: if we were to run the models after this PR alone they will be completely meaningless
but these are the features in advance of the new models

MarionJHolloway (approved):

PR #648, opened 2024-12-02
Validate cleaned capacity tracker non residential data

MarionJHolloway:
# Description
Set up validation for non res cleaned capacity tracker data
Add to pipeline

# How to test
Unit tests passing
Run in branch: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:validate-ct-clean-non-res-IngestAndCleanCapacityTrackerDataPipeline:e6469509-2452-4591-a8f5-1c1e27e742cd

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket: https://trello.com/c/DqPMvCgX/870-validation-for-clean-ct-non-res


garypolzin (commented):
code looks fine for that but do we want to set it up like other jobs whereby the 2 clean data jobs run then the 2 cleaned validations run

the rest of the pipeline won't run if either of those fail

also you'll want to set it up like any of our bronze/silver/gold where the validation crawler runs if it passes or fails

or take step function out of this and have one final job to make a step function once all validations exist?

MarionJHolloway:
I can separate it out like that, should be easy to do. The diagnostics job creates 4 datasets, so it can be a separate validation step function.

I don't know if you realised but you reviewed the second PR (which I built on top of the first - It was the res one I requested a review of!) I'll make the changes in that one so as not to confuse things.

garypolzin (approved):
all good :)

PR #662, opened 2024-12-16
Update reconciliation destinations

garypolzin:
# Description
Fixed filepaths for reconciliation outputs


MarionJHolloway (approved):
Happy to approve, but once this has merged into main, you will need to delete the old dataset from s3 and athena (in that order else it will rebuild!)

PR #666, opened 2025-01-02
Refactor clean ascwds worker

garypolzin:
# Description
Added missing docstring to function and split tests into separate classes

# How to test
No changes to codebase


jobs/clean_ascwds_worker_data.py
```diff
@@ -76,7 +76,19 @@ def main(
     )
 
 
-def remove_workers_without_workplaces(worker_df: DataFrame, workplace_df: DataFrame):
+def remove_workers_without_workplaces(
+    worker_df: DataFrame, workplace_df: DataFrame
+) -> DataFrame:
+    """
+    Removes worker records that do not have a corresponding workplace record.
```
MarionJHolloway, line 83: Might be worth adding why we do this (e.g. because the workplace was removed during the workplace cleaning step).

MarionJHolloway (approved):
Looks good. One small suggestion, but I've approved so you can merge after.

PR #670, opened 2025-01-06
Job role breakdown - old process

garypolzin:
# Description
There are no plans to use this code so I've left it quite poorly written, but it's a working version of the original process which might help with running the newer process. Added a TODO to delete the file once it's up and running properly.

Kept it completely separate to the new job so the new process stays clean and we can iteratively add to it

Created test data and a (hashed out) test for the main job. Although the process will change, the outputs will likely be the same/very similar so thought I'd leave it there as it took a while to create! Again, added a TODO to review at the end whether to keep or not

# TEST FILE GENERATOR HAS FINALLY GONE!!!
The final 2 test files in here are related to this process so it can finally go!

# How to test
No tests written but no plans to include this in the pipeline


jobs/job_role_breakdown.py
```diff
@@ -1,270 +1,246 @@
...
+    cleaned_workers_df = utils.read_from_parquet(
+        cleaned_ascwds_worker_source,
+        selected_columns=cleaned_ascwds_worker_columns_to_import,
+    )
+
+    worker_counts_df = count_workers_by_location(cleaned_workers_df)
+    master_df = merge_dataframes(estimated_posts_df, worker_counts_df)
+
+    job_roles_df = get_unique_job_roles(cleaned_workers_df)
+    master_df = master_df.crossJoin(job_roles_df)
```
MarionJHolloway, line 82: Cross joins are expensive. Might be worth making the job roles into columns in the actual job. (No action here,  just an idea for the future)

tests/test_file_data.py
```diff
@@ -4806,6 +4805,23 @@ class EstimateIndCQCFilledPostsByJobRoleData:
         ("103", date(2025, 1, 1), "9", MainJobRoleLabels.care_worker),
         ("111", date(2025, 1, 1), "10", MainJobRoleLabels.care_worker),
     ]
+    # fmt: off
+    # TODO: Temp test data to check outputs, consider removing or tidying at the end
```
MarionJHolloway, line 4809: Would we benefit from integration tests for this job? (No action again, just an idea)

tests/test_file_generator.py
```diff
@@ -1,53 +0,0 @@
-from utils import utils
```
MarionJHolloway, line 1: This is my favourite code change in a long while!

MarionJHolloway (approved):
Looks good!

PR #677, opened 2025-01-20
Blend asc pir review

garypolzin:
# Description
Changed the % function to use the average of ascwds & pir as the denominator, if you only use one (ascwds) then the rule can pass in one direction but fail the other - see below for example

image

Plus some other minor changes I spotted whilst reviewing code

# How to test
Original tests still passing
[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:blend-asc-pir-review-Ind-CQC-Filled-Post-Estimates-Pipeline:b6c12e53-3871-4d3b-81c5-b3ad0c0c6871)

# Developer checklist
- [X] Unit test


MarionJHolloway (approved):
Looks good, thanks for checking it :)

PR #682, opened 2025-01-30
Extract unique Registered Manager names into a new array column

garypolzin:
# Description
Registered manager names are stored in the contacts array for each regulated activity the location offers.
This function extracts that information, combined first and last name together, combines (and deduplicates) all the registered manager names at each location and joins them back into the dataframe.

# How to test
Branch ran and visual check of Athena outputs match expected behaviour. Tests passing

image

# Developer checklist
- [X] Unit test
- [X] Link to Trello ticket [here](https://trello.com/c/3Zw2gffO/499-epic-3-extract-registered-manager-information-must)


jobs/clean_cqc_location_data.py
```diff
@@ -704,6 +705,122 @@ def raise_error_if_cqc_postcode_was_not_found_in_ons_dataset(
...
 
 
+def extract_registered_manager_names(df: DataFrame) -> DataFrame:
+    """
+    Extracts registered manager names from the regulated activities column.
+
+    The CQC requires a registered manager for each regulated activity at a location.
+    The regulated activities column contains an array of contacts for each activity the location offers.
+    This function extracts the names for all contacts (Registered Managers) and adds them into an array column.
+    Registered manager names are deduplicated in the array so each name will only appears once in the array.
```
MarionJHolloway, line 715: appear (not appears)

jobs/clean_cqc_location_data.py
```diff
@@ -704,6 +705,122 @@ def raise_error_if_cqc_postcode_was_not_found_in_ons_dataset(
...
+
+    Args:
+        df (DataFrame): Input DataFrame with regulated_activities array.
```
MarionJHolloway, line 718: regulated activities is a column?

jobs/clean_cqc_location_data.py
```diff
@@ -704,6 +705,122 @@ def raise_error_if_cqc_postcode_was_not_found_in_ons_dataset(
...
+        df (DataFrame): Input DataFrame with exploded contacts information.
+
+    Returns:
+        DataFrame: DataFrame with selected columns and full_name column.
+    """
+    df = df.select(
+        df[CQCL.location_id],
+        df[CQCLClean.cqc_location_import_date],
+        F.concat_ws(
+            " ",
```
MarionJHolloway, line 779: I'd either abstract this as a variable or include keys with arguments for this function. It wasn't immediately obvious why you had a string with a space here - I had to google the function.

jobs/clean_cqc_location_data.py
```diff
@@ -704,6 +705,122 @@ def raise_error_if_cqc_postcode_was_not_found_in_ons_dataset(
...
+    """
+    Groups the DataFrame by location_id and cqc_location_import_date, and collects all the unique names into an array column.
+
+    Args:
+        df (DataFrame): Filtered DataFrame.
+
+    Returns:
+        DataFrame: Grouped DataFrame with unique registered manager names at each location and time period.
+    """
+    return df.groupBy(CQCL.location_id, CQCLClean.cqc_location_import_date).agg(
```
MarionJHolloway, line 797: Can you separate out the return statement, so it summarises what you're returning, please?

jobs/clean_cqc_location_data.py
```diff
@@ -704,6 +705,122 @@ def raise_error_if_cqc_postcode_was_not_found_in_ons_dataset(
...
+    Joins the DataFrame with the registered manager names column into the original DataFrame.
+
+    Args:
+        df (DataFrame): Original DataFrame.
+        registered_manager_names_df (DataFrame): Grouped DataFrame with registered manager names column.
+
+    Returns:
+        DataFrame: Original DataFrame with registered manager names column joined in.
+    """
+    return df.join(
```
MarionJHolloway, line 817: Can you separate out the return statement, so it summarises what you're returning, please?

jobs/clean_cqc_location_data.py
```diff
@@ -114,6 +114,7 @@ def main(
     registered_locations_df = realign_carehome_column_with_primary_service(
         registered_locations_df
     )
+    registered_locations_df = extract_registered_manager_names(registered_locations_df)
```
MarionJHolloway, line 117: Is it worth moving this into a utils function? There's quite a lot of steps and tests to this and the job is pretty big already.

tests/unit/test_clean_cqc_location_data.py
```diff
@@ -890,5 +890,180 @@ def test_add_column_related_location_returns_correct_values(self):
         )
 
 
+class ExtractRegisteredManagerNamesTests(CleanCQCLocationDatasetTests):
```
MarionJHolloway, line 893: If this was a utils script you'd be able to structure these tests a bit more too and set up a class for each function you're testing

MarionJHolloway (changes requested):
Looking good. I think it would be clearer if this was a utils script though, rather than in the main job. I've left a few smaller suggestions in comments as well.

jobs/clean_cqc_location_data.py
garypolzin: is this ok on the basis that the docstring explains what it does, isn't that it's purpose?
_"... and creates a full name column by concatenating given and family names"_

MarionJHolloway: It still confused me! You could ask for other people' opinion though? Happy to go with the consensus.

MarionJHolloway (approved):
Looking good - might be worth seeing what other people think about that final comment before you merge, but I've approved so I'm not blocking it.

PR #684, opened 2025-01-31
Calculate worker job role counts and convert into map column

RoyPriceCS50:
# Description
Created a script in a subfolder in utils called ascwds_jobrole_count. First function counts job roles per establishmentid and import date, second functions converts main job role and main job role count into a map column.

The main job role and main job role count are removed when map column is added.

Only roles at the establishment are added to the map.

Made a new ticket for adding all potential job roles to map and setting value to 0 when establishment has none.

Trello ticket: https://trello.com/c/3hbzXeq6

# How to test

When main job role is counted, and main job role is null, then count is 0 (unit tested). However, a map column cannot have null keys. So if main job role in ASC-WDS worker cleaned is null then job will fail. 
Can't see any null main job role in cleaned worker data in Athena.

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -78,7 +78,7 @@ def main(
     count_job_roles_per_establishment_df = count_job_roles_per_establishment(
         cleaned_ascwds_worker_df
     )
-    count_job_roles_per_establishment_df = mapped_column(
+    count_job_roles_per_establishment_df = convert_jobrole_count_to_jobrole_map(
         count_job_roles_per_establishment_df
     )
 
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -10,7 +10,7 @@ from utils.column_names.ind_cqc_pipeline_columns import (
 )
 from utils.ind_cqc_filled_posts_utils.ascwds_jobrole_count.ascwds_jobrole_count import (
     count_job_roles_per_establishment,
-    mapped_column,
+    convert_jobrole_count_to_jobrole_map,
 )
 
 PartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -8,7 +8,10 @@ from utils.column_names.ind_cqc_pipeline_columns import (
     IndCqcColumns as IndCQC,
     PartitionKeys as Keys,
 )
-from utils.ind_cqc_filled_posts_utils.ascwds_jobrole_count.ascwds_jobrole_count import count_job_roles_per_establishment, mapped_column
+from utils.ind_cqc_filled_posts_utils.ascwds_jobrole_count.ascwds_jobrole_count import (
+    count_job_roles_per_establishment,
+    mapped_column,
+)
 
 PartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
 cleaned_ascwds_worker_columns_to_import = [
@@ -72,8 +75,12 @@ def main(
         selected_columns=cleaned_ascwds_worker_columns_to_import,
     )
 
-    count_job_roles_per_establishment_df = count_job_roles_per_establishment(cleaned_ascwds_worker_df)
-    count_job_roles_per_establishment_df = mapped_column(count_job_roles_per_establishment_df)
+    count_job_roles_per_establishment_df = count_job_roles_per_establishment(
+        cleaned_ascwds_worker_df
+    )
+    count_job_roles_per_establishment_df = mapped_column(
+        count_job_roles_per_establishment_df
+    )
 
     utils.write_to_parquet(
         estimated_ind_cqc_filled_posts_df,
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -73,14 +73,14 @@ def main(
 
     # TODO 1. Make a window over establishmentid, import date and job role.
 
-    # TODO 2. Use the window to add a column called "temp_job_role_count". 
+    # TODO 2. Use the window to add a column called "temp_job_role_count".
     #         The values are a count of rows in window.
     #
     # TODO 3. Make another window over locationid and import date.    #
     #
     # TODO 3. Use create_map function to create a new column of type dictionary. Apply that function over the new window.
     #         create_map function arguments must be column or string.
-    #         For example: 
+    #         For example:
     #         df = df.withColumn(
     #           "map_column_name",
     #           F.create_map(job role column, temp_job_role_count column).alias("ascwds_main_job_role_counts").over(window)
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -71,6 +71,20 @@ def main(
         selected_columns=cleaned_ascwds_worker_columns_to_import,
     )
 
+    # TODO 1. Make a window over establishmentid, import date and job role.
+
+    # TODO 2. Use the window to add a column called "temp_job_role_count". 
+    #         The values are a count of rows in window.
+    #
+    # TODO 3. Make another window over locationid and import date.    #
+    #
+    # TODO 3. Use create_map function to create a new column of type dictionary. Apply that function over the new window.
+    #         create_map function arguments must be column or string.
+    #         For example: 
+    #         df = df.withColumn(
+    #           "map_column_name",
+    #           F.create_map(job role column, temp_job_role_count column).alias("ascwds_main_job_role_counts").over(window)
+
     utils.write_to_parquet(
         estimated_ind_cqc_filled_posts_df,
         estimated_ind_cqc_filled_posts_by_job_role_destination,
--- a/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
+++ b/jobs/estimate_ind_cqc_filled_posts_by_job_role.py
@@ -8,6 +8,7 @@ from utils.column_names.ind_cqc_pipeline_columns import (
     IndCqcColumns as IndCQC,
     PartitionKeys as Keys,
 )
+import utils.ind_cqc_filled_posts_utils.ascwds_jobrole_count.ascwds_jobrole_count as job
 
 PartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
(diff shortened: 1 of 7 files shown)


jobs/estimate_ind_cqc_filled_posts_by_job_role.py
```diff
@@ -8,6 +8,10 @@
     IndCqcColumns as IndCQC,
     PartitionKeys as Keys,
 )
+from utils.ind_cqc_filled_posts_utils.ascwds_jobrole_count.ascwds_jobrole_count import (
+    count_job_roles_per_establishment,
+    convert_jobrole_count_to_jobrole_map,
```
MarionJHolloway, line 13: Can we standardise jobrole(s) to job_role(s) please?

tests/test_file_data.py
```diff
@@ -8991,3 +8991,93 @@ class BlendAscwdsPirData:
         ("loc 1", date(2023, 1, 1), date(2020, 1, 1), 10, 20.0),
     ]
     expected_drop_temporary_columns = [IndCQC.location_id]
+
+
+@dataclass
+class AscwdsJobroleCountData:
+    # fmt: off
+    workplace_with_two_workers_rows = [
```
MarionJHolloway, line 8999: I think you're missing a test case where there are two establishments with different job roles, e.g. 
("1", date(2025, 1, 1), "Care worker"),
("2", date(2025, 1, 1), "Senior Care worker"),

tests/test_file_data.py
```diff
@@ -8991,3 +8991,93 @@ class BlendAscwdsPirData:
...
+        ("1", date(2025, 1, 1), "Care worker"),
```
MarionJHolloway, line 9000: The names of job roles are abstracted here: utils/column_values/categorical_column_values.py around line 140. You can use these in the test data as well as the main code.

tests/test_file_data.py
```diff
@@ -8991,3 +8991,93 @@ class BlendAscwdsPirData:
...
+
+    workplace_with_one_worker_in_two_roles_counted_rows = [
+        ("1", date(2025, 1, 1), "Care worker", 1),
+        ("1", date(2025, 1, 1), "Senior care worker", 1),
+    ]
+    expected_workplace_with_one_worker_in_two_roles_counted_rows = [
+        ("1", date(2025, 1, 1), {"Care worker": 1, "Senior care worker": 1}),
+    ]
+ 
+    two_workplaces_with_same_jobrole_counted_rows = [
```
MarionJHolloway, line 9066: Again, I think you're missing a test case where there are different roles at different establishments:
("1", date(2025, 1, 1), "Care worker", 1),
("2", date(2025, 1, 1), "Senior Care worker", 1),

tests/test_file_data.py
```diff
@@ -8991,3 +8991,93 @@ class BlendAscwdsPirData:
...
+ 
+    workplace_across_different_importdates_same_jobrole_counted_rows = [
+        ("1", date(2025, 1, 1), "Care worker", 1),
+        ("1", date(2025, 2, 1), "Care worker", 1),
+    ]
+    expected_workplace_across_different_importdates_same_jobrole_counted_rows = [
+        ("1", date(2025, 1, 1), {"Care worker": 1}),
+        ("1", date(2025, 2, 1), {"Care worker": 1}),
+    ]
+    # fmt: on
```
MarionJHolloway, line 9083: The data in these doesn't look like it's worth turning the linter off?

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -0,0 +1,238 @@
+import unittest
```
MarionJHolloway, line 1: Can you add checks to the tests in the glue job that these functions are being called as expected as well please? We normally mock them out (see tests/unit/test_estimate_ind_cqc_filled_posts.py for an example of this)

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -0,0 +1,238 @@
...
+        returned_df = count_job_roles_per_establishment(
+            test_two_workplaces_with_same_jobrole_df
+        )
+
+        self.assertEqual(
+            returned_df.sort(AWKClean.establishment_id).collect(),
+            expected_two_workplaces_with_same_jobrole_df.collect(),
+        )
+
+    def test_count_job_roles_per_establishment_counts_when_importdate_is_different(
```
MarionJHolloway, line 87: Can we standardise importdate(s) to import_date(s) please?

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -0,0 +1,238 @@
...
+            test_workplace_with_two_workers_counted_df
+        )
+
+        self.assertEqual(
+            returned_df.sort(AWKClean.establishment_id).collect(),
+            expected_workplace_with_two_workers_counted_df.collect(),
+        )
+
+    def test_convert_jobrole_count_to_jobrole_map_converts_two_rows_into_one_dict(self):
+        test_workplace_with_one_worker_in_two_roles_counted_df = (
```
MarionJHolloway, line 172: The test name and the variable names here don't seem to correlate. Is this the right name/ test data for what you want to check?

utils/column_names/cleaned_data_files/ascwds_worker_cleaned.py
```diff
@@ -10,3 +10,4 @@ class AscwdsWorkerCleanedColumns(AscwdsWorkerColumns):
     ascwds_worker_import_date: str = "ascwds_worker_import_date"
     main_job_role_clean: str = AscwdsWorkerColumns.main_job_role_id + "_clean"
     main_job_role_clean_labelled: str = main_job_role_clean + "_labels"
+    ascwds_main_job_role_counts: str = "ascwds_main_job_role_counts"
```
MarionJHolloway, line 13: This wants to be in utils/column_names/ind_cqc_pipeline_columns.py instead. This column isn't present in the cleaned worker dataset, it's only needed in the ind cqc pipeline, so really belongs there.

utils/ind_cqc_filled_posts_utils/ascwds_jobrole_count/template_jobrole_count_dict.py
```diff
@@ -0,0 +1,51 @@
+from utils.column_values.categorical_column_values import MainJobRoleLabels
+
+# Roy - Note - We have a dict in which the keys are job roles and values are counts per establishemnt.
```
MarionJHolloway, line 3: The extra test I've suggested above should demonstrate whether this is a problem or not. If it is a problem, it might be better in another ticket (assuming the job completes in branch), so you could write the test and skip it for now if it's failing, as it seems to go beyond the scope of the original ticket.

tests/test_file_data.py
```diff
@@ -8991,3 +8991,93 @@ class BlendAscwdsPirData:
...
+        ("1", date(2025, 1, 1), "Care worker"),
+        ("1", date(2025, 2, 1), "Care worker"),
+    ]
+
+    expected_workplace_across_different_importdates_same_jobrole_rows = [
+        ("1", date(2025, 1, 1), "Care worker", 1),
+        ("1", date(2025, 2, 1), "Care worker", 1),
+    ]
+
+    workplace_with_null_jobrole_rows = [
```
MarionJHolloway, line 9038: This shouldn't happen in the data (but obviously worth keeping the test to demonstrate behaviour). Can you check if there is a rule in the cleaned worker dataset validation to check that main job role has no null values and if not, add it in please? This validation check is run long before this job, so should flag any problems nice and early in the pipeline.

MarionJHolloway (changes requested):
I've added some feedback on this PR. As well, before it can be approved to merge into main, you'll need to demonstrate that the Ind CQC pipeline can run successfully in this branch. This is how we check there's no regression from these changes. If you put a link to the pipeline run in branch in the PR description, it's easy to check :)

utils/ind_cqc_filled_posts_utils/ascwds_jobrole_count/template_jobrole_count_dict.py
garypolzin: this file/code doesn't look like it's being used as part of this PR so I'd completely remove it for now - ideally we should only add code which is being used. There's a chance we forget this exists by the time we get round to the other ticket, or we take a different approach (or a different person picks it up) and don't think to delete this

tests/test_file_data.py
```diff
@@ -8991,3 +8991,105 @@ class BlendAscwdsPirData:
...
+        ("1", date(2025, 1, 1), None),
+    ]
+
+    expected_workplace_with_null_jobrole_rows = [
+        ("1", date(2025, 1, 1), None, 0),
+    ]
+
+
+
+
```
garypolzin, line 9048: quite a lot of blanks rows here, thought the linter would kick off about these!

tests/test_file_data.py
```diff
@@ -8991,3 +8991,105 @@ class BlendAscwdsPirData:
...
+    workplace_with_two_workers_counted_rows = [
+        ("1", date(2025, 1, 1), "Care worker", 2),
+    ]
+ 
+    expected_workplace_with_two_workers_counted_rows = [
+        ("1", date(2025, 1, 1), {"Care worker": 2}),
+    ]
```
garypolzin, line 9056: typically we don't leave a blank row in between the test data and it's expected test data, it helps group them together

then one space in between if they are unrelated data

garypolzin (changes requested):
sorry added a few more bits, more style over substance though!

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -0,0 +1,279 @@
...
+
+        returned_df = count_job_role_per_establishment(
+            test_workplace_with_two_workers_df
+        )
+
+        self.assertEqual(
+            returned_df.columns, expected_workplace_with_two_workers_df.columns
+        )
+
+    def test_convert_job_role_count_to_job_role_map_converts_one_row_into_one_dict(
```
garypolzin, line 166: can you add another class here as it's testing a different function to the code above 

```
class ConvertJobRoleCountToJobRoleMapTests(AscwdsJobroleCount):
    def setUp(self) -> None:
        super().setUp()
```

I think you'll also need to add
```
    def setUp(self) -> None:
        super().setUp()
```
to row 22

garypolzin (commented):
Only had a quick look through but looking loads better!

Added a comment regarding test structure, makes it easier to see which tests are for each function

tests/unit/test_ascwds_jobrole_count.py
RoyPriceCS50: That's using the correct test data. The test before this one is checking the most basic functionality, then this test is checking if two job roles get placed into one dictionary.

As opposed to two job roles ending up in a list of dicts or two rows with a 1 job role dict in each row

MarionJHolloway (approved):

garypolzin (approved):
all good now thanks!

PR #686, opened 2025-02-04
Calculate worker job role counts and convert into map column

tmehmed:
# Description

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


jobs/estimate_ind_cqc_filled_posts_by_job_role.py
```diff
@@ -8,7 +8,7 @@
     IndCqcColumns as IndCQC,
     PartitionKeys as Keys,
 )
-from utils.ind_cqc_filled_posts_utils.ascwds_job_role_count.ascwds_job_role_count import (
+from utils.estimate_filled_posts_by_job_utils.ascwds_mapped_column_job_role_count.ascwds_job_role_count import (
```
MarionJHolloway, line 11: "job_role" rather than "job". We previously used job to mean something else so this might cause confusion.

RoyPriceCS50 (commented):
I think this doesn't need to be a PR.

This was me making a proof of concept. The whole branch can be deleted if the PR for the pivot method is accepted.

PR #692, opened 2025-02-06
Create columns for job role and their counts per establishment

tmehmed:
# Description
Added the function that groups the data by establishment and import date and pivots the result on job role counting the total number of job roles per establishment.

Trello: https://trello.com/c/KWdlvfml

Successful Step Function Run Link: https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:create-columns-for-job-role-an-Ind-CQC-Filled-Post-Estimates-Pipeline:e46e11ac-bf93-4553-af65-1721505a4ba1

Athena Output Link: https://eu-west-2.console.aws.amazon.com/athena/home?region=eu-west-2#/query-editor/history/5875f335-988e-4c49-9039-bb11056f6e17

# How to test
I've added three tests, one for establishments which have a job for every job role, a test for an establishment which only has 1 job and also a test which has 2 jobs for the same job role.

# Developer checklist
- Unit test
- Linked to Trello ticket
- Documentation up to date


tests/test_file_data.py
```diff
@@ -9342,97 +9342,69 @@ class BlendAscwdsPirData:
...
+
+    workplace_three_jobs_roles_with_two_being_distinct = [
+    ("1", date(2025, 1, 1), MainJobRoleLabels.not_known),
+    ("1", date(2025, 1, 1), MainJobRoleLabels.middle_management),
+    ("1", date(2025, 1, 1), MainJobRoleLabels.middle_management),
+    ]
+
+    exptected_workplace_three_job_roles_with_two_being_distinct = [
+        ("1", date(2025, 1, 1), 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
     ]
```
RoyPriceCS50, line 9408: Please add tests for different establishmentid, different import dates and None job role. To demonstrate it counts them independently, and None should get assigned 0 for all job roles.

utils/estimate_filled_posts_by_job_role_utils/ascwds_job_role_count/ascwds_job_role_count.py
```diff
@@ -0,0 +1,77 @@
...
+    df = (
+        df.groupBy(
+            F.col(AWKClean.establishment_id),
+            F.col(AWKClean.ascwds_worker_import_date),
+        )
+        .pivot(AWKClean.main_job_role_clean_labelled, list_of_job_roles)
+        .count()
+    )
+
+    df = df.na.fill(0)
```
RoyPriceCS50, line 76: I think this replacing nulls would be safer to loop through a list of the job role columns rather than the whole thing.

I doubt establishmentid and import date could be null but better safe than sorry,

RoyPriceCS50 (changes requested):
Also need to run the ind-cqc-filled-post-estimate step function in your branch.
Put a link to the successful run in the PR description.

Add a temporary write to parquet that writes the df with counts into a folder in your branch. Run this job again (having run the pipeline earlier, you dont need to run the whole thing again). Then run the ind_cqc_filled_posts crawler. Then you can preview the output in Athena and put a link to it in the PR description as well.

utils/estimate_filled_posts_by_job_role_utils/ascwds_job_role_count/ascwds_job_role_count.py
garypolzin: you can add a list of columns you want to change using subset to get around this:
`df.na.fill(0, subset=list_of_job_roles)`

tests/test_file_data.py
```diff
@@ -9342,97 +9342,69 @@ class BlendAscwdsPirData:
...
 
-    workplace_with_two_different_roles_rows = [
-        ("1", date(2025, 1, 1), MainJobRoleLabels.care_worker),
+    workplace_with_one_of_every_possible_job_role = [
+        ("1", date(2025, 1, 1), MainJobRoleLabels.not_known),
+        ("1", date(2025, 1, 1), MainJobRoleLabels.senior_management),
+        ("1", date(2025, 1, 1), MainJobRoleLabels.middle_management),
+        ("1", date(2025, 1, 1), MainJobRoleLabels.first_line_manager),
+        ("1", date(2025, 1, 1), MainJobRoleLabels.registered_manager),
+        ("1", date(2025, 1, 1), MainJobRoleLabels.supervisor),
```
garypolzin, line 9352: As mentioned before, the list of job roles changes reasonably frequently so 'every_possible' will soon be out of date

we go by theory if it works for 3 job roles then it will also work for 35, so we try to keep the test files as small as possible to test the functionality of the code - plus it makes the test files much smaller and easier to make!

garypolzin (changes requested):
Looking good in general thanks, couple of comments mostly referring to the longevity of the function which you wouldn't have known about

tests/test_file_data.py
RoyPriceCS50, line 1: There are two class's for "AscwdsJobroleCountData".

Looks like none of the cases from row 9446 to 9541 are being used anymore so can go.

tests/test_file_schemas.py
RoyPriceCS50, line 1: Same as with the test data, there are two schemas with the same name.
The one on row 5901 can be removed.

utils/ind_cqc_filled_posts_utils/ascwds_job_role_count/ascwds_job_role_count.py
```diff
@@ -5,62 +5,37 @@
...
-            F.col(IndCQC.ascwds_main_job_role_counts),
-        ),
-    )
-    df_mapped = df_struct.groupBy(
-        F.col(AWKClean.establishment_id), F.col(AWKClean.ascwds_worker_import_date)
-    ).agg(
-        F.map_from_entries(F.collect_list(struct_column)).alias(
-            IndCQC.ascwds_main_job_role_counts
+
+    list_of_job_roles
```
RoyPriceCS50, line 29: I think this can be removed.

garypolzin (changes requested):
Sorry this task has turned out to be a lot more faffy than expected!

I'm still just a bit conscious that the actual job role list changes reasonably frequently and the tests we currently have will only work if these current 39 job roles exist. Hopefully the changes I've suggested make sense and test the functionality you've written without relying on the actual job roles

On the whole it's looking great, and again we probably should have thought about some of these issues in advance of the ticket!

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -22,264 +23,108 @@ class CountJobRolesPerEstablishmentTests(AscwdsJobroleCount):
...
-        test_workplace_with_two_workers_df = self.spark.createDataFrame(
-            Data.workplace_with_two_workers_rows,
-            Schemas.ascwds_worker_schema,
-        )
-        expected_workplace_with_two_workers_df = self.spark.createDataFrame(
-            Data.expected_workplace_with_two_workers_rows,
-            Schemas.ascwds_worker_with_job_role_count_schema,
-        )
+    def test_workplace_job_role_count(self):
+        self.assertEqual(len(list_of_job_roles), 39)
```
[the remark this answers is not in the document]
RoyPriceCS50: I asked for this test, thinking about how changes might be made to the service without our knowledge.
I didn't know the validation checks covered this off.
Agree with Gary's comment.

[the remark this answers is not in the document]
garypolzin: no worries, there's a lot of code now!

tests/test_file_data.py
```diff
@@ -9587,3 +9488,58 @@ class RegisteredManagerNamesCountData:
...
+        ("2", date(2025, 1, 1), MainJobRoleLabels.employment_support),
+    ]
+
+    expected_workplace_with_different_establishmentid = [
+        ("1", date(2025, 1, 1), 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
+        ("2", date(2025, 1, 1), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
+    ]
+    workplace_with_none_job_role = [
+        ("1", date(2025, 1, 1), MainJobRoleLabels.not_known),
+        ("1", date(2025, 1, 1), None),
```
RoyPriceCS50, line 9528: The not known job role can be removed from this test. Just have the None row.

RoyPriceCS50 (approved):

tests/unit/test_ascwds_jobrole_count.py
```diff
@@ -22,264 +23,121 @@ class CountJobRolesPerEstablishmentTests(AscwdsJobroleCount):
...
-
         self.assertEqual(
-            returned_df.sort(AWKClean.establishment_id).collect(),
-            expected_workplace_with_two_workers_df.collect(),
+            len(returned_df.columns),
+            len(test_workplace_with_three_distinct_job_role_df.columns) + 3,
         )
 
-    def test_count_job_role_per_establishment_counts_when_job_roles_are_different(
+    def test_workplace_with_three_distinct_job_role(
```
MarionJHolloway, line 41: We generally name our tests using the following format: test _[function]_returns_[expected_behaviour]_when_[test_case]

For example this test would be called test_count_job_role_per_establishment_as_columns_returns_correct_counts_when_workplace_has_three_distinct_job_roles

MarionJHolloway (commented):
One small tweak to test names, otherwise looks good to me :)

garypolzin (approved):
all my comments have been addressed so happy to approve, tough ticket but good work!

PR #693, opened 2025-02-10
Amend ASCWDS ingest step to also ingest NMDSSC style CSVs

garypolzin:
# Description
By default this runs on ASCWDS only so it behaves in the same way it did previously.
However, I've added the option to select NMDSSC files which includes a date amendment not relevant to ASCWDS files, and excludes mainjrid not being -1 as that was fine to happen in NMDS files

# How to test
Ran all data in branch and behaves as expected.
Added unit tests for new functions

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/tAfQrIke/704-epic-10-add-an-nmds-option-into-the-ascwds-ingestion-job-must)

### DIFF
--- a/jobs/ingest_ascwds_dataset.py
+++ b/jobs/ingest_ascwds_dataset.py
@@ -9,22 +9,24 @@ from utils.column_names.raw_data_files.ascwds_worker_columns import (
 )
 
 
-def main(source: str, destination: str):
+def main(source: str, destination: str, dataset: str = "ascwds"):
     bucket, prefix = utils.split_s3_uri(source)
 
     if utils.is_csv(source):
-        ingest_single_file(source, bucket, prefix, destination)
+        ingest_single_file(source, bucket, prefix, destination, dataset)
     else:
-        ingest_multiple_files(bucket, prefix, destination)
+        ingest_multiple_files(bucket, prefix, destination, dataset)
 
 
-def ingest_single_file(source: str, bucket: str, prefix: str, destination: str):
+def ingest_single_file(
+    source: str, bucket: str, prefix: str, destination: str, dataset: str
+):
     print("Single file provided to job. Handling single file.")
     new_destination = utils.construct_destination_path(destination, prefix)
-    handle_job(source, bucket, prefix, new_destination)
+    handle_job(source, bucket, prefix, new_destination, dataset)
 
 
-def ingest_multiple_files(bucket: str, prefix: str, destination: str):
+def ingest_multiple_files(bucket: str, prefix: str, destination: str, dataset: str):
     print("Multiple files provided to job. Handling each file...")
     objects_list = utils.get_s3_objects_list(bucket, prefix)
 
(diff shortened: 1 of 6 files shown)


jobs/ingest_ascwds_dataset.py
```diff
@@ -58,26 +63,39 @@ def ingest_multiple_files(bucket: str, prefix: str, destination: str):
...
-def handle_job(source: str, source_bucket: str, source_key: str, destination: str):
+def handle_job(
+    source: str, source_bucket: str, source_key: str, destination: str, dataset: str
+):
     """
     Handle the job of reading CSV file(s), processing then writing the data to parquet format.
 
-    An error is raised if any new ASCWDS files contain unknown main job role IDs as this should no longer occur.
+    ASCWDS and NMDSSC files are very similar in structure but have some differences in the data they contain.
+    - An error is raised if any new ASCWDS files contain unknown main job role IDs as this should no longer occur.
```
MarionJHolloway, line 76: Add this to the docstring raises section as well

jobs/ingest_ascwds_dataset.py
```diff
@@ -111,6 +129,27 @@ def raise_error_if_mainjrid_includes_unknown_values(df: DataFrame) -> DataFrame:
     return df
 
 
+def fix_nmdssc_dates(df: DataFrame) -> DataFrame:
```
MarionJHolloway, line 132: This will change the data type to date as well? I think this step was previously done in the cleaning job. Should this be part of data cleaning rather than ingestion?

tests/unit/test_ingest_ascwds_dataset.py
```diff
@@ -142,6 +153,71 @@ def test_handle_job(
...
+            self.bucket,
+            self.csv_file_name,
+            self.destination_path,
+            dataset,
+        )
+
+        read_partial_csv_content_mock.assert_called_once_with(
+            self.bucket, self.csv_file_name
+        )
+        identify_csv_delimiter_mock.assert_called_once_with(file_sample)
```
MarionJHolloway, line 190: You can also assert_not_called() for any functions you want to check are ignored. Might be worth doing in the asc and nmds tests

jobs/ingest_ascwds_dataset.py
garypolzin: raw ascwds dates are in the format dd/MM/yyyy
raw nmdssc dates are in the format MM/dd/yyyy

so this function converts the original string format to a date using `F.to_date` then converts that date back to a string again using `F.date_format`

so the end result is a string format which matches the string format of ascwds data

the cleaning step then converts that entire string format to a date

garypolzin: NMDS and ASC data are combined together, so the date format needs to be consistent

jobs/ingest_ascwds_dataset.py
garypolzin: just this one or every docstring which is affected?

this function calls another function, it's the other function which raises the error

MarionJHolloway (approved):
Looks good, thanks!

PR #694, opened 2025-02-11
Account for lastloggedin date column in ingest job

garypolzin:
# Description
lastloggedin doesn't have 'date' in the name so it isn't automatically captured. It's in the workplace file but not the worker file so set up as an if statement instead of appending to the list of date cols

# How to test
Added test function

# Developer checklist
- [X] Unit test

### DIFF
--- a/jobs/ingest_ascwds_dataset.py
+++ b/jobs/ingest_ascwds_dataset.py
@@ -7,6 +7,9 @@ from utils import utils
 from utils.column_names.raw_data_files.ascwds_worker_columns import (
     AscwdsWorkerColumns as AWK,
 )
+from utils.column_names.raw_data_files.ascwds_workplace_columns import (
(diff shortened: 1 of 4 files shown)


jobs/ingest_ascwds_dataset.py
```diff
@@ -147,6 +152,15 @@ def fix_nmdssc_dates(df: DataFrame) -> DataFrame:
             date_column,
             F.date_format(F.to_date(df[date_column], "MM/dd/yyyy"), "dd/MM/yyyy"),
         )
+
+    # last_logged_in is the only date column which doen't contain the word date so we need to handle it separately
```
MarionJHolloway, line 156: I would move this comment into the main doc string for this function

MarionJHolloway (approved):
Looks good - one small tweak

PR #725, opened 2025-03-10
Impute CQC specialisms column

garypolzin:
# Description
Copied the process for imputing gac_service_types and regulated_activities to also impute specialisms (to populate points in time where specialisms randomly appears to be missing in that period of time, it uses their known value from another period of time, or keeps as null if never known).

Some amendments

- Renamed the function (removed '_to_data' at the end, not sure where else it would be adding the new column!)
- Some specialisms remain as null (this isn't the case for the other two columns) so I've added a step to the function which sets zero as the minimum (by default F.size returns a -1 count if the value is null)
- Replaced specialisms downstream in the pipeline with imputed_specialisms (also replaced one instance of gac_service_types which should be imputed_gac_service_types)
- beefed up the tests for this function, previous test was very basic

# How to test
Aded additonal tests which are passing

[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:impute-specialisms-Ind-CQC-Filled-Post-Estimates-Pipeline:f7f6e46d-708c-469e-aa4e-13523c44eeb3)

Athena outputs match expected behaviour (less zero counts as empty cells replaced with known values)
image

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/9usyBYAH/735-epic-5-impute-specialisms-must)


RoyPriceCS50 (approved):
I've approved but since you're gap filling the specialism column, do you want to import that into the CQC coverage pipeline as well, since the other imputed things like activities and service types go into it.

PR #740, opened 2025-03-28
Added additional bucket so 1 bed gives a banding of 1 instead of zero

garypolzin:
# Description
Turns out the models treat zero as unknown, and the bucketizer returns zero for the first band.

I've added an additional category starting at 1 into the bucketizer, so any bed value of 1 will now return a banded bed category of 1 instead of zero, which subsequently adds one onto all previous bands

# How to test
The number of the band doesn't matter, every band will now just be one larger than before - updated previous tests to account for this change

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/QspzTKPV/756-epic-5-change-min-banded-bed-value-to-be-1-must)

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -4047,15 +4047,15 @@ class CleaningUtilsData:
     ]
 
     create_banded_bed_count_column_rows = [
-        ("1-001", CareHome.care_home, 5),
+        ("1-001", CareHome.care_home, 1),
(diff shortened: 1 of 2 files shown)


tmehmed (approved):
There is not much to review in this PR. As the tests pass im also happy for the PR to be merged!

PR #742, opened 2025-03-31
Combine Models - 1 - Set up basic framework

garypolzin:
# Description
This PR is to set up the outline for the process of merging both non res models together.
The 'dormancy' field first became available in 2022, so we have a 'without dormancy' model which can always be run and an improved 'with dormancy' model which is run from 2022 onwards.

In order to ensure a smooth transition from one model to the other we are setting up this function to do that

# How to test
Basic framework so no tests at this point in time

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/g2Ron9S7/759-epic-5-combine-models-1-set-up-basic-framework-must)

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -6338,6 +6338,37 @@ class ModelNonResWithoutDormancy:
 
 
 @dataclass
+class ModelNonResWithAndWithoutDormancyCombinedRows:
+    estimated_posts_rows = [
+        ("1-001", date(2021, 1, 1), CareHome.not_care_home, "Y", 1, 1.0, None),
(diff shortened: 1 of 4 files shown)


tmehmed (approved):
I'm happy to approve this PR, with the following caveats:

- After merging, the **TODO**'s still have to be complete
- Adding in **expected_estimated_posts_rows** for when you are ready to carry out tests. (There is a TODO comment missing within test_file_data which you could add)
- Adding in **expected_estimated_posts_schema** for when you are ready to carry out tests( There is TODO comment missing within test_file_schemas which you could add)

PR #746, opened 2025-04-02
Remove locations without a provider ID

garypolzin:
# Description
Currently our pipeline is failing because a location is present which doesn’t have a provider id - it’s missing most other data too

if that location appears again with a known providerid then it will be populated and they will stay in our dataset, but for now, put in a step which deletes locations who have never had one

Update validations (expected size) to match

# How to test
Main branch validations are [currently failing](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/main-validate_locations_api_cleaned_data_job/runs)
They now [pass in branch when run against main data](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/remove-locs-without-provid-validate_locations_api_cleaned_data_job/runs)

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/k5Py4y0i/767-remove-locations-without-a-provider-id-must)

### DIFF
--- a/jobs/validate_locations_api_cleaned_data.py
+++ b/jobs/validate_locations_api_cleaned_data.py
@@ -34,6 +34,7 @@ PartitionKeys = [Keys.year, Keys.month, Keys.day, Keys.import_date]
 raw_cqc_locations_columns_to_import = [
     Keys.import_date,
     CQCL.location_id,
+    CQCL.provider_id,
     CQCL.type,
     CQCL.registration_status,
     CQCL.gac_service_types,
@@ -55,9 +56,9 @@ def main(
     )
     rules = Rules.rules_to_check
 
-    rules[
-        RuleName.size_of_dataset
-    ] = calculate_expected_size_of_cleaned_cqc_locations_dataset(raw_location_df)
+    rules[RuleName.size_of_dataset] = (
+        calculate_expected_size_of_cleaned_cqc_locations_dataset(raw_location_df)
+    )
 
     cleaned_cqc_locations_df = add_column_with_length_of_string(
         cleaned_cqc_locations_df, [CQCL.location_id, CQCL.provider_id]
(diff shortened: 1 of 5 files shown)


jobs/validate_locations_api_cleaned_data.py
```diff
@@ -99,38 +105,53 @@ def calculate_expected_size_of_cleaned_cqc_locations_dataset(
...
+            new_col_name,
+            F.when(
+                F.max(F.when(df[col_to_check].isNotNull(), 1).otherwise(0)).over(
+                    window_spec
+                )
+                > 0,
+                True,
+            ).otherwise(False),
+        )
     return df
```
tmehmed, line 156: Could be good to have a single df = df.withColumn(...) outside the if-else statements, although this isn't a necessity.

in doing so the overall function would reduce by a few lines, and readability could slightly improve.

tmehmed (approved):
Im happy for this to be merged! 

I've made a small suggestion regarding the function identify_if_location_has_a_known_value. 

Also at the moment, the only test for a non-array type is when the column is a string. i.e provider_id. You could introduce tests where the column is integers, booleans etc. But I'm fairly certain the tests will pass either way as the function will still be looking for nulls regardless of the data type. So i'm happy for the tests to be kept as it is.

jobs/validate_locations_api_cleaned_data.py
garypolzin: @tmehmed I'm struggling to find a way to make this into one df.withColumn(...) function

moving the isinstance check doesn't seem to work inside F.when(...) as it's expected a pyspark column expression, do you have any suggestions?

In the meantime I've edited the current code which hopefully makes it more readable

PR #749, opened 2025-04-02
Convert ratios into posts and merge with posts data

garypolzin:
# Description
Not used in the pipeline yet but created function in preparation

# How to test
Tests passing

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/Uu9XNiWe/771-epic-5-convert-ratios-into-posts-and-merge-with-posts-data-must)

### DIFF
--- a/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
+++ b/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
@@ -123,3 +123,38 @@ class SetMinimumValueTests(EstimateFilledPostsModelsUtilsTests):
         expected_df = test_df
 
         self.assertEqual(returned_df.collect(), expected_df.collect())
+
+
+class ConvertCareHomeRatiosToFilledPostsAndMergeWithFilledPostValuesTests(
+    EstimateFilledPostsModelsUtilsTests
+):
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.test_df = self.spark.createDataFrame(
+            Data.convert_care_home_ratios_to_filled_posts_and_merge_with_filled_post_values_rows,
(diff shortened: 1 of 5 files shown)


tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
```diff
@@ -123,3 +123,38 @@ def test_set_min_value_does_not_replace_null_predictions(self):
...
+    def test_returned_column_values_match_expected_when_not_care_home(self):
+        returned_data = self.returned_df.sort(IndCqc.location_id).collect()
+        expected_data = self.expected_df.collect()
+
+        for i in range(len(returned_data)):
+            self.assertEqual(
+                returned_data[i],
+                expected_data[i],
+                f"Returned row {i} does not match expected",
+            )
```
tmehmed, line 160: Here we are iterating through returned_data, which is fine for this test case.

However another test to check that the number of records doesn't change compared to the original dataframe could also be beneficial.

The reason for this is, if our expected dataframe has more records than the returned dataframe the test **test_returned_column_values_match_expected_when_not_care_home** won't pick it up as an error. I understand that this isn't what the test is for. But a seperate test to account for possible record changes could be beneficial.

tmehmed (approved):
Im happy for this branch to be merged. 

The test data covers every possible scenario i can think of. The names for functions and variables look appropriate and the tests also cover the columns matching that of the original dataframe.

There's one comment i left but apart from that it can be approved.

tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
garypolzin: I think we'll be fine without that test but good spot.
None of 'withColumn', 'when' or 'otherwise' will affect row count, and if they did we'd have a far greater problem!

PR #752, opened 2025-04-03
Combine models 3

garypolzin:
# Description

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -6367,6 +6367,22 @@ class ModelNonResWithAndWithoutDormancyCombinedRows:
         ("1-006", date(2024, 8, 1), CareHome.not_care_home, "N", 4, 3.0, 3.0),
     ]
 
+    average_models_by_related_location_and_time_registered_rows = [
+        ("1-001", RelatedLocation.no_related_location, 1, 5.0, 14.0),
+        ("1-002", RelatedLocation.no_related_location, 1, 6.0, 15.0),
+        ("1-003", RelatedLocation.has_related_location, 1, 1.0, 10.0),
+        ("1-004", RelatedLocation.has_related_location, 1, 2.0, 11.0),
+        ("1-005", RelatedLocation.has_related_location, 2, 3.0, 12.0),
+        ("1-006", RelatedLocation.has_related_location, 2, 4.0, 13.0),
+        ("1-007", RelatedLocation.has_related_location, 2, 20.0, None),
+        ("1-008", RelatedLocation.has_related_location, 2, None, 20.0),
+    ]
+    expected_average_models_by_related_location_and_time_registered_rows = [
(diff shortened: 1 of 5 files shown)


utils/estimate_filled_posts/models/non_res_with_and_without_dormancy_combined.py
```diff
@@ -133,9 +134,101 @@ def apply_model_ratios(df: DataFrame) -> DataFrame:
...
+        TempColumns.residual_at_overlap,
+        F.col(IndCqc.non_res_with_dormancy_model)
+        - F.col(TempColumns.non_res_without_dormancy_model_adjusted),
+    )
+    residual_df = residual_df.select(
+        IndCqc.location_id, TempColumns.residual_at_overlap
+    )
+
+    return residual_df
+
```
tmehmed, line 209: I think the doc-string is very good, but it would be benefitial to mention what the first overlap date is and why we are filtering out records where the location_import_date is not equal to it.

Aoart from that the function looks great.

tests/test_file_schemas.py
```diff
@@ -3376,7 +3378,75 @@ class ModelNonResWithAndWithoutDormancyCombinedSchemas:
...
+        ]
+    )
+    expected_apply_residuals_schema = StructType(
+        [
+            *apply_residuals_schema,
+            StructField(
+                NRModel_TempCol.non_res_without_dormancy_model_adjusted_and_residual_applied,
+                FloatType(),
+                True,
             ),
```
tmehmed, line 3450: Should location_id have nullable_bool as True?
same with IndCqc_location_import_date?

tmehmed (approved):
Hello Gary,

The PR generally looks really good, i've left a few comments. Happy to discuss with you if anything doesn't make sense or if you disagree.

tests/test_file_schemas.py
garypolzin: good spot, we are pretty slack with that by the looks of it, over 300 cases of those being wrong so I'll create a tech debt ticket to change the others

tests/test_file_data.py
```diff
@@ -6414,6 +6414,66 @@ class ModelNonResWithAndWithoutDormancyCombinedRows:
...
+        ("1-001", 10.0, None),
+        ("1-002", None, -1.0),
+        ("1-003", None, None),
+    ]
+    expected_apply_residuals_when_null_values_present_rows = [
+        ("1-001", 10.0, None, 10.0),
+        ("1-002", None, -1.0, None),
+        ("1-003", None, None, None),
+    ]
+
```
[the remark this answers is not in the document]
garypolzin: yea that makes complete sense!

PR #761, opened 2025-04-07
Create log value column

garypolzin:
# Description
Creating a function in advance of the model features which takes a log of another column value

# How to test
Not used in the pipeline yet but tests passing

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/Ng6NUk4X/781-epic-5-add-log-column-must)


RoyPriceCS50 (approved):
Approved with two changes:
- Presuming this is going to be used over and over, on the function return, you could name the column input column + "_logged_value". Like we do with the categorical labels.
- Put natural log in the doc string

PR #763, opened 2025-04-07
Add date index column

garypolzin:
# Description
The models we're using can't take dates as features so we're converting them to a date index, which basically means that the first date which appears in the file is given an index of 1, the second date to appear gets 2 and so on.

# How to test
tests passing, not in the pipeline as a feature output yet, will be added alongside model transition

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/JakRnYdC/779-epic-5-create-date-index-column-must)

### DIFF
--- a/jobs/prepare_features_care_home_ind_cqc.py
+++ b/jobs/prepare_features_care_home_ind_cqc.py
@@ -20,6 +20,7 @@ from utils.feature_engineering_resources.feature_engineering_services import (
 )
 from utils.features.helper import (
     add_array_column_count,
+    add_date_index_column,
     expand_encode_and_extract_features,
     vectorise_dataframe,
 )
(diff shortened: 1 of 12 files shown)


RoyPriceCS50 (approved):
Why not use unix time to also maintain gap between dates?

garypolzin:
> Why not use unix time to also maintain gap between dates?

@RoyPriceCS50 it appears the models are not a fan of that!

found out the hard way but the models work best when the units of time align with the frequency of data. It struggled to identify relationships in the data when time was in days (unix time) and the cuts of data were monthly - it just predicted a straight line trend, completely wiped out the covid dip and everything else.

Once I switched it so each file data was one more than the one before, it suddenly learned the trends really well and predictions were miles better!

PR #767, opened 2025-04-08
Stop non dormancy features as at 1/1/2025

garypolzin:
# Description
The 'with_dormancy' model started in 2022 and is an improvement on the 'without_dormancy' model.
In other to ensure a smooth transition between the two models, we predict both models alongside each other for a 3 year period.
The features dataframe will be filtered in line with the point at which the model was last retrained.

# How to test
Not in the pipeline yet but tests passing

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


RoyPriceCS50 (approved):
Just 1 comment on test data

PR #772, opened 2025-04-09
Removed references to rolling average

garypolzin:
# Description
It's rate of change now, so change the name to not reference rolling average


RoyPriceCS50 (approved):
Looks okay

PR #774, opened 2025-04-09
Refactor non res features

garypolzin:
# Description
Follows the same format as the care home, but then splits for without dormancy model and then with dormancy model.
- capped service count and activity count
- added capped and logged time in role (model worked better with that feature)
- removed the column selection so all columns are saved instead of just a subset
- patched function calls
- split test schemas up, previously both used the same one which wasn't ideal

# How to test
[Branch ran ok](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:non-res-features-Ind-CQC-Filled-Post-Estimates-Pipeline:e3161eb1-0527-456f-8a11-ea56e96a9396)

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/JLIWSUET/788-epic-5-refactor-non-res-features-must)

### DIFF
--- a/jobs/diagnostics_on_capacity_tracker.py
+++ b/jobs/diagnostics_on_capacity_tracker.py
@@ -51,7 +51,7 @@ estimate_filled_posts_columns: list = [
 absolute_value_cutoff: float = 10.0
 percentage_value_cutoff: float = 0.25
 standardised_value_cutoff: float = 1.0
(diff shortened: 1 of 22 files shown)


RoyPriceCS50 (approved):
Looks good.

PR #783, opened 2025-04-14
Amend patch paths

garypolzin:
# Description
The patches are long and messy, and often point to the place where the function is saved, as opposed to where it is called. This new format hopefully makes patching more obvious (and shorter)

# How to test
Unit tests still passing

# Developer checklist
- [X] Unit test


tmehmed (approved):
Use of F-strings makes the test for job role look much more concise, i think this is a good idea and can't see any problems in the PR. Happy for it to be merged.

PR #789, opened 2025-04-16
Remove add_log_column function

garypolzin:
# Description
This function was added in prep for the new non-res models but it turns out they are better without it, so taking it back out again.
Haven't rerun the pipeline as it's not actually used anywhere.

# How to test
All other tests still pass

# Developer checklist
- [X] Unit test


RoyPriceCS50 (approved):
Looks good

PR #790, opened 2025-04-16
Rename features files

garypolzin:
# Description
In half the places our features files are prefixed with 'features' to make them easy to find.
In other places they are suffixed with 'features', which makes them hard to find!
This aligns them so they are all prefixed.

This renaming broke the patch links, so I've also switched them to the new PATCH_PATH format

# How to test
No code changes so circleci checks passing should be sufficient

# Developer checklist
- [X] Unit test


jobs/validate_features_non_res_ascwds_inc_dormancy_ind_cqc_data.py
RoyPriceCS50, line 1: Small thing but since you're renaming stuff, here you use "...inc_dormancy...."
All the other times you use "...with_dormancy...."

RoyPriceCS50 (approved):
Approved with one comment, i've no strong feels either way. You know the context better than me.

PR #792, opened 2025-04-16
Change rolling sum to use interpolated counts

RoyPriceCS50:
# Description
I've changed the primary service rolling sum of job role counts to use interpolated counts instead of filtered counts.
Also change test data/schema for this function so previous unit tests pass.

After the rolling sum of counts are made, they get transformed into ratios. The function for this uses aggregation, which needs all elements passed in to be the same type. 
Until now we've always passed in integers (longs), but the interpolated counts are floats (doubles), so the rolling sum now gives a map column of type string:double.
So I've also changed the transform counts to ratios function to handle this new case.
I've also changed the tests for this function since I've changed how the function works.

# How to test
Unit tests pass.
[Link to pipeline run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:change-rolling-sum-to-use-inte-Ind-CQC-Filled-Post-Estimates-Pipeline:914d385e-f5b1-4bd4-8c25-b585a4590c10)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/utils/estimate_filled_posts_by_job_role_utils/utils.py
+++ b/utils/estimate_filled_posts_by_job_role_utils/utils.py
@@ -175,7 +175,7 @@ def transform_job_role_count_map_to_ratios_map(
 
     Args:
         df (DataFrame): A dataframe containing a job role count map at workplace level.
-        count_map_column_name (str): A map column of type any:long.
+        count_map_column_name (str): A map column of type string:number.
         ratio_map_column_name (str): The name to give to the ratio map column.
 
     Returns:


utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -205,25 +205,35 @@ def calculate_total_sum_of_values_in_a_map_column(
...
-    )
+    elif df.schema[map_column_name].dataType == MapType(StringType(), DoubleType()):
+        df = df.withColumn(
+            total_sum_column_name,
+            F.aggregate(
+                F.map_values(F.col(map_column_name)),
+                F.lit(0).cast(DoubleType()),
+                lambda a, b: a + b,
+            ),
+        )
```
garypolzin, line 236: I think you can slightly tweak this so instead of the if statement, you get the map type, and then you apply that original maptype back to the F.lit(0)

so extract as you did before but store it as a value
`map_type = df.schema[map_column_name].dataType`

then I don't think you need the if statement at all, it's just
`F.lit(0).cast(map_type.valueType)`

tests/test_file_schemas.py
```diff
@@ -6465,7 +6465,7 @@ class EstimateIndCQCFilledPostsByJobRoleUtilsSchemas:
         ]
     )
 
-    create_total_from_values_in_map_column_schema = StructType(
+    create_total_from_values_in_map_column_when_counts_are_integers_schema = StructType(
```
garypolzin, line 6468: I know they are basically the same but might be best if you are consistent in the naming of tests and type in the schema with integer/long and floats/double

utils/estimate_filled_posts_by_job_role_utils/utils.py
garypolzin: you could add an 'isinstance' check if you wanted to to check if it's a maptype and/or maptype with value type of either long/double if you wanted to but I'll leave that up to you. We don't tend to go to that extreme in most of our functions

RoyPriceCS50: Yeah I tried isinstance but it could only see if a type was maptype, it couldn't see the types within the map.

garypolzin: ah that's what valueType should do I think, so
`if isinstance(map_type.valueType, (LongType, DoubleType)):`

https://spark.apache.org/docs/latest/sql-ref-datatypes.html#:~:text=MapType(keyType%2C%20valueType%2C%20valueContainsNull,allowed%20to%20have%20null%20values.

RoyPriceCS50: Decided not to bother checking the type for long/double. Function would work with ints and floats anyway, if someone tried to sum some strings they've got issues.

utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -205,22 +205,23 @@ def calculate_total_sum_of_values_in_a_map_column(
     total_sum_column_name: str,
 ) -> DataFrame:
     """
-    Adds a column which contains the total of values from a given map column of type any:long.
+    Adds a column which contains the total of values from a given map column of type string:long or string:double.
 
     Args:
         df (DataFrame): A dataframe containing a count map.
-        map_column_name (str): A map column of type any:long.
+        map_column_name (str): A map column of type string:long or string:double.
```
garypolzin, line 212: think you can take out long/double now?

utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -1,5 +1,5 @@
 from pyspark.sql import DataFrame, functions as F
-from pyspark.sql.types import LongType
+from pyspark.sql.types import LongType, MapType, StringType, DoubleType
```
garypolzin, line 2: can some/all of this import be removed?

garypolzin (approved):
happy to improve with one or two things to look at

PR #795, opened 2025-04-17
Make gold validations more timely

garypolzin:
# Description
Currently all the gold validation happen in one go at the end, which can fail up to an hour after the job has run.
I've moved each validation step to be directly after the job has run, in parallel with the job after it.

Renamed the text in each cell to be shorter and neater

# How to test
[Branch run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:validations-more-timely-Ind-CQC-Filled-Post-Estimates-Pipeline:52b6805b-d0b6-48c1-aedd-e1d99b457950)


RoyPriceCS50 (approved):
Looked at this since the other PR failed validation.
Went through the visual guide rather than the text, and checked they all pointed to right jobs.
Looks okay.

PR #799, opened 2025-04-22
Add job specific source directory

garypolzin:
# Description
Currently every single job we make is saved in the jobs folder.
Looking at changing this to better structure our folders, so the first step is to make the file paths unique to each job.

Currently this is auto set to "../../jobs"
Now it's auto set to "../../" and I've set each individual one to have a path of "jobs"

So no actual changes yet, just a bit of prep work

# How to test
The majority of our jobs are called as part of the IndCQC step function, so I've run that and [it ran ok](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:add-source-dir-Ind-CQC-Filled-Post-Estimates-Pipeline:a12685c6-a6f5-44e1-87d7-01e359463929)


RoyPriceCS50 (approved):
Looks good.
61 jobs in this branch and 61 in other branchs

PR #802, opened 2025-04-23
Apply dq rules job group perc boundaries

RoyPriceCS50:
# Description
I've added a utils function that:
- gets the job group ratio value at the 0.001 and 0.999 percentiles, by primary service type, as list of two elements.
- pivots those lists into four columns, one for each job group, then packages those columns into a dict, like this:
  {direct_care: [0.1, 0.9], managers: [0.2, 0.7]....}
- Joins those lists into the original job role dataframe by primary service.
- Removes values from ascwds_job_role_counts_filtered when job group ratios are outside of boundaries
  The boundaries live in utils > scale_variable_limits script (where other limit stuff is)

In the test data, there are six locations per service type (12 total but values are the same in each 6). 
The first 5 breaks the limits and the 6th is kept.
The first 4 are upper limit breaks, the 5th is lower direct care limit breaking.

# How to test
Unit tests pass.
[Branch is running](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:apply-dq-rules-job-group-perc--Ind-CQC-Filled-Post-Estimates-Pipeline:2dde989c-12ab-417f-8763-5fefb8f79cdb)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_estimate_ind_cqc_filled_posts_by_job_role_utils.py
+++ b/tests/unit/test_estimate_ind_cqc_filled_posts_by_job_role_utils.py
@@ -1232,11 +1232,8 @@ class ApplyQualityFiltersToAscwdsJobRoleData(
         filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one_mock: Mock,
         filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_mock: Mock,
     ):
-        returned_df = job.filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one(
+        job.apply_quality_filters_to_ascwds_job_role_data(
             self.test_df,
-        )
-        job.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
-            returned_df
         )
 
         filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one_mock.assert_called_once()
@@ -1299,7 +1296,7 @@ class FilterAscwdsJobRoleCountMapWhenJobGroupRatiosOutsidePercentileBoundaries(
             Schemas.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_schema,
         )
         self.returned_df = job.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
-            self.test_df, 0.001, 0.999
+            self.test_df, lower_percentile_limit=0.001, upper_percentile_limit=0.999
         )
 
     def test_filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_does_not_change_row_count(
--- a/tests/unit/test_estimate_ind_cqc_filled_posts_by_job_role_utils.py
+++ b/tests/unit/test_estimate_ind_cqc_filled_posts_by_job_role_utils.py
@@ -1275,6 +1275,42 @@ class FilterAscwdsByJobRoleBreakdownWhenDirectCareOrManagersPlusRegulatedProfess
         self.assertEqual(expected_data, returned_data)
 
 
+class FilterAscwdsJobRoleCountMapWhenJobGroupRatiosOutsidePercentileBoundaries(
+    EstimateIndCQCFilledPostsByJobRoleUtilsTests
+):
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.test_df = self.spark.createDataFrame(
+            Data.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_rows,
+            Schemas.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_schema,
+        )
+        self.expected_df = self.spark.createDataFrame(
+            Data.expected_filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_rows,
+            Schemas.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_schema,
+        )
+        self.returned_df = job.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
+            self.test_df
+        )
+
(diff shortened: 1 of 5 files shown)


utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -662,6 +674,94 @@ def filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_profe
     return df
 
 
+def filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
+    df: DataFrame,
+) -> DataFrame:
+    """
+    Sets ascwds_job_role_counts_filtered to null when job group ratios outside of boundaries.
+
+    The boundaires are:
```
garypolzin, line 683: spelling :)

utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -662,6 +674,94 @@ def filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_profe
...
+    )
+
+    df = df.join(df_exploded, on=IndCQC.primary_service_type, how="left")
+
+    df = df.withColumn(
+        IndCQC.ascwds_job_role_counts_filtered,
+        F.when(
+            (
+                F.col(IndCQC.ascwds_job_group_ratios)[JobGroupLabels.direct_care]
+                > F.col(percentile_boundaries)[JobGroupLabels.direct_care][0]
```
garypolzin, line 736: just to help with readability of this can you reference these values above, like

```
direct_care_ratio = F.col(IndCQC.ascwds_job_group_ratios)[JobGroupLabels.direct_care]
direct_care_lower_limit = F.col(percentile_boundaries)[JobGroupLabels.direct_care][0]
etc
```

then the function becomes
```
      F.when(
            direct_care_ratio > direct_care_lower_limit
            &  direct_care_ratio < direct_care_upper_limit
            etc etc
```

utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -662,6 +674,94 @@ def filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_profe
...
+    df: DataFrame,
+) -> DataFrame:
+    """
+    Sets ascwds_job_role_counts_filtered to null when job group ratios outside of boundaries.
+
+    The boundaires are:
+        direct_care ratio value >= 0.001 percentile and <= 0.999 percentile
+        managers ratio value <= 0.999 percentile
+        regulated_professions ratio value <= 0.999 percentile
+        other ratio value <= 0.999 percentile
```
garypolzin, line 687: we have these figures referenced elsewhere so maybe don't put the actual values in the docstring, if we decide to tighten them we'll change the values elsewhere in the code but won't know to update this docstring

still worth having the rules, but just specify upper/lower limit

might even be worth having them as an argument to the whole function - at the moment the tests will break if we change those figures

it might not be worth changing the test data now but it makes writing the tests easier too, as you can set your test limits to a nice easy number like 50% or whatever and they can become a lot simpler

utils/scale_variable_limits.py
```diff
@@ -10,3 +10,9 @@ class GeneralLimits:
 class AscwdsScaleVariableLimits(GeneralLimits):
     total_staff_lower_limit = GeneralLimits.non_zero_lower_bound
     worker_records_lower_limit = GeneralLimits.non_zero_lower_bound
+
+
+@dataclass
+class AscwdsWorkerJobGroupRatioPercentileLimits:
+    lower_percentile_limit: float = 0.001
+    upper_percentile_limit: float = 0.999
```
garypolzin, line 18: can you write a test to check these values are correct (we have similar ones for number_of_days_in_window elsewhere)

garypolzin (commented):
on the whole it looks great for a pretty tricky PR, added a couple of comments but they are mostly minor tweaks I think/hope!

utils/estimate_filled_posts_by_job_role_utils/utils.py
RoyPriceCS50: I'll add the limits as arguments and remove them from scale_varibale_limits.
I can't think of better test data since the limit given as an argument is not literally the limit. It gets the value at that percentile across all the cases.
So I first thought of having 10 cases from 0.1 to 1.0, upper limit as 0.9, then it removes the 1.0... but then you can interpret that as literally being the limit.

tests/unit/test_estimate_ind_cqc_filled_posts_by_job_role_utils.py
```diff
@@ -1224,14 +1224,23 @@ def setUp(self) -> None:
...
         filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one_mock: Mock,
+        filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries_mock: Mock,
     ):
-        job.filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one(
+        returned_df = job.filter_ascwds_job_role_map_when_direct_care_or_managers_plus_regulated_professions_greater_or_equal_to_one(
             self.test_df,
         )
+        job.filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
+            returned_df
+        )
```
garypolzin, line 1240: took me a while to figure out how this test is passing but it's because we're calling two jobs by name, and then checking that those two jobs were called

it should be running `job.apply_quality_filters_to_ascwds_job_role_data` here
you don't need `returned_df = ` as we're not doing anything with it

utils/estimate_filled_posts_by_job_role_utils/utils.py
```diff
@@ -619,6 +622,12 @@ def apply_quality_filters_to_ascwds_job_role_data(
         df
     )
 
+    df = filter_ascwds_job_role_count_map_when_job_group_ratios_outside_percentile_boundaries(
+        df,
+        0.001,
+        0.999,
```
garypolzin, line 628: for clarify we normally add the name too for integers, so 
df,
lower_percentile_limit=0.001,
upper_percentile_limit=0.999,

garypolzin (approved):
couple of minor tweaks but happy to approve now

PR #803, opened 2025-04-23
Create basic folder structure with README files

garypolzin:
# Description
Creates the basic folder structure for all projects including a simple README file in each project folder.
The README's all need work, but created these so we have something to work with.

No actual code to test


garypolzin:
> Approved but missing _09_validations from independent_cqc folder?

Good spot. Validations are a bit of an outlier as they will be present throughout each file, but wanted to mention them

PR #807, opened 2025-04-23
Functions required for automatic model retraining 

garypolzin:
# Description
There are 5 functions (and tests) in this PR which will be required for the automated model process, these can be grouped into two categories

## Run numbers
As this process will be automatic, I want to save every model it makes so I'm going to give it a run number. So the first step every time the pipeline is run will be to
- get all the model run numbers already saved - `get_existing_run_numbers`
- get_model_s3_path
     - when **saving** a new model, we want the filepath to have a run number which is one higher than the max existing model number. If one has never been saved before then set run=1
     - when **loading** a model, we want the the filepath to include the maximum run number (to get the one saved above) . If it can't find one it raises an error

## Modelling process
The modelling process can be split into 3 parts, with the first two currently happening manually in EMR scripts
- train the model
- save the model (with the latest run number from `get_model_s3_path`)
- load the model (with the latest run number from `get_model_s3_path`)

# How to test
Not incorporated into the pipeline yet so mostly relying on tests passing

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
+++ b/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
@@ -1,9 +1,9 @@
 import unittest
 import warnings
-from unittest.mock import MagicMock, Mock, patch
+from unittest.mock import ANY, MagicMock, Mock, patch
 
 from datetime import date
-from pyspark.sql import functions as F
+from pyspark.sql import DataFrame, functions as F
 from pyspark.ml.linalg import Vectors
 from pyspark.ml.regression import LinearRegressionModel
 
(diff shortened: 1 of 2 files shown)


utils/estimate_filled_posts/models/utils.py
```diff
@@ -144,3 +150,119 @@ def convert_care_home_ratios_to_filled_posts_and_merge_with_filled_post_values(
...
+    existing = s3.list_objects_v2(Bucket=bucket, Prefix=prefix).get("Contents", [])
+    run_numbers = []
+    for obj in existing:
+        match = re.search(r"run=(\d+)/$", obj["Key"])
+        if match:
+            run_numbers.append(int(match.group(1)))
+    return sorted(set(run_numbers))
+
+
+def get_model_s3_path(model_source: str, mode: str = "load") -> str:
```
RoyPriceCS50, line 210: Is it worth breaking up the model_source?
Could have "s3://pipeline-resources/models/prediction/" fixed in the function, then argument becomes just the versioning "1.0.0/".

tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
```diff
@@ -257,3 +264,167 @@ def test_returned_column_values_match_expected_when_not_care_home(self):
...
+    @patch(f"{PATCH_PATH}.get_model_s3_path")
+    @patch(f"{PATCH_PATH}.LinearRegressionModel")
+    def test_save_model_to_s3(
+        self, LinearRegressionModel_mock: Mock, get_model_s3_path_mock: Mock
+    ):
+        mock_model = MagicMock()
+        get_model_s3_path_mock.return_value = f"{self.model_source}run=4/"
+
+        returned_path = job.save_model_to_s3(mock_model, self.model_source)
+        expected_path = f"{self.model_source}run=4/"
```
RoyPriceCS50, line 408: I think if you change save_model_to_s3 to return a f string, then the expected return in this test is the message. 
So it wont look like you're giving it the path, then checking the path you've just given is expected.

RoyPriceCS50 (approved):
Approved with some ideas commented.

PR #819, opened 2025-04-28
Create test and train datasets

garypolzin:
# Description
Split features dataset into test and train datasets
There is a prewritten function to do this - we're splitting so 80% of the features dataset will train the model and we can test model performance on the remaining 20%.

# How to test
Not currently in the pipeline but tests are passing

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
+++ b/tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
@@ -425,3 +425,38 @@ class LoadLatestModelTests(EstimateFilledPostsModelsUtilsTests):
 
         mock_model_load.assert_called_once_with(f"{self.model_source}run=4/")
         self.assertEqual(result, mock_model)
+
+
+class CreateTestAndTrainDatasetsTests(EstimateFilledPostsModelsUtilsTests):
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.test_df = self.spark.createDataFrame(
+            Data.create_test_and_train_datasets_rows,
+            Schemas.create_test_and_train_datasets_schema,
+        )
+        (
+            self.returned_train_df,
+            self.returned_test_df,
+        ) = job.create_test_and_train_datasets(
+            self.test_df,
+            test_ratio=0.2,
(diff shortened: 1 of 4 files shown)


tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
```diff
@@ -432,3 +433,79 @@ def test_load_latest_model_loads_expected_model(
...
+            sorted(self.returned_train_df.columns),
+            sorted(self.test_df.columns),
+        )
+
+    def test_create_test_and_train_datasets_returns_original_number_of_rows(self):
+        returned_train_row_count = self.returned_train_df.count()
+        returned_test_row_count = self.returned_test_df.count()
+
+        self.assertGreaterEqual(returned_train_row_count, 1)
+        self.assertGreaterEqual(returned_test_row_count, 1)
```
RoyPriceCS50, line 466: I'd have done an assert almost equal here.
Bit more faff but at least it checks the split proportions for each dataframe. Could be backwards in these tests and still pass.

Cast the returned counts as floats. Assert almost equal then becomes returned_count_as_float and test_df.count() multiplied by 0.8 and 0.2 (you've fixed the splitting of self.test_df).

RoyPriceCS50 (approved):
Approved with a comment to change how you've tested.

tests/unit/test_estimate_ind_cqc_filled_posts_models_utils.py
garypolzin: yea good point, it's a 0.2 split so I'll just test counts are 4 and 1, not sure why I didn't start with that to be honest!

RoyPriceCS50: I wondered if you'd change the length of the test_df at some point

garypolzin: I tried it exact rows at first but it gave me different rows so went straight to 'at least one', no idea why!

PR #820, opened 2025-04-28
CHECK 4 - Batch of functions to create s3 file paths based on model info

garypolzin:
# Description
Previously we've had 4 almost identical jobs, tests, glue jobs etc etc for each model
I'm refactoring these so we only need one version of each which we can call multiple times for each model
This part of the process is to auto-generate file paths in s3 based on the model name and version, and depending on which branch is being used

# How to test
unittests passing

# Developer checklist
- [X] Unit test


utils/estimate_filled_posts/models/utils.py
RoyPriceCS50, line 1: Do these functions belong in the project folder, instead of this old utils folder?

garypolzin: ideally yes, but some functions already existed and some are new, so just kept them all together for now

will be easier to batch move them when I delete out all the existing code which won't be needed any more

RoyPriceCS50 (approved):

PR #832, opened 2025-04-29
Epic 2 extrapolate job role breakdowns

edemirbag:
- [x] Unit test
- [x] [Linked to Trello ticket](https://trello.com/c/oRE5sxYQ/684-epic-2-extrapolation-must)

Roy - I've moved the extrapolation into the job roles folder, next to interpolation.
       - Refactored the job role extrapolation so it uses previous code where possible.
       - Simplified the test data

Another ticket is to combine the interpolation and extrapolation columns into one.

[Branch is running (with the problem function commented out)](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:epic-2-extrapolate-job-role-br-Ind-CQC-Filled-Post-Estimates-Pipeline:d6950adf-50cf-4cc0-832c-f272fde3968a)


tests/test_file_data.py
```diff
@@ -5558,6 +5558,260 @@ class ModelExtrapolation:
...
+            },
+        ),
+        (
+            "1-001",
+            1000000400,
+            None,
+            {
+                MainJobRoleLabels.care_worker: 0.5,
+                MainJobRoleLabels.registered_nurse: 0.5,
+            },
```
RoyPriceCS50, line 5669: This should be None. We don't want to fill gaps between two known points.

RoyPriceCS50 (changes requested):
Some changes please. Main issue is this function populates nulls when between two known points, it shouldn't do this since we cover that in interpolation.

Another ticket will be to combine the interpolated and extrapolated columns into another new column where we've populated around known points.

PR #833, opened 2025-04-30
Refactor of PIR process

garypolzin:
# Description
This is a beast of a PR but two thirds of it is deleted code and a lot of new code is data/schemas moved from old to new location, so hopefully it's not as bad as it first looks!

## The current/live process:
Impute job
- calls the 'blend' function which runs the PIR model, blends the data with ascwds and then deleted the modelled PIR column
- imputes the pir to follow the basic rolling average trendline called `imputed_non_res_pir_people_directly_employed`
Prepare features job (and validate features job)
- prepares the features for the non-res PIR model ready for the estimates job
Estimates job
- runs the PIR model

## Issue/Improvements
1. In the first bullet point above, we run the PIR model. The issue is that we then delete the column, then run it again later, but it's the exact same model, so the predictions will be exactly the same. So if we don't delete the column in the impute step, we can completely get rid of the prepare features, validate features and running the model again in estimates job as the data is already in there
2. previously we only impute the PIR figures by the basic rolling average, so I've switched this to being imputed by the non-res combined model, as that takes into things personal to the location (how long they've been open for etc)

## Summary
- After the PIR model is first run, don't drop the column
- everything related to the 'non_res_pir_linear_regression' model can go, as the model has already been run
- imputed using a better trendline and renamed so `non_res_pir_linear_regression_model `becomes `imputed_pir_filled_posts_model`

# How to test
[Branch ran successfully](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:pir-model-trendline-Ind-CQC-Filled-Post-Estimates-Pipeline:2e412eb0-f5c6-491e-90a0-118f435e78e6)

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket


RoyPriceCS50 (approved):
Looks okay

PR #834, opened 2025-05-01
Amend postcodes

garypolzin:
# Description
Two new ones entered incorrectly


RoyPriceCS50 (commented):
There's 4 postcodes causing the error, and they dont have spaces separating the characters.
RG411TP
CW15DN
M417GG
DH66QZ

garypolzin:
> There's 4 postcodes causing the error, and they dont have spaces separating the characters. RG411TP CW15DN M417GG DH66QZ

This process is to fix incorrectly entered postcodes.
If the postcode is legit, but so new that it's not in the postcode file yet, then we add it manually in EMR
so 2 are entered wrong, the other 2 are new

the format it shows in the error is after they've been cleaned, not the original, just for an added layer of confusion

garypolzin:
EMR script is in Marion's notebook, we choose the nearest available postcode to where the new one will be and use that info until it appears in the next postcode file

RoyPriceCS50 (approved):

PR #837, opened 2025-05-02
WIP - Split step functions

garypolzin:
# Description
We rarely update the original cleaning steps now so there's no point rerunning them everytime we update the pipeline
This splits the pipeline into two

One for cleaning & validating the original raw data
One purely for the Ind CQC pipeline

# How to test
Both running successfully
Clean and validate link
Ind CQC link


RoyPriceCS50 (commented):
Ind CQC pipeline failed on the diagnostics near the end. 
It can't find imputed_pir_filled_posts_model, but the estimate filled posts job creates it... not sure what it's problem is

garypolzin:
> Ind CQC pipeline failed on the diagnostics near the end. It can't find imputed_pir_filled_posts_model, but the estimate filled posts job creates it... not sure what it's problem is

@RoyPriceCS50 good spot, looks like it was unfortunate timing when I merged in the PIR stuff from main branch, half of the pipeline was run on the old PIR names and then the last step used the new names, hence couldn't find them

[it's passing now[(https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/split-clean-indcqc-pipelines-diagnostics_on_known_filled_posts_job/runs)

RoyPriceCS50 (approved):

PR #842, opened 2025-05-06
Cap extrapolation ratio to stop extreme changes

garypolzin:
# Description
Currently we are extrapolating using the ratio change of the modelled value, however this appears to be massively exaggerating change in some instances.

For example, if the modelled value goes from 1 filled post to 15 filled posts, that a 1500% ratio change. If a workplace submitted a value of 30 filled posts when the model predicts 1, we also apply that 1500% ratio so they increase from 30 to 450 posts, which is unlikely. We're switching this to limit to between 25% and 400% as the min/max change.

I've also added the minimum permitted value of 1.0 at the end of the process to ensure no-one ends up below that.

# How to test
Added/amended tests accordingly

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/9Ndbe7hh/815-review-extrapolation)


tests/test_file_data.py
```diff
@@ -4968,22 +4968,22 @@ class ModelImputationWithExtrapolationAndInterpolationData:
...
+        ("1-007", 10.0, -2.0, None),
+        ("1-008", 10.0, -2.0, 30.0),
     ]
     expected_imputation_model_rows = [
         ("1-001", None, None, None, None),
         ("1-002", None, None, 30.0, 30.0),
-        ("1-003", None, 20.0, None, 20.0),
-        ("1-004", None, 20.0, 30.0, 20.0),
+        ("1-003", None, -2.0, None, 1.0),
+        ("1-004", None, -2.0, 30.0, 1.0),
```
RoyPriceCS50, line 4982: Would this case be better as   loc-id, None, None, -2.0, 1.0
Proofs it overrule interpolation too.

tests/test_file_data.py
RoyPriceCS50, line 1: Not in this PR, but the schema expected_extrapolation_forwards_schema has the wrong name for the added field.
Nit picky change :-)

RoyPriceCS50 (approved):
Approved with some comments

PR #845, opened 2025-05-07
(low priority) Postcode matching template

garypolzin:
# Description
Creating a more robust postcode matching process as the whole pipeline fails every time a new postcode doesn't match exactly. Once complete this will reduce the amount of errors we get, so the pipeline fails less

It's a 5 part process, this is the first part of matching exactly

The function is in the main pipeline but it's not returning anything yet so it won't actually affect any outputs

# How to test
Unittests passing
Haven't run in branch as the function is called but doesn't return anything back into the pipeline

# Developer checklist
- [X] Unit test
- [X] Linked to Trello ticket

### DIFF
--- a/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
+++ b/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
@@ -1,0 +1,119 @@
+from pyspark.sql import DataFrame, functions as F
+from typing import Tuple
+
+import utils.cleaning_utils as cUtils
+from utils.column_names.raw_data_files.cqc_location_api_columns import (
+    NewCqcLocationApiColumns as CQCL,
+)
+from utils.column_names.cleaned_data_files.cqc_location_cleaned import (
+    CqcLocationCleanedColumns as CQCLClean,
+)
+from utils.column_names.cleaned_data_files.ons_cleaned import (
+    OnsCleanedColumns as ONSClean,
+)
+
+
+def run_postcode_matching(
+    locations_df: DataFrame,
+    postcode_df: DataFrame,
+) -> DataFrame:
+    """
+    Runs full postcode matching logic and raises error if final validation fails.
+
+    This function consists of 5 steps:
+        - Match postcodes where there is an exact match at that point in time.
+        - (TO DO) If not, use the first exact matching postcode for that location ID where available.
+        - (TO DO) If not, replace known postcode issues using the invalid postcode dictionary.
+        - (TO DO) If not, match the postcode based on the first half of the postcode only (truncated postcode).
+        - (TO DO) If not, raise an error to manually investigate any unmatched postcodes.
+
+    Args:
+        locations_df (DataFrame): DataFrame of workplaces with postcodes.
+        postcode_df (DataFrame): ONS postcode directory.
+
+    Returns:
+        DataFrame: Fully matched DataFrame.
+    """
+    locations_df = clean_postcode_column(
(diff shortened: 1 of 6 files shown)


projects/_01_ingest/cqc_api/utils/postcode_matcher.py
```diff
@@ -0,0 +1,119 @@
...
+def run_postcode_matching(
+    locations_df: DataFrame,
+    postcode_df: DataFrame,
+) -> DataFrame:
+    """
+    Runs full postcode matching logic and raises error if final validation fails.
+
+    This function consists of 5 steps:
+        - Match postcodes where there is an exact match at that point in time.
+        - (TO DO) If not, use the first exact matching postcode for that location ID where available.
```
RoyPriceCS50, line 25: I think this is a bad idea. If a location moved to a new building then it's genuine, even if they got the details wrong in the update. If CQC never correct their address, we'll always stick them in the wrong place.

projects/_01_ingest/cqc_api/utils/postcode_matcher.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+    locations_df = cUtils.add_aligned_date_column(
+        locations_df,
+        postcode_df,
+        CQCLClean.cqc_location_import_date,
+        ONSClean.contemporary_ons_import_date,
+    )
+
+    original_matched_df, original_unmatched_df = join_postcode_data(
```
RoyPriceCS50, line 52: Instead of original, call it locations_matched_df / locations_unmatched_df

projects/_01_ingest/unittest_data/ingest_test_file_data.py
```diff
@@ -549,3 +549,32 @@ class ValidateASCWDSWorkerCleanedData:
...
     ]
+
+
+@dataclass
+class PostcodeMatcherData:
+    # fmt: off
+    locations_where_all_match_rows = [
+        ("1-001", date(2020, 1, 1), "AA1 1aa"),
+        ("1-001", date(2025, 1, 1), "AA1 1aa"),  # lower case but matches ok
+        ("1-002", date(2020, 1, 1), "AA1 ZAA"),  # wrong now but amended later (match to the first known one, not the second)
```
RoyPriceCS50, line 560: This test case will be the first known postcode for loc 1-002, so it's future rows for this location that are wrong will get assigned this period.

This one should be found in the postcode data, and a later one should be wrong?

RoyPriceCS50 (approved):
Approved with comment on missing tests (fair enough since WIP), doc string comment, variable name change in function, and test data

projects/_01_ingest/cqc_api/utils/postcode_matcher.py
garypolzin: it shouldn't be our responsibility to ensure that the CQC postcodes are perfect

if it was widespread then I'd agree, but it's got to be niche that places move location, mess up their address and never fix it. they'd have to move quite far too so it was in a completely different LA/ICB area

projects/_01_ingest/unittest_data/ingest_test_file_data.py
garypolzin: 'known' refers to it being in the postcode directory

the postcode "AA1 ZAA" has been incorrectly typed so it doesn't match with any known postcode data.

the following month they correct to "AA1 2AA" which is a known postcode. so in this scenario, the postcode "AA1 ZAA" would change to "AA1 2AA", and the match is successful

PR #853, opened 2025-05-08
Speed up CQC locations job

garypolzin:
# Description
Now we're getting more data this is slowing down, it's taking twice as long as it used to and we haven't added new functions.

Gave it the same settings as most other jobs we have


RoyPriceCS50 (approved):
Fine mate

PR #854, opened 2025-05-08
Refactor recalc managers

RoyPriceCS50:
# Description
I've got the run-time using main down to 24 minutes.

- Used the version of the function without a for loop.
- Added frequent .cache and .count calls. Obviously not optimal but it works.

# How to test
Unit test passed.
[Most recent branch run was 24 minutes using main](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/refactor-recalc-managers-estimate_ind_cqc_filled_posts_by_job_role_job/runs)

# Developer checklist
- [x] Unit test
- [x] Linked to Trello ticket


garypolzin (approved):
do we need the counts, or just the cache?

RoyPriceCS50:
> do we need the counts, or just the cache?

ChatGPT says yes. The count is was makes the job do something, so caching alone will just go into the plan then happen at the writing stage.

PR #857, opened 2025-05-09
Only include large PIR places if they have multiple submissions

garypolzin:
# Description
people just type one number into PIR so it's hard to validate it against anything.
Small places are fine as they don't really affect anything, so this is just wiping out any large places (100+) who have only submitted once.

Follow up filter will be to use those multiple submissions to validate each other - if always big then they are big, if there is a wide range between max/min then we can get rid of them

# How to test
Hard to do a proper before and after but essentially it's the same trend but takes about 5k off the most recent periods and flattens the recent upward trajectory a bit

non-res is red line, other two are CH and are unchanged

image

# Developer checklist
- [X] Unit test


RoyPriceCS50 (approved):
Looks good

PR #858, opened 2025-05-09
WIP - PIR outlier removal

garypolzin:
# Description
Using 2 methods to detect outliers as they target different issues

dispersion = (max - min) / mean

median absolute deviation = the median of the absolute differences between the submissions and the median of those submissions (docstrings contain better descriptions!)

Both run at the same time, if either of them fail then the values for the whole workplace are nulled

# How to test
NO TESTS WRITTEN (ran out of time)

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


RoyPriceCS50 (commented):
I've written the tests for these functions, and changed how the MAD function works.
It wasn't giving the median when location has two data points, it just gave the lower of the two values, and since it needs a median for the difference, I added extra steps to handle it.

But the estimates pipeline keeps failing in branch at validating the merged dataset. I can't work out why. It's fine in main, and I'm pointing the step function at main data sources.

PR #862, opened 2025-05-09
Include PIR in merged column if never submitted ASCWDS

garypolzin:
# Description
Now we've cleaned PIR data a bit we were contemplatlating how it would look if we fully merged it in with ASCWDS data

so the rules would be
- use ASCWDS if we have it
- if we have ASCWDS but it's out of date and PIR is substantially different then incorporate PIR
- this PR is the next step, if we've never had ASCWDS data then use PIR (it will prob have merge conflicts as I've created the clean version of PIR data since I started this branch

# How to test

# Developer checklist
- [ ] Unit test
- [ ] Linked to Trello ticket


RoyPriceCS50 (approved):
This changes the estimate source for thousands of locations from imputed_pir_filled_posts_model to imputed_posts_non_res_combined_model. 

So it does what you intend, since you've moved PIR data into ascwds_pir_merged, it's picking up that in the merge columns in order.

The impact on filled posts trend is negligible for care homes.
For non-res it's reducing the overall estimate, but the rate of change between years is pretty much the same.

Approving this PR to prevent blocking while on leave.

PR #870, opened 2025-06-02
First known postcode match

garypolzin:
# Description
For locations who don't match initially, this looks to find if they've ever had a match and if so, it gives them the first one which matches correctly.

Currently we've got nearly 150 postcodes in our invalid postcode list.
After this step, we only need 17 records in there - which means that most places do go on to fix their postcodes (a lot of those 17 looks to be ones just over the Welsh border

# How to test
Job ran successfully - added a print statement at the end to show which postcodes don't match which [is here](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws-glue$252Fjobs$252Foutput/log-events/jr_6074079ee9a1cd1da75e21c9ff86519b607e10eb8bb35393e6041bdc116c7c24)

# Developer checklist
- [X] Unit test


RoyPriceCS50 (approved):
This looks fine. Maybe the runtime will get excessive as you add more steps are repeatedly join the postcode_df.

In get_first_successful... function I'd have collected all the geography columns from the match_df instead of just the postcode, renamed them all and done withColumn replacement for each of them.
Then returned a tuple of matched/unmatched.

Then there's no need to join to postcode again.

garypolzin:
> This looks fine. Maybe the runtime will get excessive as you add more steps are repeatedly join the postcode_df.
> 
> In get_first_successful... function I'd have collected all the geography columns from the match_df instead of just the postcode, renamed them all and done withColumn replacement for each of them. Then returned a tuple of matched/unmatched.
> 
> Then there's no need to join to postcode again.

Sadly you can't do that as we have different geogs over time, so if a postcode was incorrect in 2020 and they fixed in 2021, you'll end up copying the 2021 geography areas into the 2020 dataset

PR #873, opened 2025-06-03
Remove dodgy ASCWDS accounts

garypolzin:
# Description
Removed two sets of unusual ASCWDS data

One group of 4 (2 locs in 2 orgs) who frequently submit exactly the same BUDI files into ASCWDS on exactly the same day.

One group of 18, who are seemingly completely unrelated but submitted the same data on the same day for all 18 accounts.

This PR is to simply remove them during the cleaning step. Once Support have called them we might get more info, but I can't imagine anything will happen which means we want to reintroduce them

# How to test
Unittests passing
Athena outputs no longer contain them

# Developer checklist
- [X] Unit test

### DIFF
--- a/tests/test_file_data.py
+++ b/tests/test_file_data.py
@@ -7262,6 +7262,10 @@ class RawDataAdjustments:
     worker_data_without_rows_to_remove = expected_worker_data
 
     workplace_data_with_duplicates_rows = [
+        ("20250101", "48904", "other_data"),
+        ("20250101", "49966", "other_data"),
+        ("20250101", "49967", "other_data"),
+        ("20250101", "49968", "other_data"),
         ("20250101", "50538", "other_data"),
         ("20250101", "50561", "other_data"),
         ("20250101", "50590", "other_data"),


tests/test_file_data.py
```diff
@@ -7261,6 +7261,33 @@ class RawDataAdjustments:
...
+        ("20250101", "50639", "other_data"),
+        ("20250101", "50640", "other_data"),
+        ("20250101", "50767", "other_data"),
+        ("20250101", "50769", "other_data"),
+        ("20250101", "50770", "other_data"),
+        ("20250101", "50771", "other_data"),
+        ("20250101", "50869", "other_data"),
+        ("20250101", "50870", "other_data"),
+        ("20250101", "67890", "other_data"),
+    ]
```
RoyPriceCS50, line 7285: You could put 48904, 49966, 49967, 49968 in here

utils/raw_data_adjustments.py
```diff
@@ -32,6 +29,51 @@ def remove_duplicate_worker_in_raw_worker_data(raw_worker_df: DataFrame) -> Data
...
+    These are duplicates in the sense that the same data has been uploaded to ASCWDS for multiple accounts.
+
+    This has been passed on to the support team (03/06/2025) to investigate which may affect what we do with
+    these locations long term but in the short term we're simply removing them from ASCWDS.
+
+    There are two sets of workplaces here.
+      - Four locations who submit the exact same ASCWDS files on the same day.
+      - 18 separate locations, seemingly unrelated, all submit identical data on the same day.
+    """
+    raw_workplace_df.sort
```
RoyPriceCS50, line 45: Why do you need this sort?

RoyPriceCS50 (approved):
Approved with two comments.

tests/test_file_data.py
garypolzin: yea they should be, good spot

PR #888, opened 2025-06-06
Identify and null non res grouped providers

garypolzin:
# Description
Added a process in to detect and null non-res grouped providers based on their size compared to their average PIR submissions

They are removed if they have at least 50 ASCWDS posts and
- 2.5 times the size of their location PIR submission
- 1.5 times the size of the sum of PIR submissions for all locations in the provider (providing at least 2 have submitted)

# How to test
unittests passing and [branch running](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:non-res-grouped-providers-Ind-CQC-Filled-Post-Estimates-Pipeline:53cbf14d-8668-43cf-8caa-1e4f5956218e)

# Developer checklist
- [X] Unit test
- [X] Linked to [Trello ticket](https://trello.com/c/oTIPwEyC/574-epic-coverage-identify-grouped-providers-for-non-res-should)


RoyPriceCS50 (approved):
I've approved this after making my own commits to change the functionality. I put notes in Teams and commit messages.

Also noting that this filter doesn't care if ASC-WDS is much smaller than PIR.

garypolzin:
> Also noting that this filter doesn't care if ASC-WDS is much smaller than PIR.

Yea this filtering is specifically for 'grouped providers' which is where the provider submits their whole workforce into one location, so we're looking at places where ASCWDS is much larger than what we're expecting (expectation is from PIR data)

If ASCWDS is a lot smaller than PIR then it's definitely not a grouped location, so all good (in terms of this filter anyway!)

PR #895, opened 2025-06-10
Truncated postcode match

garypolzin:
# Description
In order to try capture postcodes which aren't a perfect match I've created a matching process which matches postcodes without the last 2 characters. Some times they mess other parts of the postcode up but it's not our position to police the dataset. Locations tend to amend their postcode, so when they fix it they'll be captured by the earlier process of being reassigned to their first known postcode instead of the truncated process.

This will also capture the Welsh postcodes which are just over the border, as they will match to a similar postcode in England

# How to test
Added print statements to capture which postcodes were mapped this way and which are still unmatched after this step - only 3 unmatched.

**Matched postcodes**: [('1-22806116238', 'PL6 7BF'), ('1-1400353245', 'CO10 0F0'), ('1-18340608606', 'M24 4TL'), ('1-5502156131', 'CH4 0NR'), ('1-2034958646', 'CH1 4QL'), ('1-1837668758', 'CH4 0RG'), ('1-2952707685', 'CH1 6BS'), ('1-22953117116', 'HU6 7QT'), ('1-2085893771', 'CH1 4QL'), ('1-5502156131', 'CH4 8RQ'), ('1-18124659678', 'GU11 4TH'), ('1-22845529223', 'RG41 1TP'), ('1-22806116238', 'PL67BF'), ('1-550088530', 'N12 8FP')]

**Unmatched postcodes** : [('1-1180915503', 'CH5 2LY'), ('1-22164469648', 'TF7 3QH'), ('1-497248245', 'HU17 ORH)]

Checked the unmatched and it's correct that they haven't returned a match

# Developer checklist
- [X] Unit test

### DIFF
--- a/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
+++ b/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
@@ -195,8 +195,8 @@ def truncate_postcode(df: DataFrame) -> DataFrame:
     """
     return df.withColumn(
         CQCLClean.postcode_truncated,
-        F.substring(
-            CQCLClean.postcode_cleaned, 1, F.length(CQCLClean.postcode_cleaned) - 2
+        F.expr(
+            f"substring({CQCLClean.postcode_cleaned}, 1, length({CQCLClean.postcode_cleaned}) - 2)"
         ),
     )
 
--- a/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
+++ b/projects/_01_ingest/cqc_api/utils/postcode_matcher.py
@@ -96,6 +96,17 @@ def run_postcode_matching(
         errors = [(r[CQCL.location_id], r[CQCL.postal_code]) for r in rows]
         print(f"Unmatched postcodes found: {errors}")
 
+    print(f"locations_df count: {locations_df.count()}")
+    print(f"final_matched_df count: {final_matched_df.count()}")
+    print("")
+    print(f"unmatched_locations_df count: {unmatched_locations_df.count()}")
+    print(
+        f"unmatched_reassigned_locations_df count: {unmatched_reassigned_locations_df.count()}"
+    )
+    print(
+        f"unmatched_truncated_locations_df count: {unmatched_truncated_locations_df.count()}"
+    )
+
     return final_matched_df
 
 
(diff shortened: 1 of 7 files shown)


projects/_01_ingest/cqc_api/utils/postcode_matcher.py
```diff
@@ -181,3 +194,71 @@ def get_first_successful_postcode_match(
...
+        df (DataFrame): A DataFrame containing the full postcode.
+
+    Returns:
+        DataFrame: DataFrame with the truncated postcode added.
+    """
+    return df.withColumn(
+        CQCLClean.postcode_truncated,
+        F.expr(
+            f"substring({CQCLClean.postcode_cleaned}, 1, length({CQCLClean.postcode_cleaned}) - 2)"
+        ),
```
RoyPriceCS50, line 213: Use F.substring directly rather than expression.

projects/_01_ingest/cqc_api/utils/postcode_matcher.py
```diff
@@ -181,3 +194,71 @@ def get_first_successful_postcode_match(
...
+        CQCLClean.postcode_truncated,
+        ONSClean.contemporary_ons_import_date,
+        *grouping_cols,
+    )
+
+    df = df.withColumn(count_col, F.count("*").over(group_window))
+
+    rank_window = Window.partitionBy(
+        CQCLClean.postcode_truncated, ONSClean.contemporary_ons_import_date
+    ).orderBy(F.desc(count_col), *grouping_cols)
```
RoyPriceCS50, line 257: Your tests still pass without *grouping_cols in the orderBy.
I can't work out why it's needed.

RoyPriceCS50 (approved):
Approved with two comments. Works either way, they just tidy up.

projects/_01_ingest/cqc_api/utils/postcode_matcher.py
garypolzin: think the only way is to split it into two withColumns - that avoids the SQL expression but it's two calculations instead of one, do you have a preference?

```
df = df.withColumn(
    "postcode_length_minus_2",
    F.length(F.col(CQCLClean.postcode_cleaned)) - F.lit(2)
)

df = df.withColumn(
    CQCLClean.postcode_truncated,
    F.substring(F.col(CQCLClean.postcode_cleaned), 1, F.col("postcode_length_minus_2"))
).drop("postcode_length_minus_2")

```

PR #899, opened 2025-06-12
Standardising step function names

garypolzin:
# Description
They are a bit all over the place so neatened them up

Historical DPR has never been run, so I've deleted it

# How to test
These are just filenames and how they displayed in aws, nothing uses the name so haven't bothered running then
Main thing is terraform being happy


RoyPriceCS50 (approved):
Nice

PR #902, opened 2025-06-12
Remove populate estimate function with merge_columns_in_order

garypolzin:
# Description
We've replaced this function in the estimates script a while ago but didn't switch this function over.

Also spotted two magic numbers in the script, they were referenced at the top of the script but still had '370' in the script, not the reference

# How to test
[Glue job run](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/remove-populate-estimate-func-diagnostics_on_capacity_tracker_job/runs) - set this going so hopefully it finished ok!


RoyPriceCS50 (approved):
Approved with a note.
Your run failed but I just set it going again and it succeeded.
Job must be on the limit of resources available.

PR #904, opened 2025-06-16
WIP - update to CH model 6.2.0

garypolzin:
# Description
The interpolation script have been changed to only interpolate data produced within 6 months of each other (previously 12 months).

This model is the same as the previous (in terms of features etc) but it's been trained and will run the model based on this newer dataset

# How to test
Extensive testing on the outputs have been done outside of this PR in Tableau/Excel and compared against CT data


RoyPriceCS50 (approved):
Approved

PR #912, opened 2025-06-27
Top level updates from PR 705

garypolzin:
# Description
Marion opened [PR 705](https://github.com/NMDSdevopsServiceAdm/DataEngineering/pull/705) a few months ago.

I've taken the top level changes out of that PR but left all the 'job' stuff out as we've changed the structure a fair bit so might want to revisit how we do the rest. I'll close that PR so it's not hanging around

# Notes from previous PR
It is not possible to fully automate this documentation and if we were to adopt it properly, it would need maintaining with code changes, just like the doc strings.

## How to test

I'd recommend reviewing this PR by running the server locally. To do this you'll want to run the command: `sphinx-autobuild docs/source/ docs/build/` in your terminal. This will start the server and once it's ready, it will provide you with a URL. Control and click in VSCode will open the URL in your browser and you can explore the pages.

### DIFF
--- a/docs/source/_01_data_ingestion_pipelines/ingest_ons_pd.md
+++ b/docs/source/_01_data_ingestion_pipelines/ingest_ons_pd.md
@@ -60,7 +60,7 @@ These are typically resolved at runtime via the calling context or environment c
 
 ## Visual Representation
 
-```mermaid
+```{mermaid}
 graph TD
     A[Ingest ONS data] --> B{Clean and Validate}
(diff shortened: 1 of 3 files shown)


RoyPriceCS50 (commented):
I don't think this is working correctly. 
I just get a page with "STARTING POINTS" and link to the README.
Expected to see links to other pages like "Pipelines" and "Jobs".

It's formatted nicely at least

PR #915, opened 2025-06-30
Capacity Tracker CH and NR pipeline documentation

garypolzin:
# Description
Added pipeline documentation for both CT datasets (same workflow)

# How to test
Run `sphinx-autobuild docs/source/ docs/build/` to generate the link


garypolzin:
> Approved. Isn't this connected to doc strings somehow?

Yea we can add docstrings, but this is surely pipeline stuff so no docstrings in here
One of the next PRs will be for a job so you can see it in there

PR #957, opened 2025-08-19
[924] Define Polars-based deployment process for refactored code

tedmonds-mt:
## Description
Trello ticket [#924](https://trello.com/c/Q12pDdXA)

Adjusted the deployment of Fargate jobs so creating them is reproducible, and documented the process.

~I have used the .tftpl file extension for this step function in line with [terraform suggestions](https://developer.hashicorp.com/terraform/language/v1.2.x/functions/templatefile) because I was having trouble passing in a list of the subnet ids~ 

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:924-polars-deployment-polars-cqc-ingestion:030d4484-9b3e-41b1-9e13-aacf052eeab2)

The documentation for the process is in [Confluence](https://skillsforcare.atlassian.net/wiki/spaces/DE/pages/1516732417/How+To+Deploy+a+New+Fargate+Job)


terraform/pipeline/glue.tf
```diff
@@ -474,6 +474,7 @@ module "delta_clean_cqc_location_data_job" {
     "--cleaned_ons_postcode_directory_source" = "${module.datasets_bucket.bucket_uri}/domain=ONS/dataset=postcode_directory_cleaned/"
     "--cleaned_cqc_location_destination"      = "${module.datasets_bucket.bucket_uri}/domain=CQC/dataset=locations_api_cleaned/"
   }
+  extra_conf = " --conf spark.sql.autoBroadcastJoinThreshold=-1"
```
micahdougall, line 477: Ah I see - would this fail then without the space?

tedmonds-mt: It would, yeah, but didn't want a trailing space for all the rest of the glue jobs where there is no extra conf as i suspect AWS would strip it out and then it would be seen as a change every time (hence the space here rather than where you noted in the glue module)

terraform/pipeline/step-function.tf
```diff
@@ -574,6 +597,55 @@ resource "aws_iam_policy" "step_function_iam_policy" {
           "${module.datasets_bucket.bucket_arn}/*",
           module.datasets_bucket.bucket_arn
         ]
+      },
```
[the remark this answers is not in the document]
tedmonds-mt: Made it a JSON - if we make the change to template files we might as well do it all in one batch

PR #959, opened 2025-08-21
Update read_from_parquet function with optional schema parameter

AchyutRatkanthiwar:
## Description
Trello ticket [#961](https://trello.com/c/84zQqxaB/961-update-readfromparquet-with-schema-as-optional-parameter)

Added a new optional parameter in the read_from_parquet() function to future proof the function and keep it inline with polars function.

## Testing
- [x] Unit tests passing
- [x] Successful [Glue Job run](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/update-read-from-parquet-with--flatten_cqc_ratings_job/runs)

I have also added 4 tests which tests output of the function with different inputs. All passed.

No direct testing but I ran glue job flatten_cqc_ratings_job as it uses this function to check if my changes causes any issues. Link to glue job run is attached. 

- [x] Updated CHANGELOG

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -546,7 +546,7 @@ class GeneralUtilsTests(UtilsTests):
 
         self.assertCountEqual(df.columns, column_list)
 
-    def test_read_from_parquet_with_schema_extra_column_not_in_parquet(self):
+    def test_read_from_parquet_with_schema_extra_column_not_in_parquet_ignored(self):
         schema = StructType(
             [
                 StructField(CQCColNames.name, StringType(), True),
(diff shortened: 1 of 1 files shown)


tests/unit/test_utils.py
```diff
@@ -509,6 +509,90 @@ def test_read_from_parquet_only_imports_selected_columns(self):
...
+            ]
+        )
+
+        df = utils.read_from_parquet(self.example_parquet_path, schema=schema)
+
+        self.assertCountEqual(df.columns, [CQCColNames.name, "extra_col"])
+        null_count = df.filter(df["extra_col"].isNotNull()).count()
+        self.assertEqual(null_count, 0)
+
+    def test_read_from_parquet_with_empty_schema(self):
```
garypolzin, line 563: can you just add the expected behaviour in the test name please?

I'm guessing it's `test_read_from_parquet_with_empty_schema_imports_all_columns`?

garypolzin (approved):
Looks good, one minor change but happy for you to merge afterwards :)

tests/unit/test_utils.py
AchyutRatkanthiwar: ahh yes. makes sense. It will help understand what it tests. I will make those changes. Thanks

PR #972, opened 2025-09-01
[953] Refactor clean_cqc_locations_data to work on a delta dataset

tedmonds-mt:
## Description
Trello ticket [#953](https://trello.com/c/uihuIKSI)

Altered cleaning script to work on delta data. 
This includes changes to create the `time_registered_for` and `time_since_dormant` fields in the ind cqc pipeline, rather than in the cleaning script. This works because they are not used in the SfC Coverage pipeline. 

I have also modified the build snapshot function to cope with the locations cleaned schema. All schemas have been re-pointed to the correct schemas folder.

The final thing to note is that I have added "LegacyColumns" into the legacy scripts. This is so that I can make the column lists correct for the current live process, while keeping the legacy scripts functional. Doing it in this way, rather than by having a legacy column list means that the legacy scripts can be cleanly deleted down the line, when we are comfortable with how the new process is working.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:953-d-clean-locs-Transform-CQC-Data:194d92de-48a3-4ec8-a2ec-12b96a6ff04e)

Athena checks include: 
- `time_registered_for` and `time_since_dormant` are not in the cleaned locations data
- `time_registered_for` and `time_since_dormant` are in ind_cqc_merged data
- merged coverage data is query-able

~- [ ] Migration to polars - added comment to pyspark functions which have been migrated~
- [x] Updated CHANGELOG


projects/_03_independent_cqc/_01_merge/jobs/merge_ind_cqc_data.py
```diff
@@ -116,6 +115,8 @@ def main(
         cleaned_cqc_location_source,
         selected_columns=cleaned_cqc_locations_columns_to_import,
     )
+    cqc_location_df = calculate_time_registered_for(cqc_location_df)
+    cqc_location_df = calculate_time_since_dormant(cqc_location_df)
```
garypolzin, line 119: sorry for the faff but I think these will be better in the job after (cleaning IND CQC) - and keeping this one as the merging only

garypolzin (changes requested):
Unfortunately I think these functions will fit better in the follow up job to 'merge' which is the 'clean' job - but I think everything is looking good otherwise

tedmonds-mt:
We will also need to read in all of the delta dataset, and re-write them out with snappy compression

garypolzin (approved):
Agreed to hash out failing validations in filled posts by job role - will be caused by default data and we're planning on rewriting that job soon anyway so we can revisit validations during that process

tedmonds-mt:
Substantial changes were required to implement dimensions. The PR description has been updated to reflect the changes

projects/_01_ingest/cqc_api/jobs/delta_clean_cqc_location_data.py
```diff
@@ -613,25 +799,45 @@ def add_related_location_column(df: DataFrame) -> DataFrame:
...
 
     Returns:
-        DataFrame: A cleaned locations dataframe with locations which are only specialist colleges removed.
+        tuple[DataFrame, DataFrame]: cqq_df, gac_services_dimension with locations which are only specialist colleges removed.
     """
-    df = df.where(
-        (df[CQCLClean.services_offered][0] != Services.specialist_college_service)
-        | (F.size(df[CQCLClean.services_offered]) != 1)
-        | (df[CQCLClean.services_offered].isNull())
+    # noinspection PyCallingNonCallable
```
garypolzin, line 818: just checking what this is/means?!

tedmonds-mt: Ha, I'd forgotten about this. It just gets rid of a warning - because spark is so encapsulated and convoluted, the type hints it gives are sometimes unrecognised/recognised as wrong by IDEs. It doesn't really do anything except remove the visual noise of the yellow underline. I can remove it if you prefer

tedmonds-mt: (the warning is present in both the non-delta and the delta version, it's not from something I've added)

garypolzin: ah ok, wondering if it might be worth adding a short comment above it to say something like that? Just to save confusion once I've forgotten about this!

tedmonds-mt: No problem, will do!

PR #974, opened 2025-09-01
948 - Move ct imputation

RoyPriceCS50:
## Description
# 948 - Move Capacity Tracker care home and Non Res imputation functions into IND CQC pipeline

I've moved the imputation of capacity tracker data from the diagnostics job to the imputation job. Then the imputed cap tracker columns get imported into estimates and diagnostics jobs.

I've also:
- refactored and renamed the function combine_care_home_ratios_and_non_res_posts so I could use it to combine cap tracker data, as well as previously combined care home beds ratio and non-res posts.
- moved the combining function itself from the estimates > models > utils folder into the generic utils folder.
- moved associated tests/test data and schema.

[Step function run - using main](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:move-ct-imputation-Ind-CQC-Filled-Post-Estimates:17c7f18e-08b3-442f-a620-1eb16045ae4e)

This is the average capacity tracker value between original (darker colour) and imputed (lighter colour), filtered to recent periods because we are only interested in those. Capacity tracker data only started in 2022 so we will never use imputed values from say 2015:
image

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](add link)

- [ ] Migration to polars - added comment to pyspark functions which have been migrated
- [x] Updated CHANGELOG


projects/_03_independent_cqc/_03_impute/jobs/impute_ind_cqc_ascwds_and_pir.py
```diff
@@ -19,13 +19,15 @@
...
-    combine_care_home_ratios_and_non_res_posts,
     convert_care_home_ratios_to_filled_posts_and_merge_with_filled_post_values,
 )
 from projects._03_independent_cqc._03_impute.utils.model_and_merge_pir_filled_posts import (
     model_pir_filled_posts,
     merge_ascwds_and_pir_filled_post_submissions,
 )
+from projects._03_independent_cqc.utils.utils.utils import (
+    combine_care_home_and_non_res_values_into_single_column,
+)
```
garypolzin, line 30: Looks like this function is only called in this one particular job it can go in `projects._03_independent_cqc._03_impute.utils.utils` (there isn't a generic utils for this job yet so you'll need to create it)

we'll use `projects._03_independent_cqc.utils.utils.utils` for functions which only exist in the project `_03_independent_cqc` but are called by several of the jobs, not just one

garypolzin (commented):
Looks good but if you can move the function to a different utils folder that would be great please! Sorry, bit of a faffy one

projects/_03_independent_cqc/_03_impute/jobs/impute_ind_cqc_ascwds_and_pir.py
RoyPriceCS50: I've moved the function into a new impute > utils script. Also moved the tests into fresh script, and test data/schema into their own classes.

PR #975, opened 2025-09-01
[968] Switch CQC delta download to append mode

tedmonds-mt:
## Description
Trello ticket [#968](https://trello.com/c/PAc9xN05)

The Polars function for writing to parquet now defaults to appending, rather than overwriting, in line with Spark functionality. This affects the import jobs, which now both append data, rather than overwriting it - this means that if the import job is run multiple times on the same day, the data will all be available in the same folder.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run 1](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:968-delta-append-Ingest-CQC-API-Delta:c0ddc2df-6298-4e17-a1e4-efce1909a90c), followed by [run 2](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:968-delta-append-Ingest-CQC-API-Delta:aa223599-2190-484c-a9bf-42941a7e12f7) (to prove that it can be run multiple times on the same day), followed by [run 3](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:344210435447:execution:968-delta-append-Ingest-CQC-API-Delta:bd1cccf3-1007-410a-8917-e1ae4e01cd48) (to prove that if there are no changes it will still work). 
- [x] Outputs checked in Athena (including checking that the changes from the second run are present)

~- [ ] Docstrings added/updated~
~- [ ] Migration to polars - added comment to pyspark functions which have been migrated~
~- [ ] Updated CHANGELOG~

### DIFF
--- a/projects/_01_ingest/cqc_api/tests/fargate/test_delta_download_cqc_locations.py
+++ b/projects/_01_ingest/cqc_api/tests/fargate/test_delta_download_cqc_locations.py
@@ -51,18 +51,21 @@ class TestDeltaDownloadCQCLocations(unittest.TestCase):
         with self.assertRaises(InvalidTimestampArgumentError):
             main(dest, start, end)
 
+    @patch(f"{PATCH_PATH}.utils.uuid")
     @patch(f"{PATCH_PATH}.get_secret")
     @patch(f"{PATCH_PATH}.cqc.get_updated_objects")
     @patch(f"{PATCH_PATH}.SECRET_ID", new="cqc-secret-name")
     @patch(f"{PATCH_PATH}.AWS_REGION", new="us-east-1")
-    def test_main_writes_parquet(self, mock_objects, mock_get_secret):
+    def test_main_writes_parquet(self, mock_objects, mock_get_secret, mock_uuid):
         mock_get_secret.return_value = '{"Ocp-Apim-Subscription-Key": "abc1"}'
         mock_objects.return_value = [
             {"locationId": 1},
             {"locationId": 2},
             {"locationId": 3},
         ]
-        dest = f"{self.temp_dir}/data.parquet"
+        file_name = "abc"
+        mock_uuid.uuid4.return_value = file_name
+        dest = f"{self.temp_dir}/{file_name}.parquet"
         start = "2025-07-20T15:40:23Z"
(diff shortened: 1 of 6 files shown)


projects/_01_ingest/cqc_api/tests/fargate/test_delta_download_cqc_locations.py
```diff
@@ -95,3 +104,47 @@ def test_main_copes_with_malformed_destination(self, mock_objects, mock_get_secr
...
+        main(self.temp_dir + "/", middle, end)
+
+        self.assertTrue(pathlib.Path(expected_paths[0]).exists())
+        self.assertTrue(pathlib.Path(expected_paths[1]).exists())
+        self.assertTrue(pathlib.Path(expected_paths[0]).is_file())
+        self.assertTrue(pathlib.Path(expected_paths[1]).is_file())
+        self.assertTrue(pathlib.Path(expected_paths[0]).suffix == ".parquet")
+        self.assertTrue(pathlib.Path(expected_paths[1]).suffix == ".parquet")
+
+        result_a = pl.read_parquet(expected_paths[0])
```
micahdougall, line 147: Should this read in the whole dataset instead and verify that it has appended?

tedmonds-mt: So the expected behavior is that it will end up writing two files with different names, like it is here: https://sfc-968-delta-append-datasets.s3.eu-west-2.amazonaws.com/domain%3DCQC_delta/dataset%3Ddelta_providers_api/version%3D3.0.0/year%3D2025/month%3D09/day%3D01/import_date%3D20250901/. Here I'm checking that both files have been written out, and that they contain the expected content.

I figured it was ultimately up to polars and athena etc to read them in together and understand them as one (if polars changes that we would have very big problems haha) - do you disagree? If so would you have that in place of the checks I have now, or in addition to?

PR #979, opened 2025-09-05
[993] Remove old step functions

garypolzin:
## Description
Trello ticket [#993](https://trello.com/c/PIualc5K/993-should-remove-step-functions-no-longer-required)

We have a replacement set of step functions we're happy with so this is removing all the old set of step functions

## Testing
- [X] Unit tests passing

- [X] Updated CHANGELOG


RoyPriceCS50:
I dont think we can remove the bulk CQC download just yet.

Achyut's flatten ratings job uses locations_api dataset.

PR #996, opened 2025-09-22
1026 - Remove old model re-training methods

garypolzin:
## Description
Trello ticket [#1026](https://trello.com/c/hlezel05/1026-remove-old-attempt-at-auto-retraining-models)

I previously wrote a load of code/tests to retrain models until coming to the realisation that it wasn't possible.
None of the code is used anywhere as a result.
Joe is pretty far on with doing it with polars code so I'm removing this now so it's a cleaner process introducing the new code.

## Testing
- [X] Unit tests passing

Haven't rerun the pipeline as none of these functions are actually used anywhere

## Checklist


AchyutRatkanthiwar (approved):
Approved

PR #1004, opened 2025-09-30
New PR template for converting to polars tasks

garypolzin:
## Description
We're going to be doing a lot of transferring to Polars so thought I'd set up a specific template with a checklist for that

Hard to check the behaviour but in theory, once this is in main, when we open a PR we'll get a dropdown of which template we want to use


RoyPriceCS50 (approved):
I've definitely forgot to mark the pyspark equivalents as moved to polars.

PR #1013, opened 2025-10-02
1004 Fix/deploy dependencies in prod

Staberinde:
## Description
Trello ticket [1004](https://trello.com/c/wBpAMI9A/1004-move-circleci-feature-branch-spin-up-to-new-non-prod-environment?filter=label%3AToby)

Above ticket introduced a bug whereby `deploy-dependencies` would always assume it was running against non-prod account.
Also fixed an issue relating to non alphanumeric characters in branch names causing issues when that name is used in terrform workspace/ECR image URI. It implements the same logic used to generate the `workspace_prefix` in terraform, but implemented through different means

## Testing
- [ ] Unit tests passing
- [ ] Successful [Step Function run](add link)

- [ ] Docstrings added/updated
- [ ] Updated CHANGELOG

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,9 @@ jobs:
             aws ecr get-login-password --region eu-west-2 | docker login -u AWS --password-stdin ${<< parameters.aws_account_id >>}.dkr.ecr.eu-west-2.amazonaws.com
             echo "Rebuilding lambdas and pushing to ECR."
             export AWS_ACCOUNT_ID=${<< parameters.aws_account_id >>}
+            # This transformation should match the workspace_prefix set in terraform/pipeline/main.tf locals block
+            export SANITISED_CIRCLE_BRANCH=$(echo ${CIRCLE_BRANCH:0:30} | tr -c '[:alnum:]\n' '-' | tr '[:upper:]' '[:lower:]')
+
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -44,6 +44,9 @@ jobs:
             aws ecr get-login-password --region eu-west-2 | docker login -u AWS --password-stdin ${<< parameters.aws_account_id >>}.dkr.ecr.eu-west-2.amazonaws.com
             echo "Rebuilding lambdas and pushing to ECR."
             export AWS_ACCOUNT_ID=${<< parameters.aws_account_id >>}
+            # This transformation should match the workspace_prefix set in terraform/pipeline/main.tf locals block
+            export SANITISED_CIRCLE_BRANCH=$(echo ${CIRCLE_BRANCH:0:30} | tr -c '[:alnum:]\n' '-' | tr '[:upper:]' '[:lower:]')
```
micahdougall, line 48: Are we confident that 30 is the limit? I ad in my mind 20 for some resources

micahdougall: As an example, I think we had a branch called `946-flatten-assessment-column` which was too long for _some_ AWS resources

micahdougall (approved):

PR #1021, opened 2025-10-06
1006 Restrict Access to Production Infrastructure

Staberinde:
## Description
Trello ticket [1006](https://trello.com/c/2W0L1Nht/1006-implement-restriction-of-production-infra-to-only-admin-cicd-user)

This adds a `DENY` policy, scoped to only infrastructure which has the `IsProduction` tag (also added in this PR). If this tag is present on any piece of infrastructure, then it will be restricted to only those users/groups with `HasProductionAccess` tag will be able to modify this infrastructure. I have already put this tag on the circleCI user in the existing prod account; unfortunately there is no way to add this tag in terraform without moving the circleCI user itself to terraform, which is out of the scope of this ticket.

#1020 should be merged and deployed before this PR is deployed, just in case this doesn't behave as expected the backout plan would be.
* Revert this PR and redeploy
* If that fails, remove the `DenyProductionWithouthasProductionAccessExclStSIamS3` policy from `madetech` and `analysts` user groups
* If that fails, try the same having assumed `Admin` role

If all that fails, Zuhal will have permissions to role this back  

This PR also contains commits from #1002

## Testing
- [ ] Unit tests passing
- [ ] Successful [Step Function run](add link)

- [ ] Docstrings added/updated
- [ ] Updated CHANGELOG


MarionJHolloway (approved):

MarionJHolloway:
Closing as this branch is stale and changes have been made in other PRs

PR #1259, opened 2026-03-24
1368 and 1476 - Recalculate filled posts and fix memory bottlenecks

mitchedmunds-MadeTech:
## Description
Trello ticket [#1368](https://trello.com/c/OGwuP16N/1368-recalculate-total-filled-posts)
Trello ticket [#1476](https://trello.com/c/GfyQtpiT/1476-fix-the-memory-bottlenecks-with-by-job-role-step-function)

Although this ticket started out as recalculating the filled posts after managerial adjustment, it pivoted to making many changes to get the job to run performantly and not hit OOM errors in the ECS. This has taken A LOT of iteration to find out the methods necessary to keep `polars` within its streaming engine and avoid the memory bottlenecks. It does now run successfully in about 12 minutes, and this is with the metadata joined on at the end.
image

The resulting advice from this exploration and implementation has been [written up on Confluence](https://skillsforcare.atlassian.net/wiki/spaces/DE/pages/1796767745/Writing+streaming-friendly+polars+code).

Here's a list of some things that were tried but discarded:

1. Increasing ephemeral storage to 100GB - this was reverted here(6eee66f14715edcc2bc5057756a066eb51427c63) because it was determined it wasn't utilised in the end, and would add extra costs to each run.
2. Switching low_memory option off (1c17a6bf7e11545936dfc161ade1d567fa6283a5) - this was switched back on in the end but I think it did run fine without. I'm not sure if the performance difference is that large in the end - may be safer to keep on.
3. Tried collecting the calculated values to a list of structs so that the final output parquet has ~4mil rows rather than ~150mil. This was reverted here (1a2b53368910b6993b541c150cb49c99b88f6470) because it slowed things down a little bit and made subsequent Athena queries more challenging. It decreased the parquet file sizes, but the size in the long form isn't large enough to worry about.
   Here's an example query in Athena for this struct:
   ```sql
   SELECT 
       id, 
       t.job_data.*
   FROM dataset_ind_cqc_07_estimated_filled_posts_by_job_role_polars
   CROSS JOIN UNNEST(by_job_role_data) AS t(job_data)
   WHERE locationid = '1-2318854734'
   ```

There were more things that were tried, I won't explain them all but feel free to dive into the commit history/messages.

You'll see some changes to the Dockerfile where a new location was created on the image to set as the Polars temp dir. This was because of a concern at one point that the `/tmp` location was mapping to RAM storage. I'm not sure if this concern holds. Here was the code that changed. `chmod 777` makes that location readable, writable and executable by any user which is very permissive but fine when constrained to this location on an ephemeral container. There probably wouldn't be any harm in reverting this back, but unfortunately it wasn't the cleanest commit to revert (a1b39ca8c09804d3b4a18da57b132347bafaa98c) and might require a manual change.

```
RUN mkdir -p /polars_scratch && chmod 777 /polars_scratch
ENV POLARS_TEMP_DIR=/polars_scratch
```

## Testing
- [ ] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn:aws:states:eu-west-2:856699698263:stateMachine:1368-recalc-filled-posts-Ind-CQC-Filled-Post-Estimates-By-Role)

To iterate quickly, less attention was paid to testing the functionality - hence some of the previously written tests might need updating with the new methods - @RoyPriceCS50 and @AchyutRatkanthiwar are stepping in to help plug these gaps. Furthermore I switched off the `main` test using `pytest.mark.skip` because it was a pain trying to keep up with this while so much was changing. It might need a bit of work to mock this out. I would suggest that a better method would be to get a minimal working dataset to run through as a smoke test rather than mocking everything - just scans and sinks. This might still be a bit of work...

- [ ] Used LazyFrame option for test data
- [ ] Unit tests added
- [ ] Docstrings added
- [ ] Added comment to pyspark functions which have been migrated with the link to the new location
- [ ] Use new file structure when saving to S3
- [ ] Updated CHANGELOG


garypolzin (commented):
@mitchedmunds-MadeTech couple of bits for you
I've put SfC for stuff for us to look at after today

MarionJHolloway:
Closing as we have made these changes in stages in other PRs.

PR #1281, opened 2026-04-08
1483 - Reformat merge script and add tests within new estimates by job role pipeline

AchyutRatkanthiwar:
## Description
Trello ticket [#1483](https://trello.com/c/nsZbbn8n/1483-write-the-merge-job-script)

Reformatted the merge script within the new Estimates By Job Role Pipeline and added tests for the same. 
This job uses scan_parquet directly from Polars library and not the generic function we have created. This is because of the extra parameter passed (low_memory=True) while scanning. to use our generic function, we need to update our scan_parquet function which is not part of this PR.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1483-merge-job-Ind-CQC-Filled-Post-Estimates-By-Role:01f5236c-1379-4e99-b767-00e02b40d131)

Step function is failing on Impute step. I think it is because the way job roles were added initially is changed now so the data is bigger. need to sort it when doing impute job.

- [x] Updated CHANGELOG


projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
```diff
@@ -109,53 +101,45 @@ def main(
...
-        )
-
-        utils.sink_to_parquet(
-            lazy_df=metadata_lf,
-            output_path=metadata_destination,
-            append=False,
-        )
+    combined_schema = transformation_columns | metadata_columns
+    full_estimates_lf = (
+        pl.scan_parquet(estimates_source, low_memory=True)
```
MarionJHolloway: Do we still need the low_memory=True set here? I seem to remember reading somewhere that this wasn't having much impact, so might be worth testing without these flags before thinking about changing our utils function

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
```diff
@@ -167,22 +151,29 @@ def join_estimates_to_ascwds(
...
     """
     join_keys = [
         IndCQC.ascwds_workplace_import_date,
         IndCQC.establishment_id,
     ]
     job_role_labels = IndCQC.main_job_role_clean_labelled
 
-    # Narrow select "id" and join keys first to improve memory performance of
-    # cross join. From this we get the full amount of rows expected.
     narrow_keys_lf = estimates_lf.select(["id"] + join_keys)
```
MarionJHolloway: Use ROW_ID instead of literal

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
```diff
@@ -109,53 +101,45 @@ def main(
...
+    )
+
+    utils.sink_to_parquet(
+        lazy_df=metadata_lf,
+        output_path=metadata_destination,
+        append=False,
+    )
 
 
 def join_estimates_to_ascwds(
```
MarionJHolloway: Move this function to utils script?

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -1892,3 +1892,61 @@ class EstimateNonResCapacityTrackerFilledPostsData:
         ("1-008", date(2026, 1, 1), CareHome.not_care_home, 0.49, 0.98, 0.98, 1.0, IndCQC.ct_non_res_all_posts), # ct_non_res_filled_post_estimate is 1.0 because of clip.
         ("1-009", date(2021, 4, 30), CareHome.not_care_home, 1.0, 2.0, 2.0, None, None), # Nulled because import date prior to 2021-5-1
     ] # fmt: skip
+
+
+@dataclass
+class JoinEstimatesCase:
```
MarionJHolloway, line 1899: I think we should agree where the test cases etc are going to live - the merge job had it's own script for these. I'm happy to move them back to test data/ schemas, but I think we should be consistent! Any preference on where these live?

MarionJHolloway (changes requested):

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
AchyutRatkanthiwar: Yea I can try it without that param. if it runs we can use our utils function as you mentioned.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
AchyutRatkanthiwar: I kept it inline as it is just used for this function. it is not a generic function so happy either ways.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
AchyutRatkanthiwar: No preference tbh. It's just that we have all our test data and schemas in one place and I thought why not use the same.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
AchyutRatkanthiwar: It works with utils.scan_parquet. made that change now

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
MarionJHolloway: That's fine, I'll move the impute ones over. I think it's easier to stick with what we've got than make a whole new file in the long run too!

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_01_merge.py
MarionJHolloway: Our styleguide says that utils functions used once go in the utils script for that stage rather than inline. I think because we've got stages within a stage at the moment, it's a bit confusing!

AchyutRatkanthiwar: yea I agree. I have moved that function to utils script now. Will run the pipeline again just in case

MarionJHolloway (approved):
Will need main merging in and a successful step function run, but I will approve now so you're not locked out :)

PR #1287, opened 2026-04-10
1493 - Estimates by job role - 04 estimates

RoyPriceCS50:
## Description
Trello ticket [1493](https://trello.com/c/fTwkjZAF)

Refactored the estimates by job role estimate job.

Without joining the columns cssr/region/etc back into the data, the job runs in 1min45s and parquet is about 4gb.

This PR has the join, and saves to year partitions. This takes 3min20s and parquet total is 5.7gb.

Opening this PR for code feedback, i'm looking at the outputs.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn:aws:states:eu-west-2:856699698263:stateMachine:1493-jr-estimates-Ind-CQC-Filled-Post-Estimates-By-Role)

See Trello ticket for output checks.

- [x] Updated CHANGELOG


projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,285 @@
...
 
-        estimated_job_role_posts_lf = estimated_job_role_posts_lf.with_columns(
-            (
-                pl.col(IndCQC.estimate_filled_posts)
-                * pl.col(IndCQC.ascwds_job_role_ratios_merged)
-            ).alias(IndCQC.estimate_filled_posts_by_job_role)
-        )
 
-        # Try to refactor the pipe(apply_manager_adjustments) so it's readable sequence of expressions without class methods.
+def calculate_estimated_filled_posts_by_job_role(lf: pl.LazyFrame) -> pl.LazyFrame:
```
MarionJHolloway: According to our style guide, function definitions should be in a utils script, not inline. I'd suggest setting up a utils script specifically for this tasks as there's a few and they aren't generalised

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,285 @@
...
+    """
+    Coalesce the imputed and rolling average job role ratio columns in
+    that order to create ascwds_job_role_ratios_merged and multiply that by
+    overall estimated filled posts.
 
-        estimated_job_role_posts_lf = estimated_job_role_posts_lf.pipe(
-            apply_manager_adjustments
-        )
+    TODO: uncomment ascwds_job_role_ratios_filtered and update this doc string.
```
MarionJHolloway: Do we have a trello ticket for this?

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,285 @@
...
+                IndCQC.ascwds_job_role_rolling_ratio,
+            ],
+            name=IndCQC.ascwds_job_role_ratios_merged,
+        ),
+    )
 
+    lf = lf.with_columns(
+        (
+            pl.col(IndCQC.estimate_filled_posts)
+            * pl.col(IndCQC.ascwds_job_role_ratios_merged)
```
MarionJHolloway: Use polars function rather than python operator?

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,285 @@
...
+        ).alias(IndCQC.estimate_filled_posts_by_job_role)
+    )
 
-def apply_manager_adjustments(
-    estimated_job_role_posts_lf: pl.LazyFrame,
+    return lf
+
+
+def count_cqc_rm() -> pl.Expr:
```
MarionJHolloway: This looks like a flag rather than a count?

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,285 @@
...
+
+    Returns:
+        pl.LazyFrame: The input LazyFrame with column
+            'difference_between_estimate_and_cqc_registered_managers'.
+    """
+    return lf.with_columns(
+        (
+            (
+                pl.col(IndCQC.estimate_filled_posts_by_job_role)
+                - pl.col(IndCQC.registered_manager_count)
```
MarionJHolloway: polars function over python operator?

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2127,3 +2127,135 @@ class ImputeJobRoleData:
...
+        AdjustManagerialRolesSubFunctionTestCases(
+            id="reg_man_filled_posts_is_zero_and_cqc_count_is_one",
+            expected_data=[
+                (0, MainJobRoleLabels.supervisor, 5.0, 1, -1.0, 0.5, 4.5),
+                (0, MainJobRoleLabels.first_line_manager, 5.0, 1, -1.0, 0.5, 4.5),
+                (0, MainJobRoleLabels.registered_manager, 0.0, 1, -1.0, None, 1.0),
+            ],
+        ),
+        AdjustManagerialRolesSubFunctionTestCases(
+            id="reg_man_filled_posts_is_one_and_cqc_count_is_zero_and_non_rm_manager_filled_posts_is_very_low",
```
MarionJHolloway: Instead of "very low" could you say "less than one"?

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2127,3 +2127,135 @@ class ImputeJobRoleData:
...
+                (0, MainJobRoleLabels.supervisor, 0.0, 0.5),
+                (0, MainJobRoleLabels.first_line_manager, 0.0, 0.5),
+                (0, MainJobRoleLabels.registered_manager, 10.0, None),
+            ],
+        ),
+    ]
+
+    distribute_rm_difference_test_cases = [
+        AdjustManagerialRolesSubFunctionTestCases(
+            id="reg_man_filled_posts_is_one_and_cqc_count_is_zero",
```
MarionJHolloway: do we care about the value of non_rm_manager_filled_posts here?

MarionJHolloway (commented):
Well done, this was a big ticket! Generally looks good. A few small comments to action or discuss.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
RoyPriceCS50: There's a backlog ticket to do this properly in the future, but at the moment regardless of how many reg mans there are at a location, we just give 1 when reg man name list is 1 or more.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
RoyPriceCS50: Updated the strings.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
RoyPriceCS50: Updated the strings.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
RoyPriceCS50: Good point

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2127,3 +2127,135 @@ class ImputeJobRoleData:
...
+            id="non_rm_manager_filled_post_sum_is_zero",
+            expected_data=[
+                (0, MainJobRoleLabels.supervisor, 0.0, 0.5),
+                (0, MainJobRoleLabels.first_line_manager, 0.0, 0.5),
+                (0, MainJobRoleLabels.registered_manager, 10.0, None),
+            ],
+        ),
+    ]
+
+    distribute_rm_difference_test_cases = [
```
garypolzin: partly to help future proof this from when we move away from the simple 1/0 CQC RM logic, I think these tests would be better written to cater for business logic as opposed to specific 1 and 0 logic

I think the scenarios/test names we need to cover all bases are
- overrides_rm_posts_with_cqc_count
- redistributes_rm_difference_across_other_manager_roles
- does_not_reduce_other_manager_roles_below_zero
- allows_total_manager_posts_mismatch_when_cqc_override_exceeds_available_managerial_posts
- leaves_values_unchanged_when_prediction_matches_cqc

the data itself can be all sorts of figures, don't need to restrict them to 1 and 0, it's the behaviour which matters

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
RoyPriceCS50: I've made one in backlog. There's a spike ticket to think about how to create them, thinking long term when we have other characteristic estimates, we're all looking at the job role data to find issues now, then there'll be a ticket to apply the filtering,

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
RoyPriceCS50: I've moved them all to a utils script.

garypolzin (commented):
Left one comment on tests but I think a good rule in general is to try name the tests (and relevant test data) more in line with the business rules and edge cases you are testing, as opposed to more literally describing what data is in the test

the current behaviour is 1/0 for a lot of this, but the function doesn't know that and the test data should behave in the same way no matter what figures we use

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/estimate_utils.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Returns:
+        pl.Expr: Expression that produces register manager count.
+    """
+    return (
+        pl.col(IndCQC.registered_manager_names)
+        .list.len()
+        .clip(upper_bound=1)
+        .fill_null(0)
+    )
```
garypolzin: I agree with Marion on this, the function name doesn't really match the behaviour

when we do update the function it will need a lot more work than a tweak of this current function (we know we'll need to do fuzzy matching on names etc)

easiest change is just to rename to something like `has_rm_in_cqc_flag` or something along those lines and remove the TODO

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2127,3 +2127,135 @@ class ImputeJobRoleData:
...
+    ]
+    expected_adjust_managerial_roles_rows = [
+        (0, MainJobRoleLabels.care_worker, 10.0, 10.0),
+        (0, MainJobRoleLabels.supervisor, 20.0, 19.0),
+        (0, MainJobRoleLabels.registered_manager, 0.0, 1.0),
+    ]
+
+    calculate_reg_man_difference_test_cases = [
+        AdjustManagerialRolesSubFunctionTestCases(
+            id="estimated_rm_is_zero_and_cqc_count_rm_is_one",
```
garypolzin: same with these ones, focus on the function logic which has 3 or 4 meaningful behaviours

- calculate the RM difference (RMs - can be positive, negative and estimates can be decimals - can probably do that in two test cases (one decimal filled post higher than cqc, one lower)) - `test_calculates_difference_from_registered_manager_row`
- potentially could group with above, but that a non-RM doesn't give an output, should only be RMs

- broadcast it across the group (is the RM value copied to all the other job roles) - `test_copies_difference_to_all_rows_in_group`

- keep groups isolated (partition test) - `test_calculates_difference_per_group`

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2127,3 +2127,135 @@ class ImputeJobRoleData:
...
+        AdjustManagerialRolesSubFunctionTestCases(
+            id="estimated_rm_is_one_and_cqc_count_rm_is_zero",
+            expected_data=[
+                (0, MainJobRoleLabels.supervisor, 20.0, 0.0, 1.0),
+                (0, MainJobRoleLabels.registered_manager, 1.0, 0.0, 1.0),
+            ],
+        ),
+    ]
+
+    calculate_non_rm_managerial_distribution_test_cases = [
```
garypolzin, line 2204: tests here should probably be

`calculates_non_rm_managerial_proportions_from_estimated_posts` (your first test)

`distributes_evenly_when_non_rm_managerial_total_is_zero` (your second test)

possibly a third `calculates_non_rm_managerial_proportions_per_group` to test the `.over(ROW_ID)` logic

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/estimate_utils.py
```diff
@@ -0,0 +1,235 @@
...
+ROW_ID: Final[str] = "id"
+
+
+def calculate_estimated_filled_posts_by_job_role(lf: pl.LazyFrame) -> pl.LazyFrame:
+    """
+    Coalesce the imputed and rolling average job role ratio columns in
+    that order to create ascwds_job_role_ratios_merged and multiply that by
+    overall estimated filled posts.
+
+    TODO: uncomment ascwds_job_role_ratios_filtered and update this doc string.
```
garypolzin: can remove this TODO - we'll know to do when we get to that point

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/estimate_utils.py
```diff
@@ -0,0 +1,235 @@
...
+        lf (pl.LazyFrame): The input LazyFrame.
+
+    Returns:
+        pl.LazyFrame: The input LazyFrame with additional columns
+            ascwds_job_role_ratios_merged and estimate_filled_posts_by_job_role
+    """
+    lf = lf.with_columns(
+        utils.coalesce_with_source_labels(
+            cols=[
+                # IndCQC.ascwds_job_role_ratios_filtered,
```
garypolzin: should `IndCQC.ascwds_job_role_ratios` be in here instead?
that will help us with DQ if we know which values were submissions and which were imputed either side of submissions

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/estimate_utils.py
```diff
@@ -0,0 +1,235 @@
...
+            managerial but not a registered manager.
+
+    Returns:
+        pl.LazyFrame: The input LazyFrame with column
+            'estimate_filled_posts_by_job_role_manager_adjusted'.
+    """
+    redistribution_expr = pl.col(IndCQC.estimate_filled_posts_by_job_role).add(
+        pl.col(IndCQC.difference_between_estimate_and_cqc_registered_managers).mul(
+            pl.col(
+                IndCQC.proportion_of_non_rm_managerial_estimated_filled_posts_by_role
```
garypolzin, line 218: don't necessarily have to change this but I tend to find readability is easier when names are shorter

`difference_between_estimate_and_cqc_registered_managers` -> `rm_diff`

`proportion_of_non_rm_managerial_estimated_filled_posts_by_role` -> `non_rm_proportion` or `non_rm_managerial_proportion`

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,65 @@
...
 
 # Set streaming chunk size for memory management - each thread (per CPU core) will load
 # in a chunk of this size.
 pl.Config.set_streaming_chunk_size(50000)
 
+# Create a list of non registered manager managerial job roles.
+manager_roles = AscwdsWorkerValueLabelsJobGroup.manager_roles()
+non_rm_manager_roles = [
+    role for role in manager_roles if role != MainJobRoleLabels.registered_manager
+]
```
garypolzin: fine as it is but there's an option without the temporary variable (I can't see that it's used elsewhere but ignore if it is)

```
non_rm_manager_roles = [
    role
    for role in AscwdsWorkerValueLabelsJobGroup.manager_roles()
    if role != MainJobRoleLabels.registered_manager
]
```

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/_04_estimate.py
```diff
@@ -1,125 +1,65 @@
...
-    estimated_job_role_posts_lf = estimated_job_role_posts_lf.join(
-        stats_lf,
-        on=pct_share_groups,
-        how="left",
+    lf = eUtils.adjust_managerial_roles(lf, non_rm_manager_roles)
+
+    lf = lf.with_columns(
+        pl.col(IndCQC.cqc_location_import_date)
+        .dt.year()
+        .cast(pl.String)
```
garypolzin: I don't think the cast is required here - probably easier to use year if it's in integer format than string, and partitions can take intergers

garypolzin (approved):
couple of comments but all minor/optional so happy to approve now

PR #1296, opened 2026-04-20
1541 Added ROW_ID and EXPANDED_ID to ind cqc columns

RoyPriceCS50:
## Description
Trello ticket [1541](https://trello.com/c/ZQmRGDeQ)

Added ROW_ID and EXPANDED_ID to the ind cqc columns class as Final types and updated references to them throughout the estimates by job role jobs.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn:aws:states:eu-west-2:856699698263:stateMachine:1541-move-rowid-Ind-CQC-Filled-Post-Estimates-By-Role)

Job role estimates have not been run in main yet, so checked against branch 1493. Not exactly the same, but got less than 0.1% difference when looking at sum of estimate_filled_posts_by_job_role_manager_adjusted. See Trello ticket.

- [x] Updated CHANGELOG


garypolzin (commented):
Both the ID's can be lowercase now as they are referenced in Ind CQC col names, so keep them consistent - probably used capitals before as they were referenced within the job itself

Also a merge conflict with CHANGELOG to sort out (make sure it goes in the unreleased section following latest git tag)

utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -311,6 +313,7 @@ class IndCqcColumns:
         "residuals_estimate_filled_posts_non_res_pir"
     )
     rmse: str = "rmse"
+    row_id: Final[str] = "id_per_locationid_import_date"
```
MarionJHolloway: There's no real need to have Final here. It marks the variable so it's value can't be change at a later date. Really we should either do this with all the column names or none. This isn't a special case. I would suggest removing it for now and considering whether we want another ticket to mark all column names as final

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 
 
 ### Changed
-
+- Added ROW_ID and EXPANDED_ID to the ind cqc columns class and updated references to the objects.
```
MarionJHolloway: Remove capitals here

utils/column_names/ind_cqc_pipeline_columns.py
garypolzin: @RoyPriceCS50 @MarionJHolloway I'm guessing he used Final as 'row_id' is so generic that it could accidentally be used elsewhere? we should avoid that anyway by adding it to the cols list

but are we happy with called it 'row_id' or should we actually call the variable 'id_per_locationid_import_date' too - I didn't appreciate the name and reference name didn't match before

MarionJHolloway: But it's not the column that's in the variable, just the column name. It's not something we do in our code because we have all the column names extracted in their own file. 

I'd prefer the more descriptive name.

garypolzin: yea I prefer it when the names match too

utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -187,6 +187,9 @@ class IndCqcColumns:
     estimate_source: str = "estimate_source"
     estimate_value: str = "estimate_value"
     expected_filled_posts: str = "expected_filled_posts"
+    id_per_locationid_import_date_job_role: str = (
+        "id_per_locationid_import_date_job_role"
+    )
```
garypolzin: sort alphabetically

utils/column_names/ind_cqc_pipeline_columns.py
```diff
@@ -311,6 +314,7 @@ class IndCqcColumns:
         "residuals_estimate_filled_posts_non_res_pir"
     )
     rmse: str = "rmse"
+    id_per_locationid_import_date: str = "id_per_locationid_import_date"
```
garypolzin: sort alphabetically

garypolzin (approved):
Happy to approve but merge conflict and alphabetical ordering of pipeline cols to sort please

PR #1299, opened 2026-04-21
1502 - Convert extrapolation to polars

MarionJHolloway:
## Description
Trello ticket [#1502](https://trello.com/c/vQfMjfK3/1502-ind-cqc-model-refactor-extrapolation)

utils/models/extrapolation.py
tests/utils/models/test_extrapolation.py

## Testing
- [x] Unit tests passing

Can't test in branch as this is a utils function not the whole job

- [x] Unit tests added
- [x] Docstrings added
- [X] Code reviewed by AI
- [x] Updated CHANGELOG


projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2205,3 +2205,384 @@ class EstimateFilledPostsByJobRoleEstimateUtilsData:
...
+            id="when_more_than_one_location_needs_extrapolating",
+            data=[
+                ("1-001", date(2026, 1, 1), None, 10.0, date(2026, 2, 1), date(2026, 2, 1), 5.0),
+                ("1-001", date(2026, 2, 1), 10.0, 20.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-001", date(2026, 3, 1), None, 30.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-001", date(2026, 4, 1), None, 100.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-002", date(2026, 1, 1), 20.0, 100.0, date(2026, 1, 1), date(2026, 1, 1), None),
+                ("1-002", date(2026, 2, 1), None, 20.0, date(2026, 1, 1), date(2026, 1, 1), None),
+            ],
+        ),
```
RoyPriceCS50, line 2529: Location 1-002 doesn't "need" extrapolating if the since only earliest import has a value.

Better as sets of three rows per id, with different outcomes, like the ratio change at location 2 is different to location 1, or the same... just showing it does stuff at each location.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2205,3 +2205,384 @@ class EstimateFilledPostsByJobRoleEstimateUtilsData:
...
+            id="when_more_than_one_location_needs_extrapolating",
+            data=[
+                ("1-001", date(2026, 1, 1), None, 10.0, date(2026, 2, 1), date(2026, 2, 1), 0.0),
+                ("1-001", date(2026, 2, 1), 10.0, 20.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-001", date(2026, 3, 1), None, 30.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-001", date(2026, 4, 1), None, 100.0, date(2026, 2, 1), date(2026, 2, 1), None),
+                ("1-002", date(2026, 1, 1), 20.0, 100.0, date(2026, 1, 1), date(2026, 1, 1), None),
+                ("1-002", date(2026, 2, 1), None, 20.0, date(2026, 1, 1), date(2026, 1, 1), None),
+            ],
+        ),
```
RoyPriceCS50, line 2460: Location 1-002 doesn't "need" extrapolating if the since only earliest import has a value.

Better as sets of three rows per id, with different outcomes, like the nominal change at location 2 is 0 so the earliest import gets same as middle import.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2205,3 +2205,384 @@ class EstimateFilledPostsByJobRoleEstimateUtilsData:
...
+            id="when_earlier_and_later_data_points_are_missing",
+            data=[
+                ("1-002", date(2026, 1, 1), None, date(2026, 2, 1), date(2026, 2, 1), 3.0, 2.0, 2.0),
+                ("1-002", date(2026, 2, 1), 4.0, date(2026, 2, 1), date(2026, 2, 1), 4.0, 4.0, None),
+                ("1-002", date(2026, 3, 1), None, date(2026, 2, 1), date(2026, 2, 1), 5.0, 6.0, 5.0),
+                ("1-002", date(2026, 4, 1), None, date(2026, 2, 1), date(2026, 2, 1), 7.0, 8.0, 7.0),
+            ],
+        ),
+        ExtrapolationTestCase(
+            id="when_more_than_one_location_needs_extrapolating",#TODO
```
RoyPriceCS50: remove TODO

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2205,3 +2205,384 @@ class EstimateFilledPostsByJobRoleEstimateUtilsData:
...
+        ("1-001", date(2026, 1, 1), 15.0, date(2026, 1, 1), date(2026, 3, 1)),
+        ("1-001", date(2026, 2, 1), None, date(2026, 1, 1), date(2026, 3, 1)),
+        ("1-001", date(2026, 3, 1), 30.0, date(2026, 1, 1), date(2026, 3, 1)),
+        ("1-002", date(2026, 1, 1), None, date(2026, 2, 1), date(2026, 2, 1)),
+        ("1-002", date(2026, 2, 1), 4.0, date(2026, 2, 1), date(2026, 2, 1)),
+        ("1-002", date(2026, 3, 1), None, date(2026, 2, 1), date(2026, 2, 1)),
+        ("1-003", date(2026, 1, 1), None, None, None),
+    ]
+
+    # fmt: off
```
RoyPriceCS50: Instead of doing formatter off/on, we've started doing # fmt: skip next to the closing ].

So it's  [] # fmt: skip

No need to turn on/off/forget to switch back

projects/_03_independent_cqc/_06_estimate_filled_posts/tests/utils/models/test_extrapolation.py
```diff
@@ -206,6 +206,7 @@ def test_returned_extrapolation_forwards_values_match_expected_when_nominal(self
         self.returned_data = self.returned_nominal_df.sort(
             IndCqc.location_id, IndCqc.unix_time
         ).collect()
+        self.returned_nominal_df.show()
```
RoyPriceCS50: Remove.

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+        .then(pl.col(IndCqc.extrapolation_forwards))
+        .when(
+            pl.col(IndCqc.cqc_location_import_date)
+            < pl.col(IndCqc.first_submission_time)
+        )
+        .then(pl.col(IndCqc.extrapolation_backwards))
+    )
+    lf = lf.with_columns(combine_extrapolation_expr.alias(IndCqc.extrapolation_model))
+    return lf
```
RoyPriceCS50: return lf.with_columns()

(getting picky)

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+
+    if extrapolation_method == "ratio":
+        lf = lf.with_columns(ratio_expr.alias(IndCqc.extrapolation_backwards))
+
+    elif extrapolation_method == "nominal":
+        lf = lf.with_columns(nominal_expr.alias(IndCqc.extrapolation_backwards))
+
+    else:
+        raise ValueError("Error: method must be either 'ratio' or 'nominal'.")
+    lf.show()
```
RoyPriceCS50: Remove

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+    lf = lf.drop("rank", IndCqc.first_non_null_value, IndCqc.first_model_value)
+
+    return lf
```
RoyPriceCS50: return lf.drop()

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+
+    elif extrapolation_method == "nominal":
+        lf = lf.with_columns(nominal_expr.alias(IndCqc.extrapolation_forwards))
+
+    else:
+        raise ValueError("Error: method must be either 'ratio' or 'nominal'.")
+
+    lf = lf.drop("rank", IndCqc.previous_non_null_value, IndCqc.previous_model_value)
+
+    return lf
```
RoyPriceCS50: return lf.drop()

garypolzin (commented):
now it's in polars code I think there's some steps we can do to simplify how it's working - there's a fair amount of repetition in 'extrapolation_forwards' and 'extrapolation_backwards'

I think we can get all the aggregated figures in one go - I'll put this in written form to save me trying to code it up! Let me know if it doesn't make sense. I also haven't tried it, just looks possible now the code is neater

**calculate aggregated cols**
agg_lf =
filter where column_with_null_values is not null
group by locationid
aggregate
- first submission date (min import date)
- last submission date (max import date)
- first non null value (pl.col(null_col).sort_by(import_date).first().alias(...))
- first model (as above but model col)
- same again but 'last'

**one join to get all those in**

**define expressions**
forward_ratio
backward_ratio
forward_nominal 
backward_nominal 

then one if/elif/else (like the ones we have but only need one now)
if method = ratio
  forward_expr = forward_ratio
  backward_expr = backward_ratio
elif nominal
  forward_expr = forward_nominal
  backward_expr = backward_nominal
else -> raise error

**apply the expressions**
combine_extrapolations is pretty similar to now, but instead of the columns 'IndCqc.extrapolation_forwards' and 'IndCqc.extrapolation_backwards' existing, you can call the expressions. I think it helps polars to include `.otherwise(none)` after the two when calls too but not fully sure on that one

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+    ratio_expr = (
+        pl.col(IndCqc.previous_non_null_value)
+        .mul(pl.col(model_to_extrapolate_from))
+        .truediv(pl.col(IndCqc.previous_model_value))
+    )
+    nominal_expr = (
+        pl.col(IndCqc.previous_non_null_value)
+        .add(pl.col(model_to_extrapolate_from))
+        .sub(pl.col(IndCqc.previous_model_value))
+    )
```
RoyPriceCS50: Question rather than feedback:

In the backward extrapolation, you've got a when/then so it only goes backwards.
But can see in the test data that the "forwards extrapolation" goes backwards too.

Why go backwards as well?

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+    lf = extrapolation_backwards(
+        lf,
+        column_with_null_values,
+        model_to_extrapolate_from,
+        extrapolation_method,
+    )
+    lf = combine_extrapolation(lf)
+    lf = lf.drop(IndCqc.first_submission_time, IndCqc.final_submission_time)
+
+    return lf
```
RoyPriceCS50: return lf.drop()

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+def extrapolation_forwards(
+    lf: pl.LazyFrame,
+    column_with_null_values: str,
+    model_to_extrapolate_from: str,
+    extrapolation_method: str,
+) -> pl.LazyFrame:
+    """
+    Calculates the forward extrapolation and adds it as a new column 'extrapolation_forwards'.
+
+    This function fills null values occurring after the last known value by extrapolating forwards.
```
RoyPriceCS50: Not true?  Goes backwards as well

RoyPriceCS50 (commented):
Mostly commented on style, but also question on forwards extrapolation.

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
AchyutRatkanthiwar: remove show

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+
+    for date_expr in dates_needed:
+        date_expr_lf = (
+            lf.drop_nulls(column_with_null_values)
+            .sort([IndCqc.location_id, IndCqc.cqc_location_import_date])
+            .group_by(IndCqc.location_id)
+            .agg(date_expr)
+        )
+
+        lf = lf.join(
```
AchyutRatkanthiwar: instead of join why not use over? min and max should be easy to get using over. something like this:
```
lf = lf.with_columns([
    pl.col(IndCqc.cqc_location_import_date)
      .filter(pl.col(column_with_null_values).is_not_null())
      .min()
      .over(IndCqc.location_id)
      .alias(IndCqc.first_submission_time),

    pl.col(IndCqc.cqc_location_import_date)
      .filter(pl.col(column_with_null_values).is_not_null())
      .max()
      .over(IndCqc.location_id)
      .alias(IndCqc.final_submission_time),
])

File: projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py [hunk 14]
```diff
@@ -0,0 +1,323 @@
...
+        pl.LazyFrame: A LazyFrame with a new column containing forward extrapolated values.
+
+    Raises:
+        ValueError: If chosen extrapolation_method does not match 'nominal' or 'ratio'.
+    """
+    ### ADD RANK COLUMN ###
+    lf = lf.with_columns(
+        pl.when(pl.col(column_with_null_values).is_not_null())
+        .then(pl.col(IndCqc.cqc_location_import_date).rank().over(IndCqc.location_id))
+        .otherwise(None)
```
AchyutRatkanthiwar: Can remove .otherwise(None)

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+    )
+
+    ### POPULATE previous_non_null_value ###
+    previous_non_null_lf = (
+        lf.sort([IndCqc.location_id, IndCqc.cqc_location_import_date])
+        .group_by(IndCqc.location_id)
+        .agg(
+            pl.col(column_with_null_values)
+            .min_by("rank")
+            .alias(IndCqc.previous_non_null_value)
```
AchyutRatkanthiwar: In my review I have calculated previous non null value using forward_fill. ahh but your next step needs the rank so need to do it this way

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+        how="left",
+    )
+
+    ### CALCULATE EXTRAPOLATION ###
+
+    ratio_expr = pl.when(
+        pl.col(IndCqc.cqc_location_import_date) < pl.col(IndCqc.first_submission_time)
+    ).then(
+        pl.col(IndCqc.first_non_null_value)
+        .mul(pl.col(model_to_extrapolate_from))
```
AchyutRatkanthiwar: Did we say we use python operators instead?

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+    ### ADD RANK COLUMN ###
+    lf = lf.with_columns(
+        pl.when(pl.col(column_with_null_values).is_not_null())
+        .then(pl.col(IndCqc.cqc_location_import_date).rank().over(IndCqc.location_id))
+        .otherwise(None)
+        .alias("rank")
+    )
+
+    ### POPULATE first_non_null_value ###
+    first_non_null_lf = (
```
AchyutRatkanthiwar: first_non_null_lf  and previous_non_null_lf  in exptrapolation_forwards is same. better to do once and use in 2 places?

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+    )
+
+    lf = lf.join(
+        first_model_lf,
+        on=IndCqc.location_id,
+        how="left",
+    )
+
+    ### CALCULATE EXTRAPOLATION ###
```
AchyutRatkanthiwar: Calculate extrapolation can be one function as it is doing same in 2 places?

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
MarionJHolloway: The when/then was only in the backwards extrapolation in the pyspark code. The forwards extrapolation needed some significant reworking and the backwards values are never touched because of the way they are selected in the "combine extrapolation" function, so I just left them as I was working through the refactor. I think you're right though, the when/then will probably speed up the code as well (fewer claculations), so will add it in the forwards extrapolation too.

garypolzin: in terms of outputs it only goes forwards because 'previous_non_null_value' is null before the first submitted one, so the output of the calculation should be null too

forwards and backwards should be slightly differently though
forwards - extrapolate forwards from every known value (interpolation uses this)
backwards - only need to do before the first submitted

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
garypolzin: should only be forwards :)
should be null before first submission, potentially to do with the quirk where polars .mul() etc returning zero instead of null when there's a null in the calculation?

MarionJHolloway: Changed the function, rather than the docstring.

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
MarionJHolloway: previous_non_null_lf has min_by("rank"), first_non_null_lf has max_by("rank")

There could still be some deduplication of code here, but it would involve re-jigging a lot of the tests too, and restructuring the functions in model_extrapolation. Do you think it's worth the work? I'm not sure.

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
MarionJHolloway: It's not doing the same, one is populating forwards, one is backwards. Again, I could do a big restructure here. Do we think it's worth it?

AchyutRatkanthiwar: Nah I think it is fine. If we create a function it will still need calling twice with diff parameters. it is already readable tbh so can keep it same for now. The restructure might take a lot of time just to make it more abstract.

projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/utils/models/extrapolation.py
AchyutRatkanthiwar: yes what I meant was defining it as a function but restructuring seems much bigger task so I a happy with this 😄

projects/_03_independent_cqc/utils/imputation/extrapolation.py
```diff
@@ -0,0 +1,221 @@
...
+                pl.col(IMPORT_DATE).min().alias(TEMP.first_submission_time),
+                pl.col(IMPORT_DATE).max().alias(TEMP.final_submission_time),
+                pl.col(value_col).sort_by(IMPORT_DATE).first().alias(TEMP.first_value),
+                pl.col(model_col).sort_by(IMPORT_DATE).first().alias(TEMP.first_model),
+            ]
+        )
+    )
+
+
+def get_previous_value(col: str) -> pl.Expr:
```
MarionJHolloway, line 150: You could add this to the class as it's defining not calculating anything.

garypolzin: even though it's an expression, the ExtrapolationExpressions class is focused on defining extrapolation calculations and only depends on the model column, whereas get_previous_value is a generic data transformation that operates on multiple columns

I think keeping it separate avoids adding unnecessary complexity and keeps the class focused on the single task

projects/_03_independent_cqc/utils/imputation/extrapolation.py
```diff
@@ -0,0 +1,221 @@
...
+    lf: pl.LazyFrame,
+    column_with_null_values: str,
+    model_to_extrapolate_from: str,
+    extrapolation_method: str,
+) -> pl.LazyFrame:
+    """
+    Extrapolate missing values in a column using a reference model.
+
+    This function performs forward and backward extrapolation for values that
+    fall outside the observed range of `column_with_null_values`, using a
```
MarionJHolloway: "fall outside the observed range" sounds like it's creating values that are higher or lower. Maybe "fall outside the submitted dates" or similar?

projects/_03_independent_cqc/utils/imputation/extrapolation.py
```diff
@@ -0,0 +1,221 @@
...
+    specified model column and extrapolation method.
+
+    The process consists of:
+    - Computing per-group (location_id) aggregates such as first/last submission
+      dates and first observed values
```
MarionJHolloway: Not all observed values, some are estimated values

projects/_03_independent_cqc/utils/imputation/extrapolation.py
```diff
@@ -0,0 +1,221 @@
...
+    - Deriving previous observed values within each group
+    - Applying either ratio-based or nominal extrapolation logic
+    - Producing two output columns:
+        - `extrapolation_forwards`: values extrapolated after the first observation
+        - `extrapolation_model`: combined forward and backward extrapolated values
+
+    Extrapolation is applied as follows:
+    - Forward extrapolation: for dates after the last observed value
```
MarionJHolloway: I'm sure I added something to this docstring or the tests to cover why extrapolation forwards is needed for interpolation and how it should behave over multiple gaps for interpolation. Was there a reason you removed it?

I think it's important to have documented somewhere, as you said yourself, extrapolation forwards is doing more than immediately meets the eye and we wants it to be clear to anyone working on this in the future.

projects/_03_independent_cqc/utils/imputation/extrapolation.py
MarionJHolloway: In that case I would rename the class to ExtrapolationCalculationExpressions just to be clear

projects/_03_independent_cqc/utils/tests/imputation/test_extrapolation.py
```diff
@@ -0,0 +1,205 @@
...
+                ExtrapCol.first_model: [20.0],
+                self.MODEL: [10.0],
+            }
+        )
+
+        expr = job.ExtrapolationExpressions(self.MODEL)
+
+        result = lf.select(expr.backward_ratio.alias(self.OUTPUT)).collect()
+
+        assert result[self.OUTPUT][0] == 5.0
```
MarionJHolloway: Add comment with formula for consitancy?

MarionJHolloway (commented):
A few comments from me, mostly related to documenting the complexities of this for the next person who works on it :)

projects/_03_independent_cqc/utils/tests/imputation/test_extrapolation.py
```diff
@@ -0,0 +1,205 @@
...
+)
+from projects._03_independent_cqc.unittest_data.polars_ind_cqc_test_file_schemas import (
+    ModelExtrapolation as Schemas,
+)
+from utils.column_names.ind_cqc_pipeline_columns import (
+    ExtrapolationColumns as ExtrapCol,
+)
+from utils.column_names.ind_cqc_pipeline_columns import IndCqcColumns as IndCQC
+
+PATCH_PATH = "projects._03_independent_cqc.utils.utils.imputation.extrapolation"
```
garypolzin: drop - don't need this

projects/_03_independent_cqc/utils/imputation/extrapolation.py
garypolzin: I've copied the text from the pyspark docstring instead, it was worded better

projects/_03_independent_cqc/utils/imputation/extrapolation.py
garypolzin: it's the observed/submitted values which are being extrapolated because they are the ones which contain nulls - the estimated values are used to extrapolate against

projects/_03_independent_cqc/utils/imputation/extrapolation.py
garypolzin: ah yea, that wasn't intentional so I've added back in

garypolzin (commented):
@MarionJHolloway - PR comments addressed :)

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2344,3 +2343,269 @@ class InterpolationData:
...
+            id="when_model_extrapolation_only_applies_after_last_submission",
+            data=[
+                ("1-001", date(2026, 1, 1), 10.0, 20.0, None, None),
+                ("1-001", date(2026, 2, 1), None, 30.0, 20.0, None),
+                ("1-001", date(2026, 3, 1), 50.0, 40.0, 30.0, None),
+                ("1-001", date(2026, 4, 1), None, 50.0, 60.0, 60.0),
+            ],
+        ),
+        ExtrapolationTestCase(
+            id="when_multiple_gaps_exist_across_full_time_series",
```
MarionJHolloway, line 2418: Could you also add comments here to explain how the forwards extrapolation should work? I've added examples of what might help make this clearer

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2344,3 +2343,269 @@ class InterpolationData:
...
+            data=[
+                ("1-001", date(2026, 1, 1), None, 10.0, None, 0.0),
+                ("1-001", date(2026, 2, 1), 10.0, 20.0, None, None),
+                ("1-001", date(2026, 3, 1), None, 30.0, 20.0, None),
+                ("1-001", date(2026, 4, 1), 20.0, 80.0, 70.0, None),
```
MarionJHolloway: "# Because of the lagging of values, this row is a continuation of the forwards extrapolation from the 2026-02-01 datapoint"

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2344,3 +2343,269 @@ class InterpolationData:
...
+                ("1-001", date(2026, 5, 1), None, 100.0, 40.0, 40.0),
```
MarionJHolloway: "# This row is forwards extrapolation from the 2026-04-01 datapoint"

MarionJHolloway (commented):
I think it just needs a bit more on the behaviour of the forwards extrapolation. I've added suggested comments, see what you think.

MarionJHolloway (commented):
All good, happy to approve

garypolzin (approved):

PR #1302, opened 2026-04-23
1560 Change cqc api so start date is 30 days earlier

RoyPriceCS50:
## Description
Trello ticket [1560](https://trello.com/c/AfsrnIJN)

Changed the CQC locations and providers delta download jobs like this:
- Replaced main argument start_timestamp with previous_days_to_capture, default to 30 days
- Changed the order of main arguments
- Both functions now get data from cqc changes api over the 30 days prior to end_timestamp
- Even though the functions default to 30 days, the number of days is actually passed in as a magic number in the step function at the container overrides.

Ran the step function passing in the end_timestamp as "2026-04-15T02:00:00.00Z" and number of captures days as 15.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn:aws:states:eu-west-2:856699698263:stateMachine:1560-api-strt-lag-Ingest-CQC-API-Delta)

I got 1,768 unique locationid's in my branch run and same number in main delta dataset imports 08/04/2026 and 15/04/2026

There's no unit test to check the capture period, since it's not returned or a column in the delta dataframe, but you can see it in the [logs](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/%2Fecs%2F1560-api-strt-lag-cqc-api-task-logs/log-events/ecs%2Fcqc-api-container%2F4973b7306361460b96bb99435b7f335a) (4th line down)

- [x] Updated CHANGELOG

### DIFF
--- a/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
+++ b/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
@@ -26,7 +26,7 @@ CQC_ORG_TYPE = "location"
 
 def main(
     destination: str,
-    end_timestamp: str = None,
+    end_timestamp: str,
     previous_days_to_capture: int = cqc.days_to_rollback_start_timestamp,
 ) -> None:
     """
--- a/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
+++ b/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
@@ -27,7 +27,7 @@ CQC_ORG_TYPE = "location"
 def main(
     destination: str,
     end_timestamp: str,
-    previous_days_to_capture: int = cqc.days_to_rollback_start_timestamp,
+    previous_days_to_capture: str = cqc.days_to_rollback_start_timestamp,
 ) -> None:
     """
     This function performs the following steps:
@@ -46,7 +46,8 @@ def main(
             Parquet file will be written.
         end_timestamp (str): The ISO 8601 formatted string representing the
             end of the data retrieval period (e.g., '2023-01-31T23:59:59Z').
-        previous_days_to_capture (int): Number of days before end_timestamp (default 15).
+        previous_days_to_capture (str): Number of days before end_timestamp as a
+            string (default "15").
 
     Return:
         None
@@ -61,7 +62,7 @@ def main(
         destination = destination if destination[-1] == "/" else f"{destination}/"
 
         end_dt = dt.fromisoformat(end_timestamp.replace("Z", ""))
-        start_dt = end_dt - timedelta(days=previous_days_to_capture)
+        start_dt = end_dt - timedelta(days=float(previous_days_to_capture))
 
         print(f'Getting SecretID "{SECRET_ID}"')
         secret = get_secret(secret_name=SECRET_ID, region_name=AWS_REGION)
@@ -119,6 +120,10 @@ if __name__ == "__main__":
         (
             "--end_timestamp",
             "End timestamp for location changes",
+        ),
+        (
+            "--previous_days_to_capture",
+            "Number of days prior to end timestamp as a string",
         ),
     )
     print(f"Running cqc locations delta api download job")
--- a/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
+++ b/projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
@@ -32,8 +32,8 @@ CQC_ORG_TYPE = "location"
 def main(destination: str, start_timestamp: str, end_timestamp: str) -> None:
     """
     This function performs the following steps:
-    1. Validates the provided start and end timestamps.
-    2. Subtracts 15 days from input start_timestamp.
+    1. Subtracts 15 days from input start_timestamp.
+    2. Validates the provided start and end timestamps.
     3. Retrieves the CQC API subscription key from AWS Secrets Manager.
     4. Calls the CQC API to fetch updated location objects within the specified
        time range.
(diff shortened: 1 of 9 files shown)


garypolzin (commented):
I'm not sure I like the concept of a visible start date which isn't the start date but I think we can refactor slightly which pretty much uses the same code you've added anyway

instead of `start_timestamp` we could have `number_of_days` (or `lookback_days` or `backfill_days` or whatever!)

could default that to 15 start with, so every week we'll get the previous 2 weeks of changes. but it's easy to reduce/extend if we need to (permanently or adhoc)

terraform/pipeline/step-functions/dynamic/Ingest-CQC-API-Delta.json
```diff
@@ -21,8 +21,6 @@
                 "script": "cqc_providers_1_delta_api_download.py",
                 "dest_arg": "--destination_prefix",
                 "dest_value": "${dataset_bucket_uri}",
-                "start_arg": "--start_timestamp",
-                "start_value.$": "$.Parameter.Value",
```
garypolzin, line 25: for ease of amending on the fly, can you do this? it will also mean we can see which value was used on historic runs of the step function

                "days_to_capture_arg": "--previous_days_to_capture",
                "days_to_capture_value": "15" (or just 15 without quotes?),

RoyPriceCS50: The argument values have to be strings. I've changed this step function so it has pass state to get the value "15", then that gets passed into each job run. Had to change each of the ingest functions so string "15" gets converted to a float.

projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
```diff
@@ -17,21 +18,20 @@
...
 AWS_REGION = os.environ.get("AWS_REGION", "")
 CQC_OBJECT_TYPE = "locations"
 CQC_ORG_TYPE = "location"
 
 
-def main(destination: str, start_timestamp: str, end_timestamp: str) -> None:
+def main(
+    destination: str,
+    end_timestamp: str,
+    previous_days_to_capture: str = cqc.days_to_rollback_start_timestamp,
```
garypolzin: although the stepfunction requires a string, we want the function to take an int

previous_days_to_capture: int = cqc.days_to_rollback_start_timestamp,

projects/_01_ingest/cqc_api/utils/cqc_api.py
```diff
@@ -14,6 +14,8 @@
 CQC_API_BASE_URL = "https://api.service.cqc.org.uk"
 USER_AGENT = "SkillsForCare"
 
+days_to_rollback_start_timestamp = "15"
```
garypolzin: change to int

projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
```diff
@@ -138,4 +136,4 @@ def main(destination: str, start_timestamp: str, end_timestamp: str) -> None:
         date=date_today,
         version="3.1.0",
     )
-    main(destination, args.start_timestamp, args.end_timestamp)
+    main(destination, end_timestamp=args.end_timestamp)
```
garypolzin: add the arg (and convert to int) here
`previous_days_to_capture=int(args.previous_days_to_capture)`

projects/_01_ingest/cqc_api/fargate/cqc_locations_1_delta_api_download.py
```diff
@@ -62,19 +61,16 @@ def main(destination: str, start_timestamp: str, end_timestamp: str) -> None:
...
         destination = destination if destination[-1] == "/" else f"{destination}/"
 
-        start_dt = dt.fromisoformat(start_timestamp.replace("Z", ""))
         end_dt = dt.fromisoformat(end_timestamp.replace("Z", ""))
-
-        if start_dt > end_dt:
-            raise InvalidTimestampArgumentError(
-                "start_timestamp is after end_timestamp"
-            )
+        start_dt = end_dt - timedelta(days=float(previous_days_to_capture))
```
garypolzin: can remove float() as it's already an int now

terraform/pipeline/step-functions/dynamic/Ingest-CQC-API-Delta.json
```diff
@@ -49,7 +56,7 @@
                   "ContainerOverrides": [
                     {
                       "Name": "cqc-api-container",
-                      "Command.$": "States.Array($.script, $.dest_arg, $.dest_value, $.start_arg, $.start_value, $.end_arg, $.end_value)"
+                      "Command.$": "States.Array($.script, $.dest_arg, $.dest_value, $.end_arg, $.end_value, $.days_arg, $.days_value)"
```
garypolzin: we no longer need the step before this (`Read Last Providers/Locations Run`) as we don't use it, so I think you can get rid of it and change this command to

```
          "Command.$": "States.Array(
            'cqc_providers_1_delta_api_download.py',
            '--destination_prefix',
            '${dataset_bucket_uri}',
            '--end_timestamp',
            $$.State.EnteredTime,
            '--previous_days_to_capture',
            $.days_value
          )"
```

garypolzin: alternatively, get rid of the "Set number of days to capture" state and put the 15 directly in here, I'm happy either way

```
          "Command.$": "States.Array(
            'cqc_providers_1_delta_api_download.py',
            '--destination_prefix',
            '${dataset_bucket_uri}',
            '--end_timestamp',
            $$.State.EnteredTime,
            '--previous_days_to_capture',
            '15'
          )"
```

garypolzin (commented):
Added a few comments

I'm wondering if we increase from 15 to 30 - storage is cheap and missing out on data is really faffy/time consuming, so ideal scenario is to never miss data again. The impact of 30 days worth of data each time is very minimal in terms of cost and processing time

RoyPriceCS50:
> Added a few comments
> 
> I'm wondering if we increase from 15 to 30 - storage is cheap and missing out on data is really faffy/time consuming, so ideal scenario is to never miss data again. The impact of 30 days worth of data each time is very minimal in terms of cost and processing time

I've made all these changes and ran it over a 30 day period, it took 8mins30secs.

garypolzin (approved):
Looks good thanks!

PR #1304, opened 2026-04-24
1556 - Convert DPR Model Interpolation to Polars

AchyutRatkanthiwar:
## Description
Trello ticket [#1556](https://trello.com/c/9TwUdXz0/1556-dpr-convert-modelinterpolation)

Converted DPR Interpolation script to Polars and added tests for the same.

## Testing
- [x] Unit tests passing

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG


projects/_04_direct_payment_recipients/fargate/utils/models/interpolation.py
RoyPriceCS50, line 1: I think the pyspark version is overly complicated.
Prior to the estimates job, the S3 dataset is setup to have all years for every LA area. The raw data is setup like this.
And some areas have gaps between known values that need straight lining.
I think this whole job can simplified to just:

direct_payments_lf = direct_payments_lf.with_columns(
        pl.col(DP.ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF)
        .interpolate_by(DP.YEAR_AS_INTEGER)
        .over(DP.LA_AREA)
        .alias(DP.ESTIMATE_USING_INTERPOLATION)
    )

AchyutRatkanthiwar: yes after I created the PR I realised I can simplify it by loads! 
I tested this code and it works as we want it to. So, do we want it to be its own function or inline code will do given it is very small. My updated code:
```
def model_interpolation(
    direct_payments_lf: pl.LazyFrame,
) -> pl.LazyFrame:
    """
    Performs straight line interpolation of missing values for the
    estimated proportion of service users employing staff.

    Args:
        direct_payments_lf (pl.LazyFrame): Input LazyFrame with columns LA_AREA,
            YEAR_AS_INTEGER and ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF

    Returns:
        pl.LazyFrame: Original LazyFrame with an additional column
            ESTIMATE_USING_INTERPOLATION
    """
    return direct_payments_lf.with_columns(
        pl.col(DP.ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF)
        .interpolate_by(DP.YEAR_AS_INTEGER)
        .over(DP.LA_AREA)
        .alias(DP.ESTIMATE_USING_INTERPOLATION)
    )

RoyPriceCS50: I think keep it in a util for now because it gets called twice.

One other thing it the test data, can you put the area's next to each other so it's easier to look at, and have a case where only one year is populated (it doesn't get interpolated), a case where consecutive years are populated (again no interpolation).

AchyutRatkanthiwar: Done

RoyPriceCS50 (approved):
Looks good.

PR #1305, opened 2026-04-27
1551 dpr la names

RoyPriceCS50:
## Description
Trello ticket [1551](https://trello.com/c/xwEgmsCY)

Converted DPR util for correcting LA names from pyspark to polars. Done without wrapping, just called in main, and added a validation script to check LA names are correct in the output.

## Testing
- [x] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-2%3A856699698263%3AstateMachine%3A1551-dpr-la-names-Direct-Payment-Recipients?type=standard)

Validation is checking the column only has values from the contemporary cssr class.

- [x] Unit tests added
- [x] Updated CHANGELOG

### DIFF
--- a/projects/_04_direct_payment_recipients/fargate/estimate_direct_payments.py
+++ b/projects/_04_direct_payment_recipients/fargate/estimate_direct_payments.py
@@ -1,6 +1,11 @@
+import polars as pl
+
 from polars_utils import utils
 from projects._04_direct_payment_recipients.direct_payments_column_names import (
     DirectPaymentColumnNames as DP,
+)
+from projects._04_direct_payment_recipients.direct_payments_configuration import (
+    DirectPaymentsMisspelledLaNames,
 )
 
 direct_payments_columns = [
@@ -24,6 +29,12 @@ def main(
     lf = utils.scan_parquet(
         source=direct_payments_merged_source,
         selected_columns=direct_payments_columns,
+    )
+
+    lf.with_columns(
+        pl.col(DP.LA_AREA).replace(
+            DirectPaymentsMisspelledLaNames.DICT_TO_CORRECT_LA_NAMES
+        )
     )
 
     utils.sink_to_parquet(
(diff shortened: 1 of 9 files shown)


projects/_04_direct_payment_recipients/fargate/estimate_direct_payments.py
```diff
@@ -15,6 +18,17 @@
...
+la_name_replacements = {
+    "Bath & N E Somerset": ContemporaryCSSR.bath_and_north_east_somerset,
+    "Blackburn": ContemporaryCSSR.blackburn_with_darwen,
+    "Bournemouth, Christchurch and Poole": ContemporaryCSSR.bournemouth_christchurch_and_poole,
+    "Cornwall": ContemporaryCSSR.cornwall_and_isles_of_scilly,
+    "East Riding": ContemporaryCSSR.east_riding_of_yorkshire,
+    "Isles of Scilly": ContemporaryCSSR.cornwall_and_isles_of_scilly,
+    "Medway Towns": ContemporaryCSSR.medway,
+    "Southend": ContemporaryCSSR.southend_on_sea,
+}
```
garypolzin, line 30: this dictionary already exists so I'd call it from there are opposed to putting in the job itself - it's general business logic as opposed to job specific

`projects\_04_direct_payment_recipients\direct_payments_configuration.py`

garypolzin (approved):
One small tweak so I'll approve now

projects/_04_direct_payment_recipients/fargate/estimate_direct_payments.py
RoyPriceCS50: I did that originally, but the configuration script has pyspark imports, so putting it in the dockerfile means changing the requirements, or putting just that dict in a separate script for polars only.

Decided to just stick it in the job for simplicity.

garypolzin: ok, we don't want two versions of the same code - the old is generally better but once we switch this job to pyspark we'll want to remove the original so that only the new version exists

PR #1306, opened 2026-04-27
1554 - DPR estimates util rolling average

RoyPriceCS50:
## Description
Trello ticket [1554](https://trello.com/c/XYiNAK1b)

Converted DPR estimates util for calculating the rolling average of estimate proportion DPR employing staff to polars.

The native polars .rolling_mean() doesn't populate the earliest rows, until there are enough rows to satisfy the average period, therefore done it with rolling aggregation and join to original lazyframe.

## Testing
- [x] Unit tests passing

No pipeline run because only converting a util which is not being called.

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG

### DIFF
--- a/projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
+++ b/projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
@@ -62,6 +62,14 @@ rolling_mean_test_cases = [
         ],
     ),
     CalculateRollingMeanTestCase(
+        id="all_values_are_null_in_one_area",
+        data=[
+            ("area_1", 2018, None, None),
+            ("area_1", 2017, None, None),
+            ("area_1", 2016, None, None),
+        ],
+    ),
+    CalculateRollingMeanTestCase(
         id="rolling_handles_non_consecutive_years",
         data=[
             ("area_3", 2019, 0.3, 0.2),
@@ -74,6 +82,17 @@ rolling_mean_test_cases = [
         data=[
             ("area_3", 2017, 0.2, 0.15),
             ("area_3", 2016, 0.1, 0.1),
+        ],
+    ),
+    CalculateRollingMeanTestCase(
+        id="years_are_not_in_order",
+        data=[
+            ("area_1", 2020, 0.5, 0.4),
+            ("area_1", 2021, 0.6, 0.5),
+            ("area_1", 2018, 0.3, 0.2),
+            ("area_1", 2019, 0.4, 0.3),
+            ("area_1", 2016, 0.1, 0.1),
+            ("area_1", 2017, 0.2, 0.15),
         ],
     ),
 ]
--- a/projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
+++ b/projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
@@ -21,7 +21,7 @@ class CalculateRollingMeanTestCase:
         return pytest.param(self.data, id=self.id)
 
 
-create_imputed_ascwds_job_role_counts_test_cases = [
+rolling_mean_test_cases = [
     CalculateRollingMeanTestCase(
         id="all_values_are_populated_in_one_area",
         data=[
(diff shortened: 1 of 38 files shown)


projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
```diff
@@ -0,0 +1,40 @@
...
+            DP.YEAR_AS_INTEGER: pl.Int64,
+            DP.ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF: pl.Float64,
+            DP.ROLLING_AVERAGE_ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF: pl.Float64,
+        }
+        expected_lf = pl.LazyFrame(rows, schema=test_schema, orient="row")
+        test_lf = expected_lf.drop(
+            DP.ROLLING_AVERAGE_ESTIMATED_PROPORTION_OF_SERVICE_USERS_EMPLOYING_STAFF
+        )
+        returned_lf = job.calculate_rolling_mean(test_lf)
+        returned_lf.show(limit=20)
```
AchyutRatkanthiwar: remove show()

projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+
+class TestCalculateRollingMean(unittest.TestCase):
+    def test_function_returns_expected_values(self):
+        rows = [
+            ("area_1", 2021, 0.6, 0.5),
+            ("area_1", 2020, 0.5, 0.4),
+            ("area_1", 2019, 0.4, 0.3),
+            ("area_1", 2018, 0.3, 0.2),
```
AchyutRatkanthiwar, line 31: how does .rolling(index_column=DP.YEAR_AS_INTEGER, group_by=DP.LA_AREA, period="3i") work when we dont have data for a given year. say we dont have any data for 2018 does the 3i period work as expected? Not sure if we have such data but worth testing?

projects/_04_direct_payment_recipients/fargate/utils/estimate_direct_payments_utils/apply_rolling_average.py
AchyutRatkanthiwar, line 1: I haven't been deleting PySpark version code yet. are we doing that in the same PR? As we don't have outputs generated will it be better to wait and compare the outputs from both and then delete?

RoyPriceCS50: Yeah we;ll delete the pyspark stuff after signing off the outputs.

projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
RoyPriceCS50: Realised I slacked off on these... I've added more cases and formatted it better.

projects/_04_direct_payment_recipients/tests/fargate/utils/test_calculate_rolling_mean.py
```diff
@@ -0,0 +1,103 @@
...
+            ("area_3", 2016, 0.1, 0.1),
+        ],
+    ),
+    CalculateRollingMeanTestCase(
+        id="fewer_than_rolling_period_years_in_data",
+        data=[
+            ("area_3", 2017, 0.2, 0.15),
+            ("area_3", 2016, 0.1, 0.1),
+        ],
+    ),
```
MarionJHolloway, line 86: Can you do two more test cases please?

1) Using unsorted data (to test and lock in to sort function - feel free to copy and shuffle another small test case)
2) One to demonstrate what happens if a LA has only null data (can be a single row of nulls)

MarionJHolloway (changes requested):
Two quick tests to add, but otherwise looks good. Simple and clear, well done :)

MarionJHolloway (approved):

PR #1307, opened 2026-04-27
1555 - Convert DPR extrapolation to Polars

AchyutRatkanthiwar:
## Description
Trello ticket [#1555](https://trello.com/c/G2owHFct/1555-dpr-convert-modelextrapolation)

Converted DPR Extrapolation Ratio function to Polars and added tests for the same.

## Testing
- [x] Unit tests passing

No outputs generated at the moment.

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG


projects/_04_direct_payment_recipients/fargate/utils/models/extrapolation_ratio.py
RoyPriceCS50, line 1: I feel like this is too far on the Expr side and might be hard to debug when we come to actual results later on.

The ind cqc extrapolation in PR for 1502, that has a "prep step" that aggregates and joins, so along those lines, can this be similar.

boundaries_lf = lf.filter(not nulls).groupby(la areas).agg( min and max year interger aliased as first/last_year)

first_values_lf = lf.join(boundaries_lf, on=la areas).filter(year == first_year).select(la area, proportion column aliased as first_value, estimate from mean aliased as first_mean)

last_values_lf = same as above but referring to last_ each time.

Then join those three lf's to the original lf, so we've got columns showing what's what.

The conditions for before_first and after_last stay the same.

Then you can do this:

mean_ratio_first = pl.col(DP.ESTIMATE_USING_MEAN) / pl.col("first_mean")
mean_ratio_last = pl.col(DP.ESTIMATE_USING_MEAN) / pl.col("last_mean")

lf = lf.with_columns(
    pl.when(before_first)
    .then(mean_ratio_first * pl.col("first_value"))
    .when(after_last)
    .then(mean_ratio_last * pl.col("last_value"))
    .alias(DP.ESTIMATE_USING_EXTRAPOLATION_RATIO)
)

AchyutRatkanthiwar: Refactored the code to this now.

RoyPriceCS50 (approved):
Nice. Can un-drop the temp cols now if the final output has issues.

PR #1310, opened 2026-04-28
Added AI review checkbox to PR templates

garypolzin:
## Description
Added an AI reviewed checkbox to the PR templates

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/polars_migration_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/polars_migration_template.md
@@ -9,7 +9,7 @@ Trello ticket [#number](add link)
 - [ ] Outputs checked in Athena
 - [ ] Outputs compared using [SageMaker script](https://eu-west-2.console.aws.amazon.com/sagemaker/home?region=eu-west-2#/notebook-instances/data-engineering-notebook)
       - Notebook name: compare_outputs.ipynb
-- [ ] Output Screenshots added to Trello ticket 
+- [ ] Output Screenshots added to Trello ticket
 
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE/polars_migration_template.md
```diff
@@ -24,4 +24,5 @@ Trello ticket [#number](add link)
 - [ ] Use new file structure when saving to S3
       `{dataset}_{sub_dataset_if_relevant}_{order_of_process_if_relevant}_{very_brief_description}`
 - [ ] Updated CHANGELOG
+- [ ] Code reviewed by AI
```
AchyutRatkanthiwar, line 27: You can add a link to the Copilot instructions doc here for ease of access

AchyutRatkanthiwar (approved):
Looks good. just one suggestion re adding the link to instructions doc in the template for ease of access.

.github/PULL_REQUEST_TEMPLATE/polars_migration_template.md
garypolzin: good idea but is it possible? It's saved in sharepoint so I think the filepath will contain our name as part of it which pretty much makes it unique

AchyutRatkanthiwar: I think the link should still work. but it being in teams is also easy to access so we can go without adding it here.

garypolzin: yea ok, we can always come back and add it if we find it faffy

PR #1316, opened 2026-04-30
1565 dpr cornwall

RoyPriceCS50:
## Description
Trello ticket [1565](https://trello.com/c/9jjiVZ1u)

Cornwall and Isles of Scilly are separate areas in the raw DPR data, but a combined geography in our publications.

The bulk of this PR is a function to merge these areas. Counts of DPR's are summed across areas but proportions of DPR employing staff are taken from Cornwall only (Isles of Scilly has never given a proportion in over 10 years).

Some other bits i've changed are:
- Removing both areas from the dict to correct LA names.
- Remove "year" column from estimates data. We have "year_as_interger". Updated the function that was using this column.

## Testing
- [x] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn:aws:states:eu-west-2:856699698263:stateMachine:1565-dpr-cornwall-Direct-Payment-Recipients)

See Trello ticket for output comparison. 
I prepared the data as if it was last year, then took the whole summary dataset and compared it to main. No differences.
I took the estimates datasets as well. The sum of estimated_total_dpr_employing_staff last years C + IoS in main = Cornwall IoS in branch.

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ All notable changes to this project will be documented in this file.
 
 - Converted DPR estimates util estimate_service_users_employing_staff to polars.
 
+- Added a function to DPR estimates to group Cornwall and Isles of Scilly.
+
 ### Changed
 - Removed the PySpark version of IND CQC Clean and Validation jobs.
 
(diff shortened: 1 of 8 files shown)


MarionJHolloway (approved):
Looks good, my only thought is how will this affect importing data next year? Is it worth having a test case where there is a row of combined C&IOS data. I think it will just ignore it based on your code, but it might be a quick tweak to save headaches in the future. Not essential though, so I've approved.

PR #1333, opened 2026-05-21
1573 - Update Job Role Rolling Average Function

AchyutRatkanthiwar:
## Description
Trello ticket [#1573](https://trello.com/c/egY0jWu8/1573-l-improve-job-role-ratio-rolling-average)

This PR updates the `create_ascwds_job_role_rolling_ratio` function logic to support grouping by location size bands. I added configurable rolling-period grouping for COH, CHWN and NR Service Types.

The new implementation matches the old SPSS groupings.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1573-jr-rolling-avg-Ind-CQC-Filled-Post-Estimates-By-Role:a43dd4a4-1bea-4168-9bae-9ebe9164e664)

## Checklist
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -1,7 +1,31 @@
...
+        ((20, 29), "CHWN 20 to 29"),
+        ((30, None), "CHWN 30 plus"),
+    ],
+    PrimaryServiceType.non_residential: [
+        ((1, 24), "NR 1 to 24"),
+        ((25, 49), "NR 25 to 49"),
+        ((50, 74), "NR 50 to 74"),
+        ((75, 99), "NR 75 to 99"),
+        ((100, None), "NR 100 plus"),
+    ],
```
garypolzin: defining lower and upper here, and then having '< upper + 1' in the allocation feels a bit risky

can we switch to lower only
```
    PrimaryServiceType.care_home_only: [
        (1, "COH 1 to 9"),
        (10, "COH 10 to 19"),
        (20, "COH 20 to 29"),
        (30, "COH 30 plus"),
    ]
```

I'll paste what I think the follow up logic is in the expression

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -102,6 +129,37 @@ def get_percent_share_ratios(
...
+        if size_group_expr is None:
+            size_group_expr = pl.when(primary_col == service_type).then(bucket_expr)
+        else:
+            size_group_expr = size_group_expr.when(primary_col == service_type).then(
+                bucket_expr
+            )
+
+    return size_group_expr.otherwise(None).alias(
+        IndCQC.estimate_filled_posts_size_group
+    )
```
garypolzin: using lower bound only (haven't tested this)

```
estimate_col = pl.col(IndCQC.estimate_filled_posts)
primary_col = pl.col(IndCQC.primary_service_type)

expr = pl.lit(None)

for service_type, buckets in BUCKETS.items():
    for i, (lower, label) in enumerate(buckets):
        upper = buckets[i + 1][0] if i + 1 < len(buckets) else None

        condition = (
            (primary_col == service_type)
            & (estimate_col >= lower)
        )

        if upper is not None:
            condition = condition & (estimate_col < upper)

        expr = pl.when(condition).then(pl.lit(label)).otherwise(expr)
```

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/tests/fargate/utils/test_impute_utils.py
```diff
@@ -83,9 +83,35 @@ def test_create_ascwds_job_role_rolling_ratio(
...
-        pl_testing.assert_frame_equal(returned_lf, expected_lf, rel_tol=0.0001)
-        pl_testing.assert_frame_equal(returned_lf, expected_lf, rel_tol=0.0001)
-        pl_testing.assert_frame_equal(returned_lf, expected_lf, rel_tol=0.0001)
+        pl_testing.assert_frame_equal(
+            returned_lf, expected_lf, check_column_order=False, rel_tol=0.0001
+        )
+
+
+class TestEstimateFilledPostsSizeGroupExpression:
+    def test_estimate_filled_posts_size_group_expression(self):
```
garypolzin, line 97: looks a bit light on tests here

only tests one of the services and no boundary cases (eg. try 24.999 and 25.0)

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -148,5 +214,10 @@ def create_ascwds_job_role_rolling_ratio(
         estimated_job_role_posts_lf,
         input_col=IndCQC.ascwds_job_role_rolling_sum,
         output_col=IndCQC.ascwds_job_role_rolling_ratio,
+        groups=[
+            IndCQC.location_id,
+            IndCQC.cqc_location_import_date,
+            IndCQC.estimate_filled_posts_size_group,
+        ],
     )
     return estimated_job_role_posts_lf
```
garypolzin: can we drop `IndCQC.ascwds_job_role_rolling_sum` if it isn't used outside of this function?

garypolzin (commented):
Added a couple of comments, give me a shout if anything isn't clear

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
AchyutRatkanthiwar: Done and tested

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
AchyutRatkanthiwar: Do you mean estimate_filled_posts_size_group?
I kept it for output checking but I agree can remove it

garypolzin: sorry yea I did mean the _sum
we use the _ratio elsewhere, but I think the _sum are only created in order to get the _ratio and then aren't used again

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -1,7 +1,31 @@
+from typing import Optional
+
 import polars as pl
 
 from polars_utils.expressions import percentage_share
 from utils.column_names.ind_cqc_pipeline_columns import IndCqcColumns as IndCQC
+from utils.column_values.categorical_column_values import PrimaryServiceType
+
+ESTIMATE_FILLED_POSTS_SIZE_GROUPS = {
```
garypolzin: I'm not fully set on this so don't feel like you have to but when we've got defined things like this that are only used once, should we put it in the function itself?

definitely at the top if used multiple times, but when it's only used once I think my personal preference is to have it visible in the function - but again, happy either way

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -102,6 +129,27 @@ def get_percent_share_ratios(
     )
 
 
+def estimate_filled_posts_size_group_expression() -> pl.Expr:
+    """Create the expression to calculate the size group for estimated filled posts."""
```
garypolzin: this is quite a complex function to understand so could benefit from a more detailed docstring. Doesn't need to be overkill, but enough to get the gist of what's going on

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
garypolzin, line 220: this isn't part of your PR as such, by why don't we calculate the rolling ratio here after we've created the rolling sum, then drop the sum and join the ratios in?

the ratios should reflect the sums so should be more effiecient to calculate them on the aggregated file

basically flip the order A-B-D-C

garypolzin: Probbaly worth padding this docstring out too - think it's just missing a bit to say we calculate the rolling ratios at an aggregated level (service, size and job role) then join them back in at location level

garypolzin (commented):
Sorry some sub-par reviewing today, too hot to concentrate!
It's not your PR directly but I think the function can be optimised/simplified if we flip the order a bit but could be wrong

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -148,5 +204,10 @@ def create_ascwds_job_role_rolling_ratio(
         estimated_job_role_posts_lf,
         input_col=IndCQC.ascwds_job_role_rolling_sum,
         output_col=IndCQC.ascwds_job_role_rolling_ratio,
+        groups=[
+            IndCQC.location_id,
+            IndCQC.cqc_location_import_date,
+            IndCQC.estimate_filled_posts_size_group,
+        ],
```
garypolzin: then I think `groups` will be the same as `rolling_groups`?

garypolzin (commented):
edited previous comment

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
MarionJHolloway: I think when it's this big, it would just clutter the function  up a bit. Definitely for smaller stuff like a temporary value.

MarionJHolloway: Also, having it here would allow for unit testing if we wanted to, whereas in a function, it's harder to do that.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
AchyutRatkanthiwar: Yes, we can try that. I am still a bit unsure on implementation if it will change the outputs a little though. do you think creating a new ticket and working on that later is better or should I try to do and test this approach in this PR itself?

AchyutRatkanthiwar (commented):
Between me and Marion we have actioned most of all of the comments but one regarding changing the order of the rolling function. Need to decide if I should do it as part of same PR or create a new ticket for it.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
garypolzin: would a compromise be to put this above the function it's used in?
possibly a personal opinion but I like stuff close together where it makes sense (it's only used once by that function so ideally it lives in or next to that function)

AchyutRatkanthiwar: Done

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
AchyutRatkanthiwar: I have made this change and checked the outputs. They look good.
You will see I didnt use get_percent_share_ratios function as it needs id_per_locationid_import_date_job_role and it is not available at aggregation level. joining it back in meant doing what it was doing so tried normal polars calculation for getting ratios.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -137,16 +219,28 @@ def create_ascwds_job_role_rolling_ratio(
         )
     )
 
-    # STEP C: Join the rolling sum back to the main 152M row table
-    estimated_job_role_posts_lf = estimated_job_role_posts_lf.join(
+    # STEP C: Calculate ratio on the small ~50k-row dataset before the join.
+    # All locations sharing (primary_service_type, size_group, date) get the
+    # same ratio, so this is equivalent to computing it post-join but far cheaper.
```
garypolzin: keep the inline comments just to current behaviour, can get rid of the last part of this.

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -137,16 +219,28 @@ def create_ascwds_job_role_rolling_ratio(
         )
     )
 
-    # STEP C: Join the rolling sum back to the main 152M row table
-    estimated_job_role_posts_lf = estimated_job_role_posts_lf.join(
+    # STEP C: Calculate ratio on the small ~50k-row dataset before the join.
```
garypolzin: would try avoiding using actual numbers too - these will grow/change over time

it's useful information though so go for something generic like `significantly smaller aggregated dataset`

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -137,16 +219,28 @@ def create_ascwds_job_role_rolling_ratio(
...
+            pl.col(IndCQC.ascwds_job_role_rolling_sum)
+            / pl.col(IndCQC.ascwds_job_role_rolling_sum).sum().over(ratio_partition)
+        )
+        .cast(pl.Float32)
+        .alias(IndCQC.ascwds_job_role_rolling_ratio)
+    ).drop(
+        IndCQC.ascwds_job_role_rolling_sum
+    )  # drop sum before the join
+
+    # STEP D: Join only the ratio (no sum) back to the 152M-row table
```
garypolzin: Again, keep generic. The function is called create ratio so the expected behaviour would be to just add the ratio
`Join the ratio back to the full location level dataset`

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -137,16 +219,28 @@ def create_ascwds_job_role_rolling_ratio(
...
+    rolling_agg_lf = rolling_agg_lf.with_columns(
+        (
+            pl.col(IndCQC.ascwds_job_role_rolling_sum)
+            / pl.col(IndCQC.ascwds_job_role_rolling_sum).sum().over(ratio_partition)
+        )
+        .cast(pl.Float32)
+        .alias(IndCQC.ascwds_job_role_rolling_ratio)
+    ).drop(
+        IndCQC.ascwds_job_role_rolling_sum
+    )  # drop sum before the join
```
garypolzin: fairly simple/readable so you can drop this comment

projects/_03_independent_cqc/_07_estimate_filled_posts_by_job_role/fargate/utils/impute_utils.py
```diff
@@ -102,31 +108,107 @@ def get_percent_share_ratios(
...
+        IndCQC.primary_service_type,
+        IndCQC.estimate_filled_posts_size_group,
+        IndCQC.main_job_role_clean_labelled,
+    ]
     order_key = IndCQC.cqc_location_import_date
     monthly_groups = rolling_groups + [order_key]
-    # STEP A: Pre-aggregate down to monthly totals
-    # (Shrinks 152M rows -> ~50k rows instantly via Hash Aggregation)
+
+    # STEP A: Pre-aggregate to monthly totals (~50k rows)
```
garypolzin: remove row count

garypolzin (approved):
Looks good now thanks, just left a few comments about inline comments

PR #1339, opened 2026-06-05
1626 wkr val fix

RoyPriceCS50:
## Description
Trello ticket [1626](https://trello.com/c/NVHauQV0)

We recoded technicians into "other" a while ago by changing the job role code. 
We then apply labels to those codes. So technicians don't need a label anymore because their code doesn't exist.

When we did that, cleaned worker validation should have failed because there are 37 job role labels and 38 in the expected set, 
but it's been passing. 

All pyspark validations raise warnings, then there's a separate function to raise an exception if warnings are found. 
That second part was missing from raw and cleaned worker validation.

This PR:
- Fixes raw and cleaned worker validation scripts so they start raising exceptions.
- Removes technician from worker job role labels class.

## Testing
- [x] Unit tests passing
- [x] Successful:
[raw worker validation](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/1626-wkr-val-fix-validate_ascwds_worker_raw_data_job/runs)
[cleaned worker validation](https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/1626-wkr-val-fix-validate_ascwds_worker_cleaned_data_job/runs)

No Athena outputs to check since validations passed.

- [x] Updated CHANGELOG

## Reviewer checklist

### DIFF
--- a/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_cleaned_data.py
+++ b/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_cleaned_data.py
@@ -1,13 +1,18 @@
 import os
 import sys
 
+from pyspark.sql.dataframe import DataFrame
+
 os.environ["SPARK_VERSION"] = "3.5"
 
 from utils import utils
 from utils.validation.validation_rules.ascwds_worker_cleaned_validation_rules import (
     ASCWDSWorkerCleanedValidationRules as Rules,
 )
-from utils.validation.validation_utils import validate_dataset
+from utils.validation.validation_utils import (
+    raise_exception_if_any_checks_failed,
+    validate_dataset,
+)
 
 
 def main(cleaned_ascwds_worker_source: str, report_destination: str):
@@ -19,12 +24,15 @@ def main(cleaned_ascwds_worker_source: str, report_destination: str):
 
     utils.write_to_parquet(check_result_df, report_destination, mode="overwrite")
 
+    if isinstance(check_result_df, DataFrame):
+        raise_exception_if_any_checks_failed(check_result_df)
+
 
 if __name__ == "__main__":
     print("Spark job 'validate_ascwds_worker_cleaned_data' starting...")
     print(f"Job parameters: {sys.argv}")
 
-    (cleaned_ascwds_worker_source, report_destination) = utils.collect_arguments(
+    cleaned_ascwds_worker_source, report_destination = utils.collect_arguments(
         (
             "--cleaned_ascwds_worker_source",
             "Source s3 directory for parquet ascwds worker api cleaned dataset",
--- a/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_cleaned_data.py
+++ b/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_cleaned_data.py
@@ -24,8 +24,7 @@ def main(cleaned_ascwds_worker_source: str, report_destination: str):
 
     utils.write_to_parquet(check_result_df, report_destination, mode="overwrite")
 
-    if isinstance(check_result_df, DataFrame):
-        raise_exception_if_any_checks_failed(check_result_df)
+    raise_exception_if_any_checks_failed(check_result_df)
 
 
 if __name__ == "__main__":
--- a/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_raw_data.py
+++ b/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_raw_data.py
@@ -24,8 +24,7 @@ def main(raw_ascwds_worker_source: str, report_destination: str):
 
     utils.write_to_parquet(check_result_df, report_destination, mode="overwrite")
 
-    if isinstance(check_result_df, DataFrame):
-        raise_exception_if_any_checks_failed(check_result_df)
+    raise_exception_if_any_checks_failed(check_result_df)
 
 
 if __name__ == "__main__":
--- a/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_raw_data.py
+++ b/projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_raw_data.py
@@ -1,13 +1,18 @@
 import os
 import sys
 
+from pyspark.sql.dataframe import DataFrame
+
 os.environ["SPARK_VERSION"] = "3.5"
 
 from utils import utils
 from utils.validation.validation_rules.ascwds_worker_raw_validation_rules import (
     ASCWDSWorkerRawValidationRules as Rules,
 )
-from utils.validation.validation_utils import validate_dataset
+from utils.validation.validation_utils import (
+    raise_exception_if_any_checks_failed,
+    validate_dataset,
+)
 
 
 def main(raw_ascwds_worker_source: str, report_destination: str):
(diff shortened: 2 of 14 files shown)


projects/_01_ingest/unittest_data/ingest_test_file_schemas.py
```diff
@@ -551,6 +552,17 @@ class ValidateASCWDSWorkerRawData:
...
+    validated_ascwds_worker_schema = StructType(
+        [
+            StructField(Validation.check, StringType(), True),
+            StructField(Validation.check_level, StringType(), True),
+            StructField(Validation.check_status, StringType(), True),
+            StructField(Validation.constraint, StringType(), True),
+            StructField(Validation.constraint_status, StringType(), True),
+            StructField(Validation.constraint_message, StringType(), True),
+        ]
+    )
```
garypolzin: this function is already tested so we don't need new test data

I'm guessing the tests fail without it currently since you added the function into the patch call but you can mock one instead of actually creating it
`validate_dataset_patch.return_value= Mock(name=validated_df)`

projects/_01_ingest/unittest_data/ingest_test_file_schemas.py
```diff
@@ -578,6 +590,17 @@ class ValidateASCWDSWorkerCleanedData:
         ]
     )
 
+    validated_cleaned_ascwds_worker_schema = StructType(
```
garypolzin: as above

projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_cleaned_data.py
```diff
@@ -19,12 +24,15 @@ def main(cleaned_ascwds_worker_source: str, report_destination: str):
 
     utils.write_to_parquet(check_result_df, report_destination, mode="overwrite")
 
+    if isinstance(check_result_df, DataFrame):
```
garypolzin: we have this `if` statement in all of our validations but I have no idea why

`check_result_df` will always be a DataFrame because the function that creates it (`validate_dataset(...)`) returns a DataFrame

I can't think of a scenario where it won't be, and if it isn't a df then I think write_to_parquet would fail before it even gets to this step, so I'd just get rid of it

projects/_01_ingest/ascwds/jobs/validate_ascwds_worker_raw_data.py
```diff
@@ -19,12 +24,15 @@ def main(raw_ascwds_worker_source: str, report_destination: str):
 
     utils.write_to_parquet(check_result_df, report_destination, mode="overwrite")
 
+    if isinstance(check_result_df, DataFrame):
```
garypolzin: as above

projects/_01_ingest/ascwds/tests/jobs/test_validate_ascwds_worker_cleaned_data.py
```diff
@@ -19,26 +21,37 @@ def setUp(self) -> None:
...
         self,
         read_from_parquet_patch: Mock,
+        validate_dataset_patch: Mock,
+        raise_exception_if_any_checks_failed_patch: Mock,
         write_to_parquet_patch: Mock,
     ):
         read_from_parquet_patch.side_effect = [
             self.test_cleaned_ascwds_worker_df,
         ]
+        validate_dataset_patch.side_effect = [self.test_validated_ascwds_worker_df]
```
garypolzin: replace with mock df (see more detailed comment elsewhere)

garypolzin (commented):
Couple of comments - a lot of them are in other validation scripts too but doesn't mean we should replicate it

utils/validation/validation_rules/ascwds_worker_cleaned_validation_rules.py
```diff
@@ -28,7 +28,9 @@ class ASCWDSWorkerCleanedValidationRules:
             AWKClean.main_job_role_clean_labelled: CatValues.main_job_role_labels_column_values.categorical_values,
         },
         RuleName.distinct_values: {
-            AWKClean.main_job_role_clean: CatValues.main_job_role_id_column_values.count_of_categorical_values,
+            AWKClean.main_job_role_clean: (
+                CatValues.main_job_role_id_column_values.count_of_categorical_values - 2
+            ),  # Subtracting 2 because tecnician and care navigator raw worker codes get recoded before labels are applied. Therfore they exist in MainJobRoleID but not in MainJobRoleLabels classes.
```
garypolzin: spelling of `tecnician`

projects/_01_ingest/unittest_data/ingest_test_file_schemas.py
```diff
@@ -37,6 +37,7 @@
 from utils.column_names.raw_data_files.ons_columns import (
     OnsPostcodeDirectoryColumns as ONS,
 )
+from utils.column_names.validation_table_columns import Validation
```
garypolzin: delete

utils/validation/validation_rules/ascwds_worker_cleaned_validation_rules.py
```diff
@@ -28,7 +28,9 @@ class ASCWDSWorkerCleanedValidationRules:
             AWKClean.main_job_role_clean_labelled: CatValues.main_job_role_labels_column_values.categorical_values,
         },
         RuleName.distinct_values: {
-            AWKClean.main_job_role_clean: CatValues.main_job_role_id_column_values.count_of_categorical_values,
+            AWKClean.main_job_role_clean: (
+                CatValues.main_job_role_id_column_values.count_of_categorical_values - 2
+            ),  # Subtracting 2 because technician and care navigator raw worker codes get recoded before labels are applied. Therfore they exist in MainJobRoleID but not in MainJobRoleLabels classes.
```
garypolzin: I'm a bit undecided on this one, it's quite manual, but can't think about how to do it otherwise

one option is to scrap it, we have the check on the labels, and the labels are literally the numbers converted to text, so I don't think there's a way of having the correct labels without the correct IDs

thoughts @AchyutRatkanthiwar / @MarionJHolloway ?

utils/column_values/categorical_column_values.py
garypolzin, line 254: delete this too

garypolzin (approved):
Some minor changes to make and a call for ideas on the more manual step but happy to approve it on the whole

utils/validation/validation_rules/ascwds_worker_cleaned_validation_rules.py
MarionJHolloway: There is a mechanism for handling this in the column values classes that I built into it. Happy to show you Roy, just give me a ring.

utils/column_values/categorical_column_values.py
RoyPriceCS50: That class is used to validate the raw worker data, so it'l fail without technicians.

MarionJHolloway: That's because it's not set up properly. The base class has mechanisms for handling this (I wrote it ages ago.) Happy to show you on a video call.

utils/validation/validation_rules/ascwds_worker_cleaned_validation_rules.py
RoyPriceCS50: Yeah i've seen it. It removes a string, but I need to remove 2 strings.

I think separate ticket to make that change... and have be minus 2 until that ticket is done ?

PR #1352, opened 2026-06-11
1647 cnv imptn

RoyPriceCS50:
## Description
Trello ticket [1647](https://trello.com/c/o3JvBglE)

Converted the util function model_imputation_with_extrapolation_and_interpolation to polars as model_imputation.

Called the function 3/4 times in imputation job script (waiting for ascwds_pir_merged to b converted to polars).
Call the function 1/3 times in estimates job script (same issue).

## Testing
- [x] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1647-cnv-imptn-Ind-CQC-Filled-Post-Estimates:6b6eadf3-38b6-4e28-9b30-41e6936432bb)
- [ ] Outputs compared using [SageMaker script](https://eu-west-2.console.aws.amazon.com/sagemaker/home?region=eu-west-2#/notebook-instances/data-engineering-notebook)
      - Notebook name: compare_outputs.ipynb
- [x] Output Screenshots added to Trello ticket

See Trello ticket. I ran the dataset comparison query, and then got the variable sums for polars/pyspark versions of the datasets.

For the variables I could create, their values are within 0.5 at row level. The sums are slightly different because I made the variable type Float 32 instead of Float64.

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist
- [x] PR solves the Trello ticket requirements
- [x] Outputs appear reasonable and understandable
- [x] The overall approach is appropriate and not over-engineered
- [x] No obvious scalability, performance or interdependency concerns
- [x] Tests appropriately cover the main behaviour/edge cases
- [x] Naming and structure are broadly understandable and consistent
- [x] Docstrings are sufficient for future maintenance
- [x] Docstrings cover non-obvious behaviour

### DIFF
--- a/projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/estimate_ind_cqc_filled_posts.py
+++ b/projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/estimate_ind_cqc_filled_posts.py
@@ -5,6 +5,7 @@ from projects._03_independent_cqc._06_estimate_filled_posts.fargate.utils.models
 from projects._03_independent_cqc._06_estimate_filled_posts.fargate.utils.models.utils import (
     enrich_with_model_predictions,
 )
+from projects._03_independent_cqc.utils.imputation.imputation import model_imputation
 from utils.column_names.ind_cqc_pipeline_columns import IndCqcColumns as IndCQC
 
 ind_cqc_columns = [
@@ -96,11 +97,32 @@ def main(
 
     lf = combine_non_res_with_and_without_dormancy_models(lf)
 
-    # model_imputation_with_extrapolation_and_interpolation - imputed_posts_care_home_model
+    lf = model_imputation(
+        lf,
+        IndCQC.ascwds_pir_merged,
+        IndCQC.care_home_model,
+        IndCQC.imputed_posts_care_home_model,
+        care_home=True,
+        extrapolation_method="nominal",
+    )
 
-    # model_imputation_with_extrapolation_and_interpolation - imputed_posts_non_res_combined_model
+    lf = model_imputation(
+        lf,
+        IndCQC.ascwds_pir_merged,
+        IndCQC.non_res_combined_model,
+        IndCQC.imputed_posts_non_res_combined_model,
+        care_home=False,
+        extrapolation_method="nominal",
+    )
 
-    # model_imputation_with_extrapolation_and_interpolation - imputed_pir_filled_posts_model
+    lf = model_imputation(
+        lf,
+        IndCQC.pir_filled_posts_model,
+        IndCQC.non_res_combined_model,
+        IndCQC.imputed_pir_filled_posts_model,
+        care_home=False,
+        extrapolation_method="nominal",
+    )
 
     # merge_columns_in_order
 
--- a/projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/estimate_ind_cqc_filled_posts.py
+++ b/projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/estimate_ind_cqc_filled_posts.py
@@ -97,23 +97,24 @@ def main(
 
     lf = combine_non_res_with_and_without_dormancy_models(lf)
 
-    lf = model_imputation(
-        lf,
-        IndCQC.ascwds_pir_merged,
-        IndCQC.care_home_model,
-        IndCQC.imputed_posts_care_home_model,
-        care_home=True,
-        extrapolation_method="nominal",
-    )
+    # Uncommented these call when ascwds_pir_merged has been converted to polars.
+    # lf = model_imputation(
+    #     lf,
+    #     IndCQC.ascwds_pir_merged,
+    #     IndCQC.care_home_model,
+    #     IndCQC.imputed_posts_care_home_model,
+    #     care_home=True,
+    #     extrapolation_method="nominal",
+    # )
 
-    lf = model_imputation(
-        lf,
-        IndCQC.ascwds_pir_merged,
-        IndCQC.non_res_combined_model,
-        IndCQC.imputed_posts_non_res_combined_model,
-        care_home=False,
-        extrapolation_method="nominal",
-    )
+    # lf = model_imputation(
+    #     lf,
+    #     IndCQC.ascwds_pir_merged,
+    #     IndCQC.non_res_combined_model,
+    #     IndCQC.imputed_posts_non_res_combined_model,
+    #     care_home=False,
+    #     extrapolation_method="nominal",
+    # )
 
     lf = model_imputation(
         lf,
--- a/projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
+++ b/projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
@@ -2655,18 +2655,6 @@ class ModelImputation:
     model_column_name: str = "trend_model"
     imputed_values_column_name: str = "imputed_values"
 
-    imputed_rows = [
-        ("1-002", date(2023, 1, 1), CareHome.not_care_home, None, 1.0, 19.0),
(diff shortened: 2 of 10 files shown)


projects/_03_independent_cqc/_06_estimate_filled_posts/fargate/estimate_ind_cqc_filled_posts.py
```diff
@@ -96,11 +97,33 @@ def main(
 
     lf = combine_non_res_with_and_without_dormancy_models(lf)
 
-    # model_imputation_with_extrapolation_and_interpolation - imputed_posts_care_home_model
-
-    # model_imputation_with_extrapolation_and_interpolation - imputed_posts_non_res_combined_model
-
-    # model_imputation_with_extrapolation_and_interpolation - imputed_pir_filled_posts_model
+    # Uncommented these call when ascwds_pir_merged has been converted to polars.
```
MarionJHolloway, line 100: It's a bit late now but it feels like we probably should have done this step first!

MarionJHolloway: Can you add to the ticket for ascwds_pir_merged that these calls will need checking then

projects/_03_independent_cqc/utils/imputation/imputation.py
```diff
@@ -0,0 +1,121 @@
...
+        care_home_filter_value: str = CareHome.care_home
+    else:
+        care_home_filter_value: str = CareHome.not_care_home
+
+    locs_with_values_expr = (
+        pl.col(column_with_null_values)
+        .count()
+        .over([IndCqc.location_id, IndCqc.care_home])
+        > 0
+    ) & (pl.col(IndCqc.care_home) == care_home_filter_value)
```
MarionJHolloway: We have this a standard expression. Might be worth using it here. Lives in polars_utils/expressions

projects/_03_independent_cqc/utils/imputation/imputation.py
```diff
@@ -0,0 +1,121 @@
...
+
+    temp_bool = "temp_bool"
```
MarionJHolloway: Do we need to add and drop this column? Could we just pass the expression to the filter?

projects/_03_independent_cqc/utils/imputation/imputation.py
```diff
@@ -0,0 +1,121 @@
...
+        lf (pl.LazyFrame): The input LazyFrame containing the column_with_null_values.
+        column_with_null_values (str): The name of the column containing null
+            values to be imputed.
+        model_column_name (str): The name of the column containing the model
+            values used for imputation.
+        imputed_column_name (str): The name of the new imputated column.
+        care_home (bool): True if imputation is for care homes, False if it is
+            for non residential.
+        extrapolation_method (str): The choice of method.
+            Must be either 'nominal' or 'ratio'.
```
MarionJHolloway, line 47: I assume there's no error handling here because the extrapolation method covers it?

projects/_03_independent_cqc/utils/tests/imputation/test_imputation.py
```diff
@@ -0,0 +1,152 @@
...
+                ("1-004", CareHome.not_care_home, 10.0),
+                ("1-004", CareHome.not_care_home, None),
+                ("1-005", CareHome.not_care_home, 10.0),
+                ("1-006", CareHome.not_care_home, None),
+                ("1-006", CareHome.not_care_home, None),
+            ],
+            schema=Schemas.input_split_dataset_for_imputation_schema,
+            orient="row",
+        )
+        returned_inputed_when_care_home, returned_non_inputed_when_care_home = (
```
MarionJHolloway: imputed, not inputed

I've seen this a couple of times, worth a global find and replace

projects/_03_independent_cqc/utils/tests/imputation/test_imputation.py
```diff
@@ -0,0 +1,152 @@
...
+            .to_series()
+            .to_list(),
+            ["1-001", "1-001", "1-002"],
+        )
+        self.assertEqual(
+            returned_non_inputed_when_care_home.select(IndCQC.location_id)
+            .collect()
+            .to_series()
+            .to_list(),
+            ["1-003", "1-003", "1-004", "1-004", "1-005", "1-006", "1-006"],
```
MarionJHolloway: Your location id is not unique, so you're going to want to either add a key column to be able to specify which row you're checking, or if the exact rows don't matter you could use a set rather than a list.

I'd favour the first option with the keys because I think it would be clearer

projects/_03_independent_cqc/utils/tests/imputation/test_imputation.py
```diff
@@ -0,0 +1,152 @@
...
+        self,
+        split_dataset_for_imputation_mock: Mock,
+        model_extrapolation_mock: Mock,
+        model_interpolation_mock: Mock,
+    ):
+        split_dataset_for_imputation_mock.return_value = (
+            self.imputed_lf,
+            self.non_imputed_lf,
+        )
+        model_extrapolation_mock.return_value = self.imputed_lf
```
MarionJHolloway, line 47: Is there a reason you're returning lazy frames with data that you're not checking? You could just pass an empty lazy frame if you're going to use a return value for each call.

projects/_03_independent_cqc/unittest_data/polars_ind_cqc_test_file_data.py
```diff
@@ -2639,6 +2639,79 @@ class ModelExtrapolation:
...
+    non_imputed_rows = [
+        ("1-001", date(2023, 1, 1), CareHome.care_home, 10.0, 1.0, None),
+        ("1-001", date(2023, 2, 1), CareHome.care_home, None, 2.0, None),
+        ("1-001", date(2023, 3, 1), CareHome.care_home, 30.0, 3.0, None),
+        ("1-003", date(2023, 3, 1), CareHome.not_care_home, None, 1.0, None),
+    ] # fmt: skip
+
+    expected_model_imputation_test_cases = [
+        ModelImputationTestCase(
+            id="when_values_will_be_extrapolated",
```
MarionJHolloway, line 2660: Worth adding a case where you have a mix of care home/ non res data?

MarionJHolloway (changes requested):
A few small bits to check or amend.

projects/_03_independent_cqc/utils/imputation/imputation.py
RoyPriceCS50: Dont think I can do this?  

The equivalence is dependent on what gets passed in from "care_home" argument, either true or false. It's not actual care home "Y" and "N".

projects/_03_independent_cqc/utils/imputation/imputation.py
RoyPriceCS50: Yes. I didn't want to repeat it since model_extrapolation will raise the error.

projects/_03_independent_cqc/utils/imputation/imputation.py
MarionJHolloway: You need to pass the whole expression, not a bool or string, like this:
```
if care_home:
        care_home_filter_expr: pl.Expr = is_care_home()
    else:
        care_home_filter_expr: pl.Expr = is_not_care_home()

    locs_with_values_expr = (
        pl.col(column_with_null_values)
        .count()
        .over([IndCqc.location_id, IndCqc.care_home])
        > 0
    ) & (care_home_filter_expr)
```

MarionJHolloway (approved):
Looks good!

PR #1381, opened 2026-06-30
1708 - Workplace Clean - convert set_column_bounds() and cast_to_int() to Polars

garypolzin:
## Description
Trello ticket [#1708](https://trello.com/c/t0NzxVrT/1708-clean-workplace10-write-polars-code-for-setcolumnbounds-and-call-in-job)

This PR is to convert the PySpark cast_to_int() and set_column_bounds() functions to Polars.

**cast_to_int()**
Previously this was a function but is now inline and takes a list of integers. `strict=False` is set to return null values for invalid strings, such as empty " "

**set_column_bounds()**
This was a fairly complex function in PySpark with the option to have upper and/or lower bound. It's only used in the ASCWDS workplace job and only uses a minimum of 1 so I've just written this inline as it's a simple when, then, otherwise clause

## Testing
- [X] Unit tests passing
- [X] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1708-set-col-bounds-Transform-ASCWDS-Data:c0f87c92-f600-45fb-a189-1b01a246b62e)
- [X] Outputs checked in Athena
- [X] Output Screenshots added to Trello ticket

- [X] Check that similar utils functions don't already exist, or if one can be created
- [X] Check that utils are being saved in the appropriate place
- [X] Used LazyFrame option for test data
- [X] Unit tests added
- [X] Docstrings added
- [X] Added comment to pyspark functions which have been migrated with the link to the new location
- [X] Updated CHANGELOG
- [X] Code reviewed by AI

## Reviewer checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ parameters:
     default: ""
   terraform_version:
     type: string
-    default: 1.13.5
+    default: 1.15.7
 
 orbs:
   aws-cli: circleci/aws-cli@5.4.1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ parameters:
     default: ""
   terraform_version:
     type: string
-    default: 1.15.7
+    default: 1.13.5
 
 orbs:
   aws-cli: circleci/aws-cli@5.4.1
--- a/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
+++ b/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
@@ -97,7 +97,7 @@ def main(
     #     ascwds_workplace_df
     # )
 
-    lf = lf.with_columns(pl.col(INT_COLUMNS).cast(pl.Int32))
+    lf = lf.with_columns(pl.col(INT_COLUMNS).cast(pl.Int32, strict=False))
 
     lf = lf.with_columns(
         pl.when(pl.col(AWPClean.total_staff) >= MIN_VALID_WORKFORCE_COUNT)
--- a/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
+++ b/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
@@ -1,11 +1,13 @@
 import polars as pl
 
-from polars_utils import cleaning_utils as cUtils
 from polars_utils import utils
 from projects._01_ingest.ascwds.fargate.utils import clean_workplace_utils as wUtils
 from utils.column_names.cleaned_data_files.ascwds_workplace_cleaned import (
     AscwdsWorkplaceCleanedColumns as AWPClean,
 )
+
+INT_COLUMNS: list[str] = [AWPClean.total_staff, AWPClean.worker_records]
+MIN_VALID_WORKFORCE_COUNT: int = 1
 
 
 def main(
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ parameters:
     default: ""
   terraform_version:
     type: string
-    default: 1.13.5
+    default: 1.15.7
```
RoyPriceCS50: Separate PR?

projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
```diff
@@ -95,24 +97,18 @@ def main(
...
-    # trello 1708
-    # ascwds_workplace_df = cUtils.set_column_bounds(
-    #     ascwds_workplace_df,
-    #     AWPClean.worker_records,
-    #     AWPClean.worker_records_bounded,
-    #     AscwdsScaleVariableLimits.worker_records_lower_limit,
-    # )
+    lf = lf.with_columns(pl.col(INT_COLUMNS).cast(pl.Int32, strict=False))
+
+    lf = lf.with_columns(
```
RoyPriceCS50, line 103: set_column_bounds is called inside of set_bounds_for_columns.

That loops through bounding columns in cleaning capacity tracker data.

.circleci/config.yml
garypolzin: ah yea that was just temporary to see if it helped - it didn't
I've reverted it back

projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
RoyPriceCS50: I think you can use the same column expansion idea as line 100.

Something like:

lf = lf.withColumns(
    pl.when(pl.col([LIST_OF_COLUMNS]) >= MIN_VALID_WORKFORCE_COUNT)
        .then(pl.col([LIST_OF_COLUMNS]))
        .otherwise(None)
        .name.suffix("_bounded")

garypolzin: ah yea I spotted that, it's a strange combo of functions which you'd completely rewrite in polars. The CT version has lower and upper bounds whereas this is just lower - essentially you'd be overcomplicating a function to cater for several optional scenarios when it's simple/readable enough to just put it inline as is

RoyPriceCS50: It can be done like this:

class TestColumnBoundingLoop:
    def test_idea(self):
        test_lf = pl.LazyFrame(
            {
                AWPClean.total_staff: [0, 1, 5],
                AWPClean.worker_records: [0, 1, 5],
            }
        )

        returned_lf = test_lf.with_columns(
            pl.when(pl.col([AWPClean.total_staff, AWPClean.worker_records]) >= 1)
            .then(pl.col([AWPClean.total_staff, AWPClean.worker_records]))
            .otherwise(None)
            .name.suffix("_bounded")
        )

        expected_lf = pl.LazyFrame(
            {
                AWPClean.total_staff: [0, 1, 5],
                AWPClean.worker_records: [0, 1, 5],
                "totalstaff_bounded": [None, 1, 5],
                "wkrrecs_bounded": [None, 1, 5],
            }
        )

        pl_testing.assert_frame_equal(returned_lf, expected_lf)

RoyPriceCS50: ooof sorry about formatting

garypolzin: ah that's nice, didn't know you could do that!

RoyPriceCS50 (approved):
Looks good.

PR #1385, opened 2026-07-01
1705 Migrate apply_categorical_labels

MarionJHolloway:
## Description
Trello ticket [#1705](https://trello.com/c/aNGdtcHs/1705-clean-workplace6-write-polars-code-for-applycategoricallabels-and-call-in-job)

apply_categorical_labels()

## Testing
- [x] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1705-cl-wkp-6-Transform-ASCWDS-Data:c58f1b1a-b0ba-47c0-9c09-6804f98a634c)

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


polars_utils/cleaning_utils.py
```diff
@@ -50,6 +51,99 @@ def add_aligned_date_column(
...
+    suffix = f"_{DLC.label}s"
+
+    for column_name in column_names:
+        filtered_labels_df = (
+            labels_lf.filter(pl.col(DLC.column_name) == column_name)
+            .sort(DLC.code, DLC.label)
+            .unique(subset=[DLC.column_name, DLC.code], keep="first")
+            .collect()
+            .sort(DLC.code, DLC.label)
+        )
```
garypolzin: I potentially might need talking through this but in case it's easy I'll ask here

1 - what's the purpose of the second .sort()? It looks like the data has already been sorted before unique() so just wondering if it's still required.

2 - I also wonder whether we should silently resolve duplicate code mappings using unique(keep="first"). If the lookup table contains duplicate codes with different labels, that's a definite issue and raising an error might be safer than secretly selecting one

3 - is it better to collect once before rather than collecting once per column?
```
    labels_df = labels_lf.collect()

    for column in column_names:
        filtered_labels_df = labels_df.filter(...)
```

garypolzin: 3B - can we avoid the initial .collect() in rows 84-88 and incorporate that here too

```
    labels_df = labels_lf.collect()

    for column in column_names:
        filtered_labels_df = labels_df.filter()

        if filtered_labels_df .is_empty():
            raise KeyError(f"No label mapping found for {column_name}.")
```

This way there's only ever one collection of the labels_df

polars_utils/cleaning_utils.py
```diff
@@ -50,6 +51,99 @@ def add_aligned_date_column(
...
+                .name.suffix(suffix)
+            )
+        elif add_as_new_column == False:
+            yield (
+                pl.col(column_name).replace_strict(
+                    old=old_vals,
+                    new=new_vals,
+                    default=pl.col(column_name),
+                )
+            )
```
garypolzin: can save on repeating code here with

```
expr = pl.col(column_name).replace_strict(
    old=old_vals,
    new=new_vals,
    default=pl.col(column_name),
)

if add_as_new_column:
    expr = expr.name.suffix(suffix)

yield expr
```

garypolzin (commented):
Added a couple of comments which potentially might be my lack of understanding!

garypolzin (commented):
Added a couple of comments which potentially might be my lack of understanding!

polars_utils/cleaning_utils.py
MarionJHolloway: So, you're right to flag this. 

The unique step was for the reverse mapping (duplicate labels, not codes) so need removing. 

The first sort was to prep for the unique step, so that can also go. 

The second sort is so that the two .get_column methods get a list in the same order (s0 that one needs to stay).

If we collect before filtering etc, we lose the benefits of lazy evaluation for that dataset.

The column name collection is the way polars recommends to do it most efficiently and we want the error handling to come before data processing, so I wouldn't combine the collections on that basis.

MarionJHolloway: Don't forget collect automatically uses optimisations similar to scan parquet, and in each of these collects we're collecting a different thing. That allows polars to control as much of the lazy evaluation as possible and pushes collection to the last possible moment wherever possible.

garypolzin (approved):
Happy to approve following call

PR #1398, opened 2026-07-08
1763 - Use Polars selectors for date column casting

garypolzin:
## Description
Trello ticket [#1763](https://trello.com/c/iKsM4yN6/1763-s-use-polars-selectors-for-castdatestringstodates)

Replace manual schema inspection and column list creation with Polars selectors when identifying date columns to cast.

- Use selectors to identify string columns containing the date identifier.
- Exclude import_date using selector logic.
- Apply the date parsing expression directly to the selected columns.

## Testing
- [X] Unit tests passing
- [X] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1763-polars-selectors-Transform-ASCWDS-Data:802731f1-3432-40cd-b9c1-6e7ccf1d10f5)
- [X] Outputs checked in Athena

image

- [X] Docstrings added/updated
- [X] Updated CHANGELOG
- [X] Code reviewed by AI

## Reviewer checklist

### DIFF
--- a/polars_utils/cleaning_utils.py
+++ b/polars_utils/cleaning_utils.py
@@ -1,6 +1,7 @@
 from typing import Generator, List
 
 import polars as pl
+import polars.selectors as cs
 
 from polars_utils.expressions import is_care_home, is_not_care_home
 from utils.column_names.data_labels_columns import DataLabelsColumns as DLC
@@ -251,7 +252,7 @@ def cast_date_strings_to_dates(
     raw_date_format: str = "%d/%m/%Y",
 ) -> pl.LazyFrame:
     """
-    Converts columns to date type.
+    Parses string date columns into `Date` values.
 
     Columns in LazyFrame with date_column_identifier in their name excluding
     'import_date' and values formatted as raw_date_format will be type cast.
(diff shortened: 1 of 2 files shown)


polars_utils/cleaning_utils.py
```diff
@@ -265,15 +266,12 @@ def cast_date_strings_to_dates(
...
         pl.LazyFrame: The input LazyFrame with columns cast to dates.
     """
-    date_columns = [
-        col
-        for col in lf.collect_schema().names()
-        if date_column_identifier in col and "import_date" not in col
-    ]
+    date_columns = (
+        cs.string()
+        & cs.contains(date_column_identifier)
```
MarionJHolloway, line 271: Is it worth adding logic to include last loggedin here too? Would save renaming it in the pipeline.

MarionJHolloway (commented):
Were you planning to call this in the cleaning job as well or just keep it unused for now?

polars_utils/cleaning_utils.py
garypolzin: as it's a generic util function which will eventually be used in several datasets when we convert them to polars I'd be tempted to not include the specific last loggedin logic and deal with that separately with the rename

it is already called in the cleaning job

MarionJHolloway (approved):
Makes sense, approved :)

PR #1401, opened 2026-07-09
1763 - Remove .as_expr() following polars upgrade

garypolzin:
## Description
Trello ticket [#1763](https://trello.com/c/iKsM4yN6/1763-s-use-polars-selectors-for-castdatestringstodates)

My version of polars was older than the repo version but I didn't realise this at the time so added .as_expr() to make this function work. The version we actually use doesn't require .as_expr() so I've removed it


RoyPriceCS50 (approved):
Approved.

PR #1403, opened 2026-07-09
1767 - Add generic function to deal with Workplace data issues

AchyutRatkanthiwar:
## Description
Trello ticket [#1767](https://trello.com/c/k61vcYJq)

Implemented a generic function:
1. to set parent_permission to NULL when its value is 3 (legacy values from Oct–Dec 2020 in the workplace raw dataset).
2. to replace empty or whitespace-only strings with NULL across all columns imported in the clean job. 

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1767-gen-wrkplc-fixs-Transform-ASCWDS-Data:1e0b4fad-9f13-423e-96b7-4d278b20bc10)

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist

### DIFF
--- a/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
+++ b/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
@@ -98,6 +98,8 @@ def main(
     """
     lf = utils.scan_parquet(workplace_source, selected_columns=COLUMNS_TO_IMPORT)
 
+    lf = wUtils.apply_data_corrections(lf)
+
     lf = lf.filter(wUtils.valid_workplace_filter())
 
     lf = lf.with_columns(pl.col(AWPClean.nmds_id).str.strip_chars())
--- a/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
+++ b/projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
@@ -102,8 +102,6 @@ def main(
 
     lf = lf.filter(wUtils.valid_workplace_filter())
 
-    lf = lf.with_columns(pl.col(AWPClean.nmds_id).str.strip_chars())
-
     lf = lf.rename({AWPClean.last_logged_in: AWPClean.last_logged_in_date})
 
     lf = cUtils.cast_date_strings_to_dates(lf)
--- a/projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
+++ b/projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
@@ -201,3 +201,41 @@ def create_purged_lfs_for_reconciliation_and_data(
     )
 
     return ascwds_workplace_lf, reconciliation_lf
+
+
+def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
+    """
+    Apply legacy data corrections to an ASC-WDS workplace LazyFrame.
+
+    The following corrections are applied:
+        - Convert empty and whitespace-only string values to NULL across all
+          string columns.
+        - Set `parent_permission` to NULL where its value is `3`, as this is a
+          legacy value present in older data.
+
+    Args:
+        lf (pl.LazyFrame): Input LazyFrame containing ASC-WDS workplace data.
+
+    Returns:
+        pl.LazyFrame: A LazyFrame with the data corrections applied.
+    """
+    schema = lf.collect_schema()
+
+    string_columns = [name for name, dtype in schema.items() if dtype == pl.String]
+
+    lf = lf.with_columns(
+        pl.when(pl.col(col).str.strip_chars() == "")
+        .then(None)
+        .otherwise(pl.col(col))
+        .alias(col)
+        for col in string_columns
+    )
+
+    lf = lf.with_columns(
+        pl.when(pl.col(AWPClean.parent_permission) == "3")
+        .then(None)
+        .otherwise(pl.col(AWPClean.parent_permission))
+        .alias(AWPClean.parent_permission)
+    )
+
+    return lf
--- a/projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
+++ b/projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
@@ -1,4 +1,5 @@
 import polars as pl
+import polars.selectors as cs
 
 from utils.column_names.cleaned_data_files.ascwds_workplace_cleaned import (
     AscwdsWorkplaceCleanedColumns as AWPClean,
@@ -219,23 +220,10 @@ def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
     Returns:
         pl.LazyFrame: A LazyFrame with the data corrections applied.
     """
-    schema = lf.collect_schema()
+    # Treat blank strings as missing values.
+    lf = lf.with_columns(cs.string().str.strip_chars().replace("", None))
 
-    string_columns = [name for name, dtype in schema.items() if dtype == pl.String]
-
-    lf = lf.with_columns(
-        pl.when(pl.col(col).str.strip_chars() == "")
-        .then(None)
-        .otherwise(pl.col(col))
-        .alias(col)
-        for col in string_columns
-    )
-
-    lf = lf.with_columns(
-        pl.when(pl.col(AWPClean.parent_permission) == "3")
-        .then(None)
-        .otherwise(pl.col(AWPClean.parent_permission))
-        .alias(AWPClean.parent_permission)
-    )
+    # legacy parent permission temporarily contained invalid "3" codes
+    lf = lf.with_columns(pl.col(AWPClean.parent_permission).replace(3, None))
 
     return lf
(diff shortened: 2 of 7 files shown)


projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -201,3 +201,41 @@ def create_purged_lfs_for_reconciliation_and_data(
     )
 
     return ascwds_workplace_lf, reconciliation_lf
+
+
+def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
```
MarionJHolloway, line 207: When you'd said you were making a generic function, I was thinking it would be more generic (more like a wrapper around when/then).

Are we likely to need something more generic than this? I'm wondering if we want to be able to pass in columns to check and values to check for, rather than hardcoding them here. 

If these are the only two case we're likely to encounter, this is probably enough.

AchyutRatkanthiwar: yeah, I started doing it that way but AI review asked me to change it slightly because the whitespace/empty-string to NULL correction is already generic and it applies to every string column via schema introspection, no column names hardcoded and the the parent_permission to NULL correction when value is "3" seemed like very specific case. 
Happy to split them in 2 functions, one with current logic for whitespace/empty-string to NULL correction and another more generic where we can pass column name and valid values? What do you think?

MarionJHolloway: What if it took the polars expression inside when() as an argument?

Then we call is twice in the job, once for string cols and once for parent permission.

AchyutRatkanthiwar: Do you mean:
Option A:
```
def null_where(col: str, condition: pl.Expr) -> pl.Expr:
    """
    Build an expression that nulls a column's values where a condition is met.

    Args:
        col: Name of the column to correct.
        condition: Boolean expression - rows where this evaluates True have
            the column's value replaced with NULL.

    Returns:
        A Polars expression aliased to `col`.
    """
    return (
        pl.when(condition)
        .then(None)
        .otherwise(pl.col(col))
        .alias(col)
    )
def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
    schema = lf.collect_schema()
    string_columns = [name for name, dtype in schema.items() if dtype == pl.String]

    lf = lf.with_columns(
        null_where(col, pl.col(col).str.strip_chars() == "")
        for col in string_columns
    )

    lf = lf.with_columns(
        null_where(AWPClean.parent_permission, pl.col(AWPClean.parent_permission) == "3")
    )

    return lf
    
``` 
Or Option B:
```

def is_blank(col: str) -> pl.Expr:
    return pl.col(col).str.strip_chars() == ""


def equals(col: str, value: str) -> pl.Expr:
    return pl.col(col) == value

def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
    schema = lf.collect_schema()
    string_columns = [name for name, dtype in schema.items() if dtype == pl.String]

    lf = lf.with_columns(
        pl.when(is_blank(col)).then(None).otherwise(pl.col(col)).alias(col)
        for col in string_columns
    )

    lf = lf.with_columns(
        pl.when(equals(AWPClean.parent_permission, "3"))
        .then(None)
        .otherwise(pl.col(AWPClean.parent_permission))
        .alias(AWPClean.parent_permission)
    )

    return lf

Comment (MarionJHolloway) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: Kind of in between the two?
```
schema = lf.collect_schema()
string_columns: list = [name for name, dtype in schema.items() if dtype == pl.String]
string_cols_expr = pl.col(*string_columns).str.strip_chars() == ""

parent_permission_expr = pl.col(AWPClean.parent_permission) == "3"

def apply_data_corrections(lf: pl.LazyFrame, expr:pl.Expr. cols:list) -> pl.LazyFrame:Expand 


    lf = lf.with_columns(
        pl.when(expr)
        .then(None)
        .otherwise(pl.col(*cols))
        .name[...] # if you want to handle renaming
)
    return lf

lf = apply_data_corrections(lf, string_cols_expr, string_columns)
lf = apply_data_corrections(lf, parent_permission_expr, [AWPClean.parent_permission])
```

Comment (MarionJHolloway) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: Can probably refactor it a bit neater!

Comment (garypolzin) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: just throwing my thoughts in - I actually like the original

it's the most easily readable and not sure if there's any performance benefits to the refactoring?

Comment (garypolzin) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: you could potentially use selectors instead

`lf = lf.with_columns(cs.string().str.strip_chars().replace("", None))`

should caveat that with I'm not very good with selectors yet so can't say 100% that works!

Comment (AchyutRatkanthiwar) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: I tested the function using column selectors, and the existing test cases are still passing. I also asked AI for a review, and it suggested that using column selectors is a better approach than my original implementation.

Regarding whether we should split the function, I still prefer the approach I initially took because it is more direct and keeps the logic together. If we split it, apply_data_corrections would essentially just wrap calls to Polars' native when, then, and otherwise expressions, which then raises the question of whether this logic should simply be kept inline instead.

Comment (garypolzin) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: actually can you use .replace() in the second one as well?

pl.col(AWPClean.parent_permission).replace("3", None)

Comment (AchyutRatkanthiwar) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: Yea, I changed it to use .replace in both cases.

Comment (garypolzin) [reply projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py:207]: yea just on the basis that it's short and readable as is I think stick with it in the one function

File: projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py [hunk 2]
```diff
@@ -98,6 +98,8 @@ def main(
     """
     lf = utils.scan_parquet(workplace_source, selected_columns=COLUMNS_TO_IMPORT)
 
+    lf = wUtils.apply_data_corrections(lf)
+
     lf = lf.filter(wUtils.valid_workplace_filter())
 
     lf = lf.with_columns(pl.col(AWPClean.nmds_id).str.strip_chars())
```
garypolzin: is this redundant now?

projects/_01_ingest/ascwds/tests/fargate/utils/test_clean_workplace_utils.py
```diff
@@ -5,9 +5,15 @@
...
+)
 from projects._01_ingest.unittest_data.polars_ingest_test_file_data import (
     TestCreatePurgedLfsForReconciliationAndData as Data,
 )
+from projects._01_ingest.unittest_data.polars_ingest_test_file_schema import (
+    TestApplyDataCorrectionsSchemas as CorrectionsSchemas,
+)
 from projects._01_ingest.unittest_data.polars_ingest_test_file_schema import (
     TestCreatePurgedLfsForReconciliationAndDataSchemas as Schemas,
 )
```
garypolzin, line 13: we usually have one data/schema class per file, so you could rename the previous `TestCreatePurgedLfsForReconciliationAndData` to `TestCleanAscwdsWorkplaceUtilsData` and both sets of tests can go in there

projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
AchyutRatkanthiwar: yes, I will remove this

projects/_01_ingest/ascwds/tests/fargate/test_clean_ascwds_workplace.py
garypolzin, line 69: this isn't part of your PR but just spotted the duplicate entry

AchyutRatkanthiwar: yea spotted it too. removed in this PR itself as its a small change.

garypolzin: another pesky merge conflict no doubt!

projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
garypolzin: marking as resolved as it looks like it's referring to a different row of code now

garypolzin (approved):
Looks good to me

PR #1404, opened 2026-07-09
1733 slv cols

RoyPriceCS50:
## Description
Trello ticket [1733](https://trello.com/c/VGSJ8hkR)

Added a Selector to ascwds workplace cleaning that brings job role columns into the dataset.

I've defined the Selector in clean_workplace_utils because it will be needed in the validation script. The validation script will need to know all the columns for the expected schema check to pass. I could make the Selector in the job script but then the Dockerfile would need the job script as a script and a dependency which doesn't appeal to me.

Also part of this PR is refactoring remove_rows_with_duplicate_location_ids to be streaming friendly instead of using .over() which requires materialisation.

The job succeeds but the validation is failing. This is because scan_parquet infers the schema from **some** parquet files in the partitions and doesn't expand as more columns exist in **later** files.

I've made a backlog ticket to work on this with a potential solution.

## Testing
- [x] Unit tests passing
- [x] Partially Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1733-slv-cols-Transform-ASCWDS-Data:bead2a3c-e35f-4005-be84-69264b049bc6) - Fails validation because schema is not working correctly.

New job role columns exist in the dataset and are int type.

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -227,3 +237,22 @@ def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
...
+
+    Attributes:
+        jr_cols_selector (cs.Selector): Select job role columns in ascwds workplace datasets.
+    """
+
+    jr_cols_selector: cs.Selector
+
+    def __init__(self):
+        self.jr_cols_selector = (
+            cs.string() & cs.starts_with("jr") & ~cs.contains("flag", "date")
```
garypolzin: We should only need 'strt, 'stop', 'vacy' and 'emps' for now

Pretty sure the employment status cols will come from the worker file when we get round to that part (same reason as job role breakdown, NMDS probably differs in the matrix to worker records so we'll use worker records for consistency)

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -227,3 +237,22 @@ def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
     lf = lf.with_columns(pl.col(AWPClean.parent_permission).replace(3, None))
 
     return lf
+
+
+class SelectJrCols:
```
garypolzin: just a query on why it's a class too, is it overcomplicating it a bit?

can we not just have
```
def jr_cols_selector() -> cs.Selector:
    """Return a selector for ASC-WDS job role columns."""
    return cs.string() & cs.starts_with("jr") & ~cs.contains("flag", "date")
```

RoyPriceCS50: I just looked at the purge expressions and job role cleaning utils.

Two ways of achieving the same thing?

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -80,21 +80,31 @@ def valid_workplace_filter() -> pl.Expr:
     )
 
 
-def remove_rows_with_duplicate_location_ids() -> pl.Expr:
+def remove_rows_with_duplicate_location_ids(lf: pl.LazyFrame) -> pl.LazyFrame:
```
garypolzin, line 83: I know we said to add a comment to this but I'm actually thinking it's best left this way - even in streaming mode, .over() won't be great for wide datasets. So still add a comment to say why it's written like this but I don't think it's one we'd look to switch back on that basis

Something like
```
Identify duplicate keys separately rather than using a window expression.
This allows the aggregation to stream over only the key columns, avoiding
the memory overhead of `.over()` on wide datasets.
```

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -227,3 +237,14 @@ def apply_data_corrections(lf: pl.LazyFrame) -> pl.LazyFrame:
...
+
+
+def slv_cols_selector() -> cs.Selector:
+    """
+    Returns a Selector for string columns that start with 'jr' and end with
+    either 'emp', 'strt', 'stop' or 'vacy.
+    """
+
+    return (
+        cs.string() & cs.starts_with("jr") & cs.ends_with("emp", "strt", "stop", "vacy")
```
garypolzin: ahh just realised that 'temp' also ends in 'emp', so we'll have to explicitly exclude 'temp'!

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
garypolzin: Yea a class makes sense when it's grouping multiple related expressions
but this is only going to be a single selector so a function is a much simpler approach

projects/_01_ingest/ascwds/tests/fargate/utils/test_clean_workplace_utils.py
```diff
@@ -226,3 +227,43 @@ def test_function_returns_expected_values(self, case):
...
+    def test_excludes_job_role_flag_columns(self):
+        for i in self.selected_cols:
+            self.assertNotIn("flag", i)
+
+    def test_excludes_job_role_date_columns(self):
+        for i in self.selected_cols:
+            self.assertNotIn("date", i)
+
+    def test_returns_expected_column_count(self):
+        self.assertEqual(len(self.selected_cols), 4)
```
garypolzin: I think one simple test is fine here, some of them look to be testing the previous functionality (eg test excludes 'flag' and 'date' are not overly relevant now and the function doesn't mention those by name)
```
    def test_selects_expected_columns(self):
        test_lf = (fine as it is)

        returned_cols= test_lf.select(job.slv_cols_selector()).collect_schema().names()

        expected_cols = [
            AWPClean.job_role_01_employees,
            AWPClean.job_role_01_starters,
            AWPClean.job_role_01_leavers,
            AWPClean.job_role_01_vacancies,
        ]

        self.assertEqual(returned_cols, expected_cols)
```

garypolzin: actually we don't test the cs.string() part currently, something like this should do

```
def test_ignores_non_string_job_role_columns(self):
    test_lf = pl.LazyFrame(
        {
            "jr_int_strt": [1],     # int
            "jr_str"strt": ["1"],  # string
        }
    )

    returned_cols= test_lf.select(job.slv_cols_selector()).collect_schema().names()

    self.assertEqual(returned_cols, ["jr_str"strt"])
```

garypolzin (commented):
I think it's best to hash out the validations which are failing so it completes successfully - that way we'll know if something which currently works starts to break

Couple of comments on tests following changes

RoyPriceCS50:
Job and validation succeed now

garypolzin (approved):

PR #1411, opened 2026-07-14
1775 - Create CLAUDE.md file

garypolzin:
A CLAUDE.md file is a standing instructions document for Claude Code. It acts as an "AI README," holding your project's tech stack, coding conventions, and behavioral rules, which it reads automatically at the start of every session so you don't have to keep re-explaining yourself

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,6 +22,22 @@ Datasets here are large in both rows and columns, and OOM is a recurring real pr
 - Prefer `float32` over `float64` where precision requirements allow — call out explicitly when a trade-off matters (e.g. long aggregations/sums where error can accumulate).
 - If a lazy chain is unintentionally broken (an early `.collect()`, a `.to_pandas()`, list-comprehension row-wise logic), flag it — that's a correctness-adjacent issue here, not a style nitpick. Some `.collect()` are ok as they directly follow an aggregation which results in a very small LazyFrame.
 
+### Polars streaming engine — what isn't covered yet
+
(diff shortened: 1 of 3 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,103 @@
...
+│   ├── utils/               # helpers shared across stages within this project
+│   └── unittest_data/       # test fixtures shared across stages within this project
+└── utils/                   # helpers shared across projects
+```
+
+Place a helper at the narrowest scope that covers its actual usage (stage → project → repo-wide) — check whether something similar already exists at that scope before adding a new one.
+
+## Column names
+
+Never hardcode column-name strings. Each project keeps its own column-name class near the project root, e.g. `projects/_04_direct_payment_recipients/direct_payments_column_names.py` defining `DirectPaymentColumnNames`, imported under a short alias:
```
MarionJHolloway: Most of these are in utils - do we want to migrate them?

MarionJHolloway (commented):
Do we want to add a reference to the polars streaming list on github, something about using this a a reference for which functions might help with OOM issues and which might be causing problems?

AchyutRatkanthiwar:
> Do we want to add a reference to the polars streaming list on github, something about using this a a reference for which functions might help with OOM issues and which might be causing problems?

If claude is able to read it every time we start a session, can we give it the link to "Tracking issue for the new streaming engine" which has list of what is in streaming now?

garypolzin:
@MarionJHolloway @AchyutRatkanthiwar it can't do it directly but it's suggested a more basic way of doing it. Looks like the last update was Jan 2025 so it's not a fast changing list so this more static approach should be fine. We can always update manually if we notice something in between

CLAUDE.md
garypolzin: I've made a backlog ticket

AchyutRatkanthiwar (approved):
Looks good.

PR #1412, opened 2026-07-14
1742 Select columns from clean workspace for SLV

MarionJHolloway:
## Description
Trello ticket [#1742](https://trello.com/c/KrrDwQV9/1742-s-mc1-programmatically-make-a-list-of-the-columns-required-to-load-into-the-merge-job)

Select columns from clean workplace data
I have ignored the request for a schema as the strings have already been converted to integers in the clean workplace job.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-2%3A856699698263%3AstateMachine%3A1742-mk-col-ls-Ind-CQC-SLV?type=standard)

Screenshots in trello ticket
I had to adjust the dataset we read in to the polars one and save out the workspace data temporarily to get these outputs checked. I've reverted the code back to point to the main ascwds clean job and save the job role estimates for validation purposes.

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -254,9 +255,4 @@ def slv_cols_selector() -> cs.Selector:
      - do not end with 'temp'
     """
 
-    return (
-        cs.string()
-        & cs.starts_with("jr")
-        & cs.ends_with("emp", "strt", "stop", "vacy")
-        & ~cs.ends_with("temp")
-    )
+    return cs.string() & expr.is_slv_job_role_column()
```
garypolzin: feels risky having the docstring logic for what in expr.is_slv_job_role_column() in this function - if we ever change the is_slv_job_role_column() expression then we'd never think to change this docstring

personally I wouldn't bother with this function at all, I'd just put `cs.string() & expr.is_slv_job_role_column()` straight into the job

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/utils/merge_utils.py
```diff
@@ -1,10 +1,21 @@
...
-    Placeholder function to create a list of columns for the ASCWDS dataset."""
-    pass
+    Returns a Selector for columns that are:
+     - numeric datatype
+     - start with 'jr'
+     - end with either 'emp', 'strt', 'stop' or 'vacy
+     - do not end with 'temp'
+    """
+
+    return cs.numeric() & expr.is_slv_job_role_column()
```
garypolzin: as above, I'd completely get rid of this - depending on how you use it you might not need cs.numeric() either (see reason for cs.string() inclusion above)

just import expr.is_slv_job_role_column() straight into the job

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
garypolzin: the reason for `cs.string()` part of this was because we cast to an int, so it checked it was a string before casting and if it wasn't, it didn't cast it - if there's no benefit to identifying whether it's string or numeric then we don't need to bother

tests/test_polars_utils/test_utils.py
```diff
@@ -6,6 +6,7 @@
 import tempfile
 import unittest
 from contextlib import redirect_stdout
+from dataclasses import dataclass
```
garypolzin: I'm guessing these 3 new imports can be removed

CHANGELOG.md
```diff
@@ -72,7 +72,7 @@ All notable changes to this project will be documented in this file.
 
 - National care work thresholds removed from job role estimates validation.
 
-- Pull metadata columns into merge job within SLV pipeline.
+- Pull metadata and cleaned workplace columns into merge job within SLV pipeline.
```
garypolzin: now we've git tagged you'll need to add this step as a new bullet point in it's own right

garypolzin (commented):
Added a couple of comments :)

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
MarionJHolloway: I don't think we need the string check then, that makes it easier to simplify! Casting a numeric type to an int wouldn't be a problem.

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_01_merge.py
```diff
@@ -41,14 +50,17 @@ def main(
...
         source=metadata_source, selected_columns=metadata_columns
     )
     job_role_estimates_lf = utils.scan_parquet(job_role_estimates_source)
-    cleaned_ascwds_workplace_lf = utils.scan_parquet(cleaned_ascwds_workplace_source)
+    cleaned_ascwds_workplace_lf = utils.scan_parquet(
+        cleaned_ascwds_workplace_source
+    ).select(
+        *[AWPClean.establishment_id, AWPClean.ascwds_workplace_import_date],
+        expr.is_slv_job_role_column()
+    )
```
garypolzin, line 63: happy for this to be a separate PR if you want but if we apply a filter here to only keep rows with a non-null locationid it should get rid of about a quarter of the data

garypolzin (approved):
Looks good - one minor comment which happy for you to put in this PR or a separate one

RoyPriceCS50:
I've replaced the selector expression with a function that returns a schema in my branch 1731. It'll clash with these changes.

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_01_merge.py
MarionJHolloway: Yeah, I'd need to rerun everything in AWS which involves making changes and reversing them (see the PR description). I'll put a ticket in the backlog instead. Once the data is actually merged, this will be an easier change.

MarionJHolloway:
Discussed changes with @RoyPriceCS50 . Agreed to merge this PR first as he is changing direction with his ticket.

PR #1418, opened 2026-07-16
Update Claude.md .over() usage

garypolzin:
Update CLAUDE.md following real world usage of .over() compared to group-by+join


MarionJHolloway (approved):
Looks good. My only suggestion would be that it might be useful to ask Claude to be specific about the reason .over() was chosen or not in the output by default, so that we can learn from some examples. We could then remove that request in a few weeks/ months as we learn more? 

Obviously we could do this individually, but I could see it being a useful standard in the short term, perhaps in code review?

PR #1419, opened 2026-07-17
1807 - Create discover_combined_schema function

AchyutRatkanthiwar:
## Description
Trello ticket [#1807](https://trello.com/c/94easVG1/1807-bring-job-roles-higher-than-41-into-the-cleaning-dataset-create-discovercombinedschema)

To bring job roles higher than 41 into the cleaning dataset, I have created a new function to discover combined schema from all partitions in the dataset. 

## Testing
- [x] Unit tests passing

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist
- [x] PR solves the Trello ticket requirements
- [x] Outputs appear reasonable and understandable
- [x] The overall approach is appropriate and not over-engineered
- [x] No obvious scalability, performance or interdependency concerns
- [x] Tests appropriately cover the main behaviour/edge cases
- [x] Naming and structure are broadly understandable and consistent
- [x] Docstrings are sufficient for future maintenance
- [x] Docstrings cover non-obvious behaviour


tests/test_polars_utils/test_utils.py
```diff
@@ -445,6 +445,57 @@ def test_bucket_and_prefix_parsing(self, mock_boto_client_mock: Mock):
...
+        def scan_parquet_side_effect(path):
+            mock_lf = MagicMock()
+            mock_lf.collect_schema.return_value = schemas_by_path[path]
+            return mock_lf
+
+        mock_scan_parquet.side_effect = scan_parquet_side_effect
+
+        result = utils.discover_combined_schema("s3://test_bucket/domain=x/dataset=y/")
+
+        self.assertEqual(
```
MarionJHolloway: Use pytest format, not unit test, e.g.:

assert a == b

tests/test_polars_utils/test_utils.py
```diff
@@ -445,6 +445,57 @@ def test_bucket_and_prefix_parsing(self, mock_boto_client_mock: Mock):
...
+            result, pl.Schema({"a": pl.String, "b": pl.String, "c": pl.String})
+        )
+
+    @patch(f"{PATCH_PATH}.boto3.client")
```
MarionJHolloway, line 488: Can we add a test of the output here?

tests/test_polars_utils/test_utils.py
```diff
@@ -445,6 +445,57 @@ def test_bucket_and_prefix_parsing(self, mock_boto_client_mock: Mock):
         self.assertEqual(result, [])
 
 
+class TestDiscoverCombinedSchema(unittest.TestCase):
```
MarionJHolloway: Remove inherited unittest.TestCase here

tests/test_polars_utils/test_utils.py
```diff
@@ -445,6 +445,57 @@ def test_bucket_and_prefix_parsing(self, mock_boto_client_mock: Mock):
...
+            result, pl.Schema({"a": pl.String, "b": pl.String, "c": pl.String})
+        )
+
+    @patch(f"{PATCH_PATH}.boto3.client")
+    def test_raises_when_no_parquet_files_found(self, mock_boto_client: Mock):
+        mock_paginator = MagicMock()
+        mock_paginator.paginate.return_value = [{"Contents": []}]
+        mock_boto_client.return_value.get_paginator.return_value = mock_paginator
+
+        with self.assertRaises(FileNotFoundError):
```
MarionJHolloway: Pytest syntax for testing exceptions is here: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

polars_utils/utils.py
```diff
@@ -290,6 +293,58 @@ def list_s3_parquet_import_dates(s3_prefix: str) -> list[int]:
     return sorted(dates)
 
 
+def discover_combined_schema(source: str) -> pl.Schema:
+    """
```
MarionJHolloway, line 297: This doc string reads more like a sales pitch! I'm guessing Claude wrote it? Suggest we change it to (may need reformatting!):

 ```
Builds the union of every column found across all parquet files under a
    partitioned S3 dataset.

    Generates a full schema for wide datasets where columns get added or dropped between partitions over time.  `pl.scan_parquet` alone only infers columns from the first file it encounters, silently missing anything added later. This wrapper reads each file's schema (the parquet footer), not a full data scan. 

    Each file's footer is read in paralel on its own thread - these are independent,
    I/O-bound S3 calls, therefore runtime is close to the cost of one
    read rather than multiple back-to-back reads.

    Args:
        source (str): S3 directory containing the partitioned parquet dataset.

    Returns:
        pl.Schema: The union of every column and dtype found across all files.

    Raises:
        FileNotFoundError: If there are no parquet files in the source directory.
```

MarionJHolloway (changes requested):
I've suggested changes to the docstring, feel free to improve further.

Also, the unit tests for this are written for unittest, not pytest, so this will need a few tweaks.

tests/test_polars_utils/test_utils.py
AchyutRatkanthiwar: Sorry, what do you mean?

MarionJHolloway (approved):

PR #1430, opened 2026-07-22
1731 fix legacy job roles

RoyPriceCS50:
## Description
Trello ticket [1731](https://trello.com/c/f5VzPY7N)

Add function in ascwds cleaning util that fixes old job roles SLV data by merging the old job role columns into ones that currently exist.
The old job role columns are then dropped.

Job role 33 is hard coded to be dropped without moving values because it's personal assistants. Doc string highlights this.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1731-fix-old-jr-Transform-ASCWDS-Data:d70cb6cb-0b23-4c8d-9630-7fcd70754dd4)

See Trello ticket Excel attachment.
Got the cleaned ascwds workplace data from main and summed job roles affected by merging, and got jr28 to check a column not affected by merging.

Manually summing job roles gave same as branch output.

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


utils/column_names/raw_data_files/ascwds_workplace_columns.py
```diff
@@ -305,6 +305,17 @@ class AscwdsWorkplaceColumns:
...
+    job_role_33_employees: str = "jr33emp"
+    job_role_33_flag: str = "jr33flag"
+    job_role_33_leavers: str = "jr33stop"
+    job_role_33_other: str = "jr33oth"
+    job_role_33_permanent: str = "jr33perm"
+    job_role_33_pool: str = "jr33pool"
+    job_role_33_starters: str = "jr33strt"
+    job_role_33_temporary: str = "jr33temp"
+    job_role_33_vacancies: str = "jr33vacy"
+    job_role_33_workers: str = "jr33work"
```
garypolzin: I wouldn't bother adding these to this list - we're dropping them so we'll never need to reference them by name

If we add anything to this list it should probably be job roles 46-52, but the main purpose of this is for when we want to use them in code, and so far we've used generic selectors instead so I don't think they need adding

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -279,3 +282,77 @@ class BoundingExpressions:
...
+) -> pl.LazyFrame:
+    """
+    Merge ASC-WDS workplace job role columns.
+
+    The ASC-WDS workplace job role columns corresponding to each key
+    and all its values from given dict are summed, then all value columns
+    are dropped.
+
+    Job role 33 (personal assistant) is hard coded to be dropped without
+    merging their values.
```
garypolzin: Including this within this function feels a bit hidden, I'd be tempted to have a separate function before it in the main job to do that part

it's small enough to be inline but I think it's worth it's own function to add a bit of context

```
def drop_personal_assistant_columns(lf):
    """very brief explanation why"""
    return workplace_lf.drop(cs.starts_with("jr33")
```

projects/_01_ingest/ascwds/tests/fargate/test_clean_ascwds_workplace.py
```diff
@@ -76,6 +80,8 @@ def test_main_runs(
         create_purge_date_columns_mock.assert_called_once()
         remove_rows_with_duplicate_location_ids_mock.assert_called_once()
         discover_combined_schema_mock.assert_called_once_with(self.WORKPLACE_SOURCE)
+        assert is_slv_job_role_column_mock.call_count == 2
+        fix_legacy_job_roles_mock.assert_called_once()
```
garypolzin: (optional) could firm this up a bit with `.assert_called_once_with()`?

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ All notable changes to this project will be documented in this file.
 - Get Workplace data schema from `discover_combined_schema` within Clean Workplace Job. Updated tests for the same.
 
 ### Fixed
-- Fixed the Transform ASCWDS Data pipeline, which was failing due to an incorrect dataset name in Terraform and the clean workplace job dropping the `import_date` column that the clean worker job depends on. Corrected the Terraform dataset name and removed the drop statement for `import_date`. 
+- Fixed the Transform ASCWDS Data pipeline, which was failing due to an incorrect dataset name in Terraform and the clean workplace job dropping the `import_date` column that the clean worker job depends on. Corrected the Terraform dataset name and removed the drop statement for `import_date`.
```
garypolzin, line 31: merge conflict?

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -279,3 +282,77 @@ class BoundingExpressions:
         .otherwise(slv_bounding_cols)
         .name.keep()
     )
+
+
+def fix_legacy_job_roles(
+    lf: pl.LazyFrame, legacy_job_roles_dict: dict[str, list[str]]
+) -> pl.LazyFrame:
+    """
+    Merge ASC-WDS workplace job role columns.
```
garypolzin: 'merge' in the docstring is a better description of this - it's not broken

```
def merge_legacy_job_roles():
    """
    Merge legacy ASC-WDS workplace job role columns into their current equivalents.
    ...
```

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -279,3 +282,77 @@ class BoundingExpressions:
...
+
+    The ASC-WDS workplace job role columns corresponding to each key
+    and all its values from given dict are summed, then all value columns
+    are dropped.
```
garypolzin: we haven't said/shown what the key and value actually are which makes this a bit tricky to understand. Describing in words might be clearer?

```
    For each new job role code in legacy_job_roles_dict, sums the new job role's
    columns together with the legacy columns being merged into them. The legacy
    codes' columns are then dropped.
```

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
```diff
@@ -279,3 +282,77 @@ class BoundingExpressions:
...
+        legacy_replacement_expressions(legacy_job_roles_dict, job_role_suffixes),
+    )
+
+    old_roles = [old for olds in legacy_job_roles_dict.values() for old in olds]
+    cols_to_drop = [
+        f"jr{role}{suffix}"
+        for role in old_roles + ["33"]
+        for suffix in job_role_suffixes
+    ]
+    lf = lf.drop(cols_to_drop)
```
garypolzin: I don't think you need suffix here, can get rid of everything based on jr{number} alone

`lf = lf.drop(cs.starts_with(*[f"jr{role}" for role in old_roles]))`

garypolzin (commented):
Looks good - I'd take the '33' out of that function and into one on it's own, it's a bit hidden where it is

Most other comments are minor wording changes

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
RoyPriceCS50: Or do this in line, with a comment next to it?   Function means testing a wrapper around standard code.

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
RoyPriceCS50: I think the idea of new and 'legacy' is not necessary to understand what's happening. We're merging columns together and dropping ones we don't want anymore.

I'll re-word.

We're going to do the same later on in prepare step of slv step function. Reducing to 11 published roles.

projects/_01_ingest/ascwds/fargate/clean_ascwds_workplace.py
```diff
@@ -150,6 +158,11 @@ def main(
         *[AWPClean.establishment_id, AWPClean.import_date],
         expr.is_slv_job_role_column(),
     )
+
+    # Personal assistant data was collected in the NMDS-SC up to Oct 2017.
+    # This role has a dedicated survey, therefore NMDS-SC data is not required.
```
garypolzin: for simplicity I'd refer to it as ASC-WDS here

projects/_01_ingest/ascwds/fargate/utils/clean_workplace_utils.py
garypolzin: @RoyPriceCS50 so are we planning on using this function as it is for that step as well?

If so, is it worth some name changes to make it completely generic now? Replace references to 'new' and 'legacy' and change to something like 'role to keep' and 'roles to merge and drop' and the general concept of grouping multiple columns together into one combined role which we keep and drop the rest? Those suggested names don't feel great but along those lines

RoyPriceCS50: Moved it to polars utils cleaning utils.

polars_utils/cleaning_utils.py
```diff
@@ -275,3 +276,69 @@ def cast_date_strings_to_dates(
...
+        legacy_replacement_expressions(job_role_mapping, job_role_suffixes),
+    )
+
+    old_roles = [old for olds in job_role_mapping.values() for old in olds]
+    lf = lf.drop(cs.starts_with(*[f"jr{role}" for role in old_roles]))
+
+    return lf
+
+
+def legacy_replacement_expressions(
```
garypolzin: nice, I think it's just this function which needs a similar amendment to remove legacy/old/new then all good

tests/test_polars_utils/test_cleaning_utils.py
```diff
@@ -389,3 +389,46 @@ def test_column_is_not_converted_if_already_date_type(self):
...
+            AWPClean.job_role_03_employees: 12,
+            AWPClean.job_role_06_employees: 6,
+            AWPClean.job_role_01_starters: 30,
+            AWPClean.job_role_03_starters: 120,
+            AWPClean.job_role_06_starters: None,
+            "not_a_job_role_column": "A",
+        }
+    )
+
+    def test_function_returns_expected_data(self):
```
garypolzin: with the job group function in mind, can we have a test which covers the case where the 'new' column doesn't exist yet. For example we'll be making an 'other DC' column but that won't exist before this is run

RoyPriceCS50: Refactored the tests to pytest style and Copilot recommended some extra checking.

garypolzin (approved):
Looks good now thanks!

PR #1434, opened 2026-07-23
1796 Reduce cleaned workplace lazyframe to published roles in slv prep

RoyPriceCS50:
## Description
Trello ticket [1796](https://trello.com/c/MLY0CvhG)

Called cleaning util merge_job_role_columns in slv prep job script.
Updated job test script to reflect function call.

I had to change the merge function to prevent dropping job role columns with similar chars in name. 
Targeting jr10emp would also drop jr101emp.
I added a test for this.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1796-less-roles-Ind-CQC-SLV:90f630f4-f57b-4b8c-97d2-331103ebb6f7)

See Excel attached to Trello ticket.
First tab shows how job roles are mapped.
Second tab compares manually summing job roles in workplace_cleaned to slv_prep output. There's zero difference.

- [x] Updated CHANGELOG

## Reviewer checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ All notable changes to this project will be documented in this file.
 
 - Renamed 'slv' datasets in AWS so they are grouped together in alphabetical order.
 
+- Called merge_job_role_columns in slv prepare job.
+
 ### Fixed
 - Fixed the Transform ASCWDS Data pipeline, which was failing due to an incorrect dataset name in Terraform and the clean workplace job dropping the `import_date` column that the clean worker job depends on. Corrected the Terraform dataset name and removed the drop statement for `import_date`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@ All notable changes to this project will be documented in this file.
 
 - Renamed 'slv' datasets in AWS so they are grouped together in alphabetical order.
 
-- Called merge_job_role_columns in slv prepare job.
+- Called merge_job_role_columns in slv prepare job to reduce job role columns to only published roles plus
+  'other direct care/manager/etc'
 
 ### Fixed
 - Fixed the Transform ASCWDS Data pipeline, which was failing due to an incorrect dataset name in Terraform and the clean workplace job dropping the `import_date` column that the clean worker job depends on. Corrected the Terraform dataset name and removed the drop statement for `import_date`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,10 +37,6 @@ All notable changes to this project will be documented in this file.
 
 - Reduced the SLV prepare step further to one file per calendar month, matching the granularity the downstream job role estimates dataset already uses. Generalised the existing earliest-file-per-month reduction into a shared `earliest_file_per_month_filter_expr` in Polars Utils (alongside `reduced_data_filter_expr`) so both pipelines use the same definition, and updated its existing callers in the independent CQC clean job accordingly.
 
-- Widened the ASC-WDS file-arrival polling in the orchestrator step function to check every 15 minutes for up to 15 hours on the main workspace, to accommodate ASC-WDS data now arriving much later than before, and made the polling interval and attempt count workspace-configurable so non-main workspaces poll every 10 seconds for up to 100 seconds instead.
-
-- Reduced the rows carried through the SLV prepare step to the same retention window the downstream job role estimates already use, and moved `reduced_data_filter_expr` into Polars Utils so both pipelines share one definition of that window.
-
 - Called merge_job_role_columns in slv prepare job to reduce job role columns to only published roles plus
   'other direct care/manager/etc'
 
--- a/polars_utils/cleaning_utils.py
+++ b/polars_utils/cleaning_utils.py
@@ -302,9 +302,15 @@ def merge_job_role_columns(
         {re.sub(r"^jr\d+", "", col) for col in job_role_cols if col.startswith("jr")}
     )
 
+    print("column names before merging:")
+    print(lf.collect_schema().names())
+
     lf = lf.with_columns(
         merge_job_roles_expressions(job_role_mapping, job_role_suffixes),
     )
+
+    print("column names after merging:")
+    print(lf.collect_schema().names())
 
     old_roles = [old for olds in job_role_mapping.values() for old in olds]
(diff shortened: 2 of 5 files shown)


projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
```diff
@@ -1,22 +1,39 @@
...
+import polars_utils.cleaning_utils as cUtils
 import projects._07_workforce_characteristics._01_starters_leavers_vacancies.fargate.utils.prepare_utils as pUtils
 from polars_utils import utils
-from polars_utils.cleaning_utils import apply_categorical_labels
+
+unpublished_roles_mapping = {
+    "101": ["02", "03", "05", "24", "45", "47", "49", "50"], # other managers
+    "102": ["35", "37"], # other regulated professions
+    "103": ["10", "11", "23", "38"], # other direct care
+    "104": ["25", "26", "27", "34", "36", "39", "40", "42", "44", "46", "48", "51"], # other
```
garypolzin: Trying to think if there's a benefit to introducing new numbers here or whether we switch to text - my concern is that this is the only place where it tells us that 101='other managers' for example

@MarionJHolloway - not sure if either approach helps/hinders/overlaps with what you're currently looking at (#1814)?

RoyPriceCS50: I see what you mean.

The polars version of apply_categorical_labels uses the csv. That doesn't have any job role codes or labels in it at the moment. 

I think it's easier to add "101", "102" etc to the csv when the time comes.

The job role column merging function is build around changing "10" to "jr10emp". This prevents the input mapping having to be explicit "jr01emp", "jr01strt"... and so on.

MarionJHolloway: The restructuring using the column name strings, so we would need a code for these groups, unless we renamed all the columns first and then restructured the data. Maybe we should wait until the PR for mine is ready and consider them together (or even test in a new branch together).

RoyPriceCS50: If we renamed them first, they'd still need to distinguish between emps, starters etc, so there not really job roles at that stage.

Having them in the csv means we're committing the job roles codes for these made up roles. Same way we fix that job role 01 refers to the same thing everywhere.

Scratching my head about how we'd label this one in the csv though. We already have other direct care and other non direct care. This one would be "other - other" that includes the other others.

MarionJHolloway: I'm not suggesting that we can't have job role codes for these, but when we come to doing the applying labels ticket, there's a question of when and how to do that. Perhaps we could accept this for now but review as we're doing the other ticket? Would that work for you @garypolzin

garypolzin: yea I just thought I'd check this bit first - happy either way

I haven't properly looked a the PR, just spotted that bit so wanted to flag it before I finish

garypolzin (commented):
just an FYI that you have merge conflicts in 2 files

polars_utils/cleaning_utils.py
```diff
@@ -307,7 +307,10 @@ def merge_job_role_columns(
     )
 
     old_roles = [old for olds in job_role_mapping.values() for old in olds]
-    lf = lf.drop(cs.starts_with(*[f"jr{role}" for role in old_roles]))
+    roles_to_drop = [
```
MarionJHolloway, line 293: Can we have a comment to make this easier to scan please?

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
```diff
@@ -1,22 +1,39 @@
+import polars.selectors as cs
+
+import polars_utils.cleaning_utils as cUtils
 import projects._07_workforce_characteristics._01_starters_leavers_vacancies.fargate.utils.prepare_utils as pUtils
 from polars_utils import utils
-from polars_utils.cleaning_utils import apply_categorical_labels
+
+unpublished_roles_mapping = {
+    "101": ["02", "03", "05", "24", "45", "47", "49", "50"], # other managers
```
MarionJHolloway: Do we want to go higher here? 996-999 maybe?

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
```diff
@@ -1,22 +1,39 @@
...
     """
     workplace_lf = utils.scan_parquet(cleaned_ascwds_workplace_source)
 
-    # TODO: 1796 - Placeholder only.
-    # pUtils.reduce_to_published_roles()
+    workplace_lf = cUtils.merge_job_role_columns(
+        workplace_lf, unpublished_roles_mapping
+    )
+
+    # These columns refer to overall and job groups.
```
MarionJHolloway: "These columns refer to the toal for all job roles (28) and the total for job groups (29-32)."

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ All notable changes to this project will be documented in this file.
 
 - Renamed 'slv' datasets in AWS so they are grouped together in alphabetical order.
 
+- Called merge_job_role_columns in slv prepare job.
```
MarionJHolloway: Give the reason here as well.

MarionJHolloway (changes requested):
A few small changes

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
RoyPriceCS50: I think 1001, 1002 etc.  What if we require a fifth group in the future?  Easier to say 4 digits refers to groups.

MarionJHolloway (approved):
Looks good, thanks :)

PR #1435, opened 2026-07-24
1816 - Create initial set of Claude Code skills

garypolzin:
## Description
Trello ticket [#1816](https://trello.com/c/ViDdW6AX/1816-update-claudemd)

Adds `.claude/skills/`
**Skills** are on-demand instruction sets that Claude Code loads automatically when a task matches (e.g. "migrate this to polars"), instead of relying on it to recall/re-derive our conventions each time from CLAUDE.md alone.

The benefit is consistent, repo-specific behaviour (checklists, naming, memory/streaming judgement) applied automatically on the matching task, without re-explaining it every conversation. Trimmed the equivalent CLAUDE.md sections to short pointers so detail lives in one place instead of drifting across two.

8 skills added, covering our most repeated processes:
- polars-migration
- polars-streaming-check
- over-vs-join
- pytest-pattern
- review-checklist
- new-ticket — branch check → short interview → hands off to Plan Mode for the SPEC
- commit-push — stage, derive `<ticket> - <Description>` commit message, push
- open-pr — pick the right PR template, run an AI review (blocking on Critical findings), create the PR

**Claude.md**
Also added the following from trello card
- branch names must be 16 chars or less
- all new code and full unit test suite is manually run on full data in branch using CICD pipeline
- Improve current guidance on the scope
- New "Output style" section: keep responses concise, ask one interview question at a time
- New ticket-workflow convention

## Checklist
- [X] Updated CHANGELOG
- [X] Code reviewed by AI


CLAUDE.md
```diff
@@ -9,11 +9,28 @@ Skills for Care Workforce Intelligence Team. Builds reproducible AWS data pipeli
...
+  - Editing one function/method → upgrade just that function's docstring and its corresponding test(s).
+  - Editing a whole class → upgrade the whole class (all its methods, docstrings, and its test class).
+  - Editing a whole script/module → upgrade the whole file's docstrings and its test file.
+  - Editing at folder level (e.g. migrating a full pipeline stage) → upgrade everything within that folder.
+- Don't upgrade sibling functions/files you haven't otherwise touched just because they're nearby and old-format — that's still a mass-rewrite, just a smaller one.
+
+## Environment & workflow
+
+- Run `pipenv shell` (or prefix commands with `pipenv run`) to access this project's installed packages — don't assume they're on the system/global Python.
+- The `gh` CLI is not installed in this environment. Don't shell out to `gh`; if a GitHub API action is genuinely needed, say so explicitly rather than assuming it's available.
```
MarionJHolloway: I've managed to get this working now, so might not be needed. I think we could also add skills for /new-ticket ,  /commit-push and /open-pr to cover the shell commands plus how we want them structured consistantly (using worktrees/ templates/ consice/ descriptive. Can make this another ticket if you like!

Edit - I've made these for my own workflow, happy to push to this branch if you want to test them out. It's hard to know how much is just how I like to work and how much is suitable for the team

CLAUDE.md
```diff
@@ -9,11 +9,28 @@ Skills for Care Workforce Intelligence Team. Builds reproducible AWS data pipeli
...
+- Branch names must be **16 characters or fewer**.
+- CI/CD automatically runs all new code plus the full unit test suite against full data once pushed to a branch. Local/manual validation should run the relevant unit tests and `terraform validate` where applicable — never attempt to deploy or run code against AWS directly; that's the CI/CD pipeline's job.
 
 ## Scale is the top constraint
 
-Datasets here are large in both rows and columns, and OOM is a recurring real problem. When writing or reviewing Polars code, correctness and "idiomatic" style are necessary but not sufficient — always also check for unnecessary materialisation. Concretely:
+Datasets here are large in both rows and columns — typically 4M+ rows and 20+ columns, as an order-of-magnitude anchor rather than a hard spec — and OOM is a recurring real problem. When writing or reviewing Polars code, correctness and "idiomatic" style are necessary but not sufficient — always also check for unnecessary materialisation. Concretely:
```
MarionJHolloway: 20-500 columns?

Also worth noting how much of the data is null/ datatypes that affect storage space

CLAUDE.md
```diff
@@ -9,11 +9,28 @@ Skills for Care Workforce Intelligence Team. Builds reproducible AWS data pipeli
...
+  - Editing a whole script/module → upgrade the whole file's docstrings and its test file.
+  - Editing at folder level (e.g. migrating a full pipeline stage) → upgrade everything within that folder.
+- Don't upgrade sibling functions/files you haven't otherwise touched just because they're nearby and old-format — that's still a mass-rewrite, just a smaller one.
+
+## Environment & workflow
+
+- Run `pipenv shell` (or prefix commands with `pipenv run`) to access this project's installed packages — don't assume they're on the system/global Python.
+- The `gh` CLI is not installed in this environment. Don't shell out to `gh`; if a GitHub API action is genuinely needed, say so explicitly rather than assuming it's available.
+- Branch names must be **16 characters or fewer**.
+- CI/CD automatically runs all new code plus the full unit test suite against full data once pushed to a branch. Local/manual validation should run the relevant unit tests and `terraform validate` where applicable — never attempt to deploy or run code against AWS directly; that's the CI/CD pipeline's job.
```
MarionJHolloway, line 32: Something along the lines of: When full scale testing is needed, ask the user questions about the results or request output files?

.claude/skills/review-checklist/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+
+- The overall approach is appropriate and not over-engineered.
+- No obvious scalability, performance, or interdependency concerns.
+- Tests appropriately cover the main behaviour and edge cases (see `pytest-pattern` skill for what "appropriate" looks like here).
+- Naming and structure are broadly understandable and consistent with the rest of the repo.
+- Docstrings are sufficient for future maintenance and cover non-obvious behaviour.
+
+## Validating a change
+
+If asked to validate before/during review, see CLAUDE.md's "Environment & workflow" section for what's in scope (tests + `terraform validate`, never deploying or running against AWS) and what tooling isn't available (`gh` CLI).
```
RoyPriceCS50: Does the part about never deploying or running against AWS want to be a "hook" ? 

Training says Claude can occasionally skip stuff in the claude.md file. And if its in a skills that's only accessed when required, then it's more prone to be ignored?

garypolzin: So this isn't an instruction telling it never to do it because we don't want it to

It's more to stop it trying using that approach to validate a PR on the basis that it can't do it but it might try to - essentially it's to stop the loop of it trying and failing each time

CLAUDE.md
garypolzin: the idea was quite generic guidance to quantify what a minimum sized 'large' dataset refers to for a bit of context

null/datatype stuff will change with every dataset/job so I guess they will be more task specific as opposed to standard info?

CLAUDE.md
garypolzin: yea there will be a lot of change early on whilst we get stuff set up - can push as separate PR's

I'm not too bothered about these being perfect initially, we'll tweak accordingly once we start using them

so shall I remove the 'gh' CLI not installed bulletpoint?

CLAUDE.md
MarionJHolloway: True, in which case do we want to give a maximum constraint? e.g.:

Datasets here can be large in both rows and columns — maximum size is typically 4M+ rows and 200+ columns, as an order-of-magnitude anchor rather than a hard spec, but can vary depending on the project and task/job —

garypolzin: given this is general guidance I think we're best catering for the typical dataset size which is what 20+ was meant to convey

it's worth flagging explicitly if the task/job you're doing differs considerably from the norm, but I don't think it should be the default here

CLAUDE.md
garypolzin: similar vibe to the last one, I think this is more of a specific scenario rather than something the default guidance needs to cover. If full-scale results are genuinely needed, it's worth flagging that when setting up the task rather than making it into the general rule here

Or we could mention it in a skill if there's a specific task which does that more frequently

CLAUDE.md
garypolzin: Removed references to numbers - left as just 'large'

CLAUDE.md
garypolzin: created a new trello ticket to make sure we all have gh CLI working

.claude/skills/new-ticket/SKILL.md
```diff
@@ -0,0 +1,32 @@
+---
+name: new-ticket
+description: Use when starting a new piece of work — "/new-ticket", "start a new ticket", "let's start on card #1234". Runs branch check → short interview → hands off to Plan Mode for the SPEC. Pairs with the `commit-push` and `open-pr` skills for the rest of the cycle.
```
RoyPriceCS50: let's start on ticket #1234

.claude/skills/open-pr/SKILL.md
```diff
@@ -0,0 +1,20 @@
...
+
+1. Confirm the branch is pushed and up to date with remote — run `commit-push` first if not.
+2. Run a sub-agent review of the diff against `main`, following the `review-checklist` skill. This is what satisfies the "Code reviewed by AI" checklist item on the PR template — don't tick it without actually running this.
+   - Any **Critical** finding: stop and resolve or discuss with the user before continuing.
+   - Important/Optional findings: show them, then continue regardless.
+3. Pick the template from `.github/PULL_REQUEST_TEMPLATE/`:
+   - `polars_migration_template.md` if the diff contains a `# converted to polars ->` marker.
+   - `standard_template.md` otherwise.
+4. Populate it:
+   - Trello link — only the ticket number is embedded in the branch name, so confirm/paste the actual link with the user rather than guessing it.
```
RoyPriceCS50: Change this to ask the user for the Trello hyperlink?  
Claude has no idea what's in our Trello board ?

Or just have the ticket number in the PR description?  We all know where the board is.

.claude/skills/new-ticket/SKILL.md
garypolzin: I've changed the wording - these are just example phrases though, you don't need exact matches

.claude/skills/open-pr/SKILL.md
garypolzin: changed to leave the hyperlink as it is

.claude/skills/pytest-pattern/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+|---|---|
+| One job/dataset only | `projects/<project>/<stage>/unittest_data/*.py` |
+| Multiple jobs/datasets in a project | `projects/<project>/unittest_data/*.py` |
+| Multiple projects | `projects/utils/unittest_data.py` |
+
+File naming: `test_<module>.py`, mirroring the module under test.
+
+## Running tests locally
+
+Use `uv run pytest <path>` — see CLAUDE.md's "Environment & workflow" section for what's in scope locally vs CI/CD.
```
MarionJHolloway, line 85: you can use uvx pytest <path> here as an alias

MarionJHolloway (approved):

PR #1438, opened 2026-07-27
1820 - Reduce SLV prepare rows to the downstream retention window

MarionJHolloway:
## Description
Trello ticket [#1820](https://trello.com/c/iMtTr0h9/1820-pivot-6a-reduce-workplace-dataset-for-pivoting-to-match-job-role-datset)

`_00_prepare` carried the full cleaned ASCWDS workplace data, but its merge partner `ind_cqc_07_04_estimate_job_roles` is already reduced upstream, so most of those rows could never survive the join. Applies the same retention rule (last 2 financial years in full, quarterly snapshots before that) directly to the scan, so the predicate is pushed down to the parquet source.

`reduced_data_filter_expr()` moves from `_03_independent_cqc/_07` into `polars_utils/filtering_utils.py` so both pipelines share one definition. `validate_00_prepare` applies the same filter to its compare frame — it's the only validate script affected.

Note — the filter windows off `date.today()`, so SLV and the estimates pipeline can reduce differently either side of a financial year rollover, and prepare/validate can disagree across the 31 March boundary. Pre-existing rather than introduced here, but this PR spreads it to a second pipeline. Follow-up ticket should window off the data's max import date instead.

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1820-slv-reduce-Ind-CQC-SLV:66621531-b7c1-41c0-a1e6-28ce2255dd71)

6,625,694 rows down to 3,682,430 (44.4% dropped), filtered compare count matched the prepare output exactly. Peak RSS 5.0GB, of which 4.74GB is the full-width `source_df` read in the validate script — only used for a row count, so selecting one column would save more than this ticket did. Worth a follow-up.

## Checklist
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
```diff
@@ -1,19 +1,31 @@
...
+    reading it in full before filtering is what we are avoiding.
 
     Args:
         cleaned_ascwds_workplace_source (str): path to the cleaned ascwds workplace data
         prepared_data_destination (str): destination for output
     """
-    workplace_lf = utils.scan_parquet(cleaned_ascwds_workplace_source)
+    workplace_lf = utils.scan_parquet(cleaned_ascwds_workplace_source).filter(
+        reduced_data_filter_expr(date_col=AWPClean.ascwds_workplace_import_date)
+    )
```
RoyPriceCS50, line 28: Need to also call reduce_dataset_to_earliest_file_per_month from polars cleaning utils. 

Otherwise you'll keep old periods where day is not 1of month.

RoyPriceCS50 (commented):
Need to add reduce_dataset_to_earliest_file_per_month to prepare job.

projects/_07_workforce_characteristics/_01_starters_leavers_vacancies/fargate/_00_prepare.py
RoyPriceCS50: Todo in a separate ticket.

RoyPriceCS50 (approved):
Approved

PR #1450, opened 2026-08-05
1852 - Reduces CircleCI credit usage

garypolzin:
## Description
Trello ticket [#1852](https://trello.com/c/LQLMUeuw/1852-circleci-quick-win-cost-fixes)

Reduces CircleCI credit usage (`docker-large`/`remote-docker-large` were 50-66% of monthly spend) by fixing two config issues found during the #1850 SPIKE investigation:

- Removed `no-cache = true` from all 7 `docker-bake.hcl` image build targets, so the Docker Layer Caching CircleCI already pays for (`docker_layer_caching: true` in the `task-containerisation` job) actually takes effect, instead of every image rebuilding from scratch on every push.

All 7 Dockerfiles were checked beforehand and confirmed safe for layer-cache reuse (pinned `requirements.txt`, no `apt-get`, no unpinned URL/git fetches).

## Testing
Ran multiple circleCI branches and checked caching behaviour is working as expected

## Checklist
- [x] Updated CHANGELOG
- [x] Code reviewed by AI


MarionJHolloway:
Can you check the cache storage used across one deployment? It should print the size of the data saved to cache in the circleCI outputs

garypolzin:
> Can you check the cache storage used across one deployment? It should print the size of the data saved to cache in the circleCI outputs

do you know where abouts?

MarionJHolloway:
It will be in the print out for whichever job and section you put the caching in. Give me a ring if you want a hand working it out.

garypolzin:
> It will be in the print out for whichever job and section you put the caching in. Give me a ring if you want a hand working it out.

Found it:
Total size uploaded: 1.5 GiB

MarionJHolloway:
That's quite high. We have 2GB storage and everything persists for 24hours. The way I understand it, that would mean pushing to more than one branch in one day would tip our storage over the edge.

garypolzin:
> That's quite high. We have 2GB storage and everything persists for 24hours. The way I understand it, that would mean pushing to more than one branch in one day would tip our storage over the edge.

Retention is actually 15 days for cache (looks like we didn't reduce that one), not 24 hours, but that works in our favour here.

The 2GB allowance is 2 GB-months, not a flat 2GB cap - it's `size × retention length (as a fraction of a month)`, billed once when the cache is saved. So a 1.5GB cache at 15-day retention is ~0.75 GB-months per save, not 1.5GB flat.

More importantly, this cache isn't per-branch. The key is `v1-test-deps-{{ checksum "uv.lock" }}` - it only depends on the lockfile's content, not the branch. Every branch with the same uv.lock (almost all of them, almost all the time) restores and shares the same single cache entry, and save_cache skips re-uploading if that key already exists. So pushing to multiple branches in a day doesn't create multiple caches - it's one shared entry, only re-saved when uv.lock actually changes (or the entry naturally expires).

MarionJHolloway:
Okay, my other concern is that caching actually makes the test step of the pipeline slower (by around 10-20secs) won't that increase our usage costs, rather than decrease them? I'm assuming usage is billed by how long a container is running?

garypolzin:
> Okay, my other concern is that caching actually makes the test step of the pipeline slower (by around 10-20secs) won't that increase our usage costs, rather than decrease them? I'm assuming usage is billed by how long a container is running?

Might be hard to quantify slower. I've looked at a few branches in the past where the commit was just updating CHANGELOG and the length of time it took to complete the tests varied by 20 seconds despite the tests being identical

I've fed your response back to Claude anyway and the response was:
_"Do I still think there's a benefit? Structurally, yes — it's not doing nothing, and there's no real reason to think it's net negative. But it's small enough that it's genuinely not worth arguing over. Given it's already implemented, tested, and confirmed to hit correctly, I'd lean toward keeping it — but if the preference is to keep this PR minimal and only ship the change we can actually prove moved the needle (the docker-bake.hcl fix), dropping the test job caching is an equally defensible call. Your choice either way, and I can pull it out cleanly if you'd rather simplify the diff."_

I'm happy to agree with that, I don't mind either way

MarionJHolloway:
How about spliting the PR in 2. Get the dockerbake stuff in and then we could add logging to the remaining step to see how long it takes to do the cache vs the download? Happy to take over that half of the ticket, shouldn't be much work with claude's help.

garypolzin:
> How about spliting the PR in 2. Get the dockerbake stuff in and then we could add logging to the remaining step to see how long it takes to do the cache vs the download? Happy to take over that half of the ticket, shouldn't be much work with claude's help.

OK that's out of this PR now and heading to a new branch

MarionJHolloway (approved):

PR #1451, opened 2026-08-05
1853 - Reverted moving of `split_s3_uri` functions

garypolzin:
## Description
Trello ticket [#1853](https://trello.com/c/0jS3E7gZ/1853-add-missing-util-to-ascwds-prepare-job-role-counts-docker)

Reverted part of PR 1825 which moved `split_s3_uri` - didn't clock that there were already pyspark and polars versions and the new location added complications in every Dockerfile, so I've done a straight revert to what it was before.

## Testing
- [X] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1853-reorder-uri-Ind-CQC-Filled-Post-Estimates:4467946b-54cb-404b-8732-1c608658cd7c)

- [x] Amended CHANGELOG


projects/_01_ingest/capacity_tracker/jobs/ingest_capacity_tracker_data.py
```diff
@@ -6,21 +6,21 @@
 from pyspark.sql import DataFrame
 
 from projects._01_ingest.utils import utils as iUtils
-from utils import s3_file_utils, utils
+from utils import utils
```
MarionJHolloway, line 9: Some of the scripts import this function from utils.py and some from polars_utils/utils.py. What is the logic behind that? I would have thought we'd want them all pointing to the same function located in either one of those two place?

garypolzin: Pyspark jobs import from utils and Polars from polars_utils - this is what always happened before I attempted merging them together and hit issues so I've gone back to that

MarionJHolloway (approved):
Now I get it. Yeah that makes sense and will make ripping out the pyspark code easier at the end. Looks good. I did a manual check that each was importing from the right utils.

PR #1462, opened 2026-08-10
1868 - Fix ModuleNotFoundError breaking PySpark Glue ingestion jobs

MarionJHolloway:
## Description
Trello ticket [#1868](add link)

The ASC-WDS Glue ingestion job was failing in production with `ModuleNotFoundError: No module named 'polars'`. Root cause: `projects/_01_ingest/utils/utils.py` (imported by all four legacy PySpark Glue ingestion jobs — ASC-WDS, Capacity Tracker, CQC PIR, ONS) imported `split_s3_uri` from `polars_utils.utils`, which does `import polars` at module level — unavailable in the Glue PySpark runtime. This broke all four jobs, not just ASC-WDS.

Fix: inlined `split_s3_uri`'s two-line body directly into `projects/_01_ingest/utils/utils.py` instead of importing it from either `polars_utils` or the repo-wide `utils.utils` (which would have introduced a pyspark/pydeequ dependency for the same reason, in the opposite direction). This keeps the file dependency-free, as it was before the regression.

A full repo audit confirmed this was the only incorrect `split_s3_uri` import — no other Glue job or the CQC ingestion pipeline (already Polars/Fargate) is affected.

## Testing
- [x] Unit tests passing
- [ ] Successful [Step Function run](add link) - not possible to run in branch - permissions issues

- `uv run pytest projects/_01_ingest/tests/utils/test_utils.py` — 14 passed.
- Confirmed `projects._01_ingest.utils.utils` now imports cleanly with `polars`, `pyspark`, and `pydeequ` all blocked from `sys.modules` (simulating the Glue runtime it broke in).
- Manually triggered `Ingest-ASCWDS` against a real file in the branch's non-prod environment: it got past the previously-failing import and reached the parquet write step (proving the `ModuleNotFoundError` is fixed), then hit an unrelated `PATH_ALREADY_EXISTS` caused by the branch's dataset bucket doubling as both source and destination in this test setup (no separate raw bucket / cross-account access from non-prod) — not a defect in this change. 

- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist

### DIFF
--- a/projects/_01_ingest/utils/utils.py
+++ b/projects/_01_ingest/utils/utils.py
@@ -3,7 +3,7 @@ from typing import Any
 
 import boto3
 
-from polars_utils.utils import split_s3_uri
+from utils.utils import split_s3_uri
 
 TWO_MB = 2000000
 
--- a/projects/_01_ingest/utils/utils.py
+++ b/projects/_01_ingest/utils/utils.py
@@ -3,8 +3,6 @@ from typing import Any
 
 import boto3
 
-from utils.utils import split_s3_uri
-
 TWO_MB = 2000000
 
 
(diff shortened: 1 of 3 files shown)


projects/_01_ingest/utils/utils.py
```diff
@@ -119,6 +117,21 @@ def construct_s3_uri(bucket_name: str, key: str) -> str:
     return f"s3://{trimmed_bucket_name}/{key}"
 
 
+# converted to polars -> polars_utils.utils.py
+def split_s3_uri(uri: str) -> tuple[str, str]:
```
garypolzin, line 121: just highlighting that we already have 2 versions of this function so possibly not the best idea to create a 3rd one

garypolzin (changes requested):

projects/_01_ingest/utils/utils.py
MarionJHolloway: Yeah, I can do that but it pulls in extra dependencies that we don't need here and might cause more problems when we eventually start converting these  to polars. Neither solution is great.

garypolzin: can the other pyspark ones point to this one instead of the one in `utils\utils.py`?

garypolzin (approved):
Decided to have this as a temporary home and opening a new ticket to combine related file for all file helpers

PR #1465, opened 2026-08-10
1861 Part 1 of 2 - Uncomment remaining polars impute and refactor calculate rolling average

RoyPriceCS50:
## Description
Trello ticket [1861](https://trello.com/c/93gsR9c4)

This is first PR for 1861.

This PR has uncommented polars function calls in imputation job. This caused oom error.

I logged the running memory usage of the job and found that calculate_rolling_average increased it by ~10GB per 5 seconds until reaching 60GB limit and crashing. 

The .over() was attempting the calculate over large groups (3 primary services) whereas the refactor aggregates the data up to reduce it's size then joins back to the input.

Testing memory usage with the refactor showed a tiny change in the amount of memory being used as this function was called.

However, even with refactor, the job as a whole still fails due to oom. 
Big contributors to usage are forward filling, model imputation and model rate of change.
When I'm testing, i'm calling each function individually by using .collect().lazy() after each. This lets the job succeed.
Since a big memory user was model_interpolation, these are now followed by collect calls to split the job into three parts.

This branch still fails to run because the collect calls are not in it from PR part 2. 

## Testing
- [x] Unit tests passing
- [x] Successful [Job / Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1861-rlg-avg-oom-Ind-CQC-Filled-Post-Estimates:14700dee-c74c-4d10-a3a9-77dbe5bde323)

See Trello ticket for Athena query. That ran against a pipeline run that had both parts of PR in it.

- [x] Unit tests added
- [x] Docstrings added
- [x] Updated CHANGELOG
- [x] Code reviewed by AI

## Reviewer checklist


projects/_03_independent_cqc/_03_impute/fargate/impute_ind_cqc_ascwds_and_pir.py
```diff
@@ -86,23 +86,41 @@ def main(cleaned_ind_cqc_source: str, destination: str) -> None:
...
     )
 
-    # model_calculate_rolling_average - posts_rolling_average_model
+    lf = calculate_rolling_average(
+        lf,
+        IndCQC.imputed_filled_post_model,
+        f"{NumericalValues.number_of_days_in_window}d",
+        [IndCQC.primary_service_type],
+        IndCQC.posts_rolling_average_model,
+    )
```
garypolzin: I think there are still some things we can try in terms of making existing functions more efficient before we resort to the .collect().lazy() in part 2. Given that will take some time to investigate, we can't merge this PR on the basis that it is OOM-ing and therefore the pipeline will break, which we can't have.

I'm happy with the calculate_rolling_average rewrite, that looks good and everything points to that being more efficient than it was, so I'd say we'd be best having the first PR as purely 'make the calculate_rolling_average function more efficient' and then the 2nd PR is fully focussed on introducing the remaining functions and solving OOM issues

PR #1466, opened 2026-08-10
1873 - Replace groupby-agg-explode-join broadcasts with .over() in job role impute utils

garypolzin:
## Description
Trello ticket [#1873](https://trello.com/c/dcoshNNJ/1873-make-job-role-impute-more-efficient)

Reviewed `impute_utils.py` for efficiency. Two functions (`create_imputed_ascwds_job_role_counts`, `get_percent_share_ratios`) used a groupby→agg→explode→join pattern to broadcast a per-group computed value back onto every row — a leftover streaming-engine workaround. Per the repo's `over-vs-join` finding, a join that broadcasts a value onto every row costs more peak memory than `.over()`'s in-place fallback, so both functions were rewritten to use `.over()` directly (matching the same idiom already used in `primary_service_rate_of_change.py`). The rolling-ratio and size-group functions in the same file were reviewed and left unchanged — the rolling ratio genuinely benefits from pre-aggregation before its expensive windowed computation, and the size-group expression is a small fixed branch chain with no real perf concern.

Impute job ran in 1m45 compared to 2m in main - this will be more impactful when further functionality is added to the job. 

Peak memory is ~3-4x lower than main's join-based version on a 5.76M-row synthetic benchmark (109MB vs 371MB when pre-sorted, 19MB vs 455MB when shuffled), at the cost of some extra time when input isn't pre-sorted (3.1s vs 1.5s in the same benchmark).

## Testing
- [x] Unit tests passing
- [x] Successful [Step Function run](https://eu-west-2.console.aws.amazon.com/states/home?region=eu-west-2#/v2/executions/details/arn:aws:states:eu-west-2:856699698263:execution:1873-impute-perf-Ind-CQC-Filled-Post-Estimates-By-Role:04d5e10a-1759-42ff-a5c2-f0e1dbe4a5a0) (crawler failed as I'd set it going manually so it was already running when triggered, not a genuine issue)
- [x] Outputs checked in Athena - returns zero differences when compared to main

- [x] Updated CHANGELOG
- [x] Code reviewed by AI


AchyutRatkanthiwar (approved):
Looks good

